PR #13, opened 2015-06-07
Don't force the exit code to 0 by default

cpitclaudel:
Although it's convenient to force the exit code to zero for lit testing,
it is more conventional (and better for interfacing with other systems)
to return a non-zero error if verification fail: at the moment, the only
way (IIUC) to find out whether verification succeeded (on the command
line) was to parse the textual output of Boogie.

This patch adds a CountVerificationErrors option, and enables it by
default. The patch also sets this option to zero for tests, lest lit get
confused in test cases where verification actually fails.

### DIFF
--- a/Source/BoogieDriver/BoogieDriver.cs
+++ b/Source/BoogieDriver/BoogieDriver.cs
@@ -89,8 +89,14 @@ namespace Microsoft.Boogie {
           goto END;
         }
       }
-      ExecutionEngine.ProcessFiles(fileList);
-      return 0;
+      
+      int retv = ExecutionEngine.ProcessFiles(fileList);
+      if (CommandLineOptions.Clo.CountVerificationErrors == 0) {
+          return 0;
+      }
+      else {
+          return retv;
+      }
 
     END:
       if (CommandLineOptions.Clo.XmlSink != null) {
--- a/Source/Core/CommandLineOptions.cs
+++ b/Source/Core/CommandLineOptions.cs
@@ -412,6 +412,7 @@ namespace Microsoft.Boogie {
     public string PrintFile = null;
     public int PrintUnstructured = 0;
     public bool UseBaseNameForFileName = false;
+    public int CountVerificationErrors = 1;
     public int DoomStrategy = -1;
     public bool DoomRestartTP = false;
     public bool PrintDesugarings = false;
@@ -530,6 +531,7 @@ namespace Microsoft.Boogie {
       Contract.Invariant(0 <= ModifiesDefault && ModifiesDefault < 7);
       Contract.Invariant((0 <= PrintErrorModel && PrintErrorModel <= 2) || PrintErrorModel == 4);
       Contract.Invariant(0 <= EnhancedErrorMessages && EnhancedErrorMessages < 2);
+      Contract.Invariant(0 <= CountVerificationErrors && CountVerificationErrors < 2);
       Contract.Invariant(0 <= StepsBeforeWidening && StepsBeforeWidening <= 9);
       Contract.Invariant(-1 <= this.bracketIdsInVC && this.bracketIdsInVC <= 1);
       Contract.Invariant(cce.NonNullElements(this.proverOptions));
@@ -1037,6 +1039,10 @@ namespace Microsoft.Boogie {
 
         case "enhancedErrorMessages":
           ps.GetNumericArgument(ref EnhancedErrorMessages, 2);
+          return true;
+
+        case "countVerificationErrors":
+          ps.GetNumericArgument(ref CountVerificationErrors, 2);
           return true;
 
         case "printCFG":
@@ -1879,6 +1885,11 @@ namespace Microsoft.Boogie {
 
   /useBaseNameForFileName : When parsing use basename of file for tokens instead
                             of the path supplied on the command line
+
+  /CountVerificationErrors:<n>
+                0 - Always set exit code to 0 after verification.
+                1 (default) - Set exit code to non-zero value if errors were
+                              encoutered during verification.
 
   ---- Inference options -----------------------------------------------------
 
--- a/Source/ExecutionEngine/ExecutionEngine.cs
+++ b/Source/ExecutionEngine/ExecutionEngine.cs
@@ -450,9 +450,10 @@ namespace Microsoft.Boogie
 
     static readonly ConcurrentDictionary<string, CancellationTokenSource> RequestIdToCancellationTokenSource = new ConcurrentDictionary<string, CancellationTokenSource>();
 
-    public static void ProcessFiles(List<string> fileNames, bool lookForSnapshots = true, string programId = null)
+    public static int ProcessFiles(List<string> fileNames, bool lookForSnapshots = true, string programId = null)
     {
       Contract.Requires(cce.NonNullElements(fileNames));
+      int nErrors = 0;
 
       if (programId == null)
       {
@@ -463,9 +464,9 @@ namespace Microsoft.Boogie
       {
         foreach (var f in fileNames)
         {
-          ProcessFiles(new List<string> { f }, lookForSnapshots, f);
+          nErrors += ProcessFiles(new List<string> { f }, lookForSnapshots, f);
         }
-        return;
+        return nErrors;
       }
 
       if (0 <= CommandLineOptions.Clo.VerifySnapshots && lookForSnapshots)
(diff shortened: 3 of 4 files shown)


MontyCarter:
I just wanted to add a vote for this patch.  I am also needing an exit code for failed parsing, name resolution and type checking, and this patch solves my problem.

delcypher:
@cpitclaudel Thanks for taking a look at this. I never liked the fact the ``boogie`` returned a 0 exit code even if verification or type checking/parsing failed.

I don't want to accept the patch as it is right now because I think it can be improved.

* Hardcoded exit codes (i.e. ``1``). We shouldn't be doing that. Use an enum.
* I'm afraid I don't have time to test this right now but I am right in thinking that a parse error gives the same exit code as a verification failure. Do we want that behaviour? Also what if there are more than 255 verification failures? I don't think the exit code should be the number of verification failures. Rather there should just be an exit code that means not all procedures could be verified.
* I don't like that fact you've added `` CountVerificationErrors`` as an option and disabled it for ``lit``. With ``lit`` we should be testing ``boogie`` as close to its normal operation as possible. So I don't think there should be a  `` CountVerificationErrors`` option at all. Instead this behaviour should be on **all the time**. We would then need to teach the ``lit`` tests to check the exit code is as expected. I use a simple python script to do this in my tests

```python
#!/usr/bin/env python
"""
This simple script runs a program and returns 0
if the exit code was as expected, otherwise
it returns 255
e.g.
$ eec 0 true
$ eec 1 ls -l
"""
import sys
import subprocess

if __name__ == '__main__':
    if len(sys.argv) < 3:
       sys.stderr.write('Must pass at least 2 arguments\n')
       sys.exit(255)

    expectedExitCode = int(sys.argv[1])
    cmdLine = sys.argv[2:]
    p = subprocess.Popen(cmdLine)
    result = p.wait()

    if result == expectedExitCode:
        sys.exit(0)
    else:
        print("ERROR: Expected exit code {} but was {}".format(expectedExitCode, result))
        sys.exit(255)
```

It can then be invoked in lit by adding a substitution and it can be used in tests like so

```
RUN: %eec 1 %boogie %s
```

cpitclaudel:
> Hardcoded exit codes (i.e. ``1``). We shouldn't be doing that. Use an enum.

What would we do with an enum? We're returning an int in the end, right? And nothing else that Main() seems to use the result of that function. Also, how do we combine enum values? (that is, what is the result of VerificationError + SomeOtherError if we go the enum route?)

> I'm afraid I don't have time to test this right now but I am right in thinking that a parse error gives the same exit code as a verification failure. Do we want that behaviour? Also what if there are more than 255 verification failures? I don't think the exit code should be the number of verification failures. Rather there should just be an exit code that means not all procedures could be verified.

That seems fine to me. I aligned onto the Dafny version, which returns the number of errors. We use it to track whether chances have decreased or increased the number of error in tests that are expected to fail due to current limitations of the engine. But a saturating 0-1 counter seems fine, too.

> I don't like that fact you've added `` CountVerificationErrors`` as an option and disabled it for ``lit``. With ``lit`` we should be testing ``boogie`` as close to its normal operation as possible. So I don't think there should be a  `` CountVerificationErrors`` option at all.

Agreed as well. There, too, I modelled this on a recent Dafny fix where we introduced that option (it used to be the case that, when Dafny saw the `useBaseName` flag, it would return 0 instead of an error count.

> Instead this behaviour should be on **all the time**. We would then need to teach the ``lit`` tests to check the exit code is as expected. I use a simple python script to do this in my tests
> (snip)

Agreed, too; I don't really have time to run and add expected return codes to the whole test suite at the moment, though. Feel free to build upon the current patch!

zvonimir:
Team SMACK would greatly appreciate if this functionality would be added sooner rather than later. I have to say I am not particularly excited about exit code dynamically encoding the number of failures. I think that exit codes should be enum constants, and you would have different codes for parsing errors, type checking errors, etc.
Anyhow, please let me know if we could help out somehow.

zvonimir:
Oh, and another thing. I think that exit codes for "verified" and "X procedures could not be verified" should be the same. Exit codes should capture internal Boogie errors, and not the fact that Boogie did not manage to verify something.

delcypher:
@cpitclaudel 
> What would we do with an enum? We're returning an int in the end, right? And nothing else that Main() seems to use the result of that function. Also, how do we combine enum values? (that is, what is the result of VerificationError + SomeOtherError if we go the enum route?)

The enum would contain the different return codes we can return. There is no combining of error codes.

@zvonimir 
> Oh, and another thing. I think that exit codes for "verified" and "X procedures could not be verified" should be the same. Exit codes should capture internal Boogie errors, and not the fact that Boogie did not manage to verify something.

I disagree. What is your justification for not being able to determine from the exit code if verification? For ``lit`` and general scripting knowing if all procedures were verified or if one or more could not be verified.

In regards to getting this done, I know how to do it but I can't right now because I'm doing an internship and right now I don't have permission to contribute to Boogie (code ownership issues). So if you want this done quickly someone else needs to do it.

zvonimir:
"Unix-like systems have a strong convention that an exit status of 0 denotes success, and any non-zero exit status denotes failure."
See for example here: http://programmers.stackexchange.com/questions/153926/non-zero-exit-status-for-clean-exit
In my mind, Boogie reporting an assertion violation in the input Boogie program does not constitute a failure but rather a successful run. But I guess that is open to various interpretations.
Anyhow, even if I don't completely agree with that, I can live with 0 exit code being reserved just for "verified" and nothing else, as long as there are some kind of error codes being returned.

cpitclaudel:
> > What would we do with an enum? We're returning an int in the end, right? And nothing else that Main() seems to use the result of that function. Also, how do we combine enum values? (that is, what is the result of VerificationError + SomeOtherError if we go the enum route?)

> The enum would contain the different return codes we can return. There is no combining of error codes.

I may have expressed my thoughts improperly. What I meant to point out is that, if Boogie is run on a set of files, then there may be different outcomes for each file. If one of them succeeds, another fails to typecheck, and the third fails to verify, then we have three error conditions to report about, but only one error code. This is what I meant by combining.

It may be enough to have three possible return states: 
* everything was properly verified
* everything went fine, but verification failed for some files
* Some more serious error occured

Or we could follow http://www.freebsd.org/cgi/man.cgi?query=sysexits&sektion=3

@zvonimir Being able to easily tell if verification failed is a big plus for me, so I'm not too big a fan of Boogie returning 0 if verification does not go through.

cpitclaudel:
> Also what if there are more than 255 verification failures? I don't think the exit code should be the number of verification failures. Rather there should just be an exit code that means not all procedures could be verified.

NUnit does that. I don't particularly like it. Also, the limit of 255 only apply to posix systems; windows has 32-bits signed exit codes.

cpitclaudel:
> In regards to getting this done, I know how to do it but I can't right now because I'm doing an internship and right now I don't have permission to contribute to Boogie (code ownership issues). So if you want this done quickly someone else needs to do it.

That's not an issue. I can merge this PR myself when we reach a consensus, and the patch is trivial enough that anyone can implement it once we know what we want.

delcypher:
@cpitclaudel 
> I may have expressed my thoughts improperly. What I meant to point out is that, if Boogie is run on a set of files, then there may be different outcomes for each file. If one of them succeeds, another fails to typecheck, and the third fails to verify, then we have three error conditions to report about, but only one error code. This is what I meant by combining.

Ah I forgot about the fact boogie can take more than one boogie program.

I've thought about it a little more and I think it's worth pointing out an issue I've had with making the exit code indicate verification failure/success in my own project. Under linux sometimes ``mono`` will crash and when this happens the exit code will be non-zero but the exit code is not always consistent. This makes it impossible to determine from the exit code alone whether mono crashed or whether verification failed. In light of this it would probably be wise to reserve non zero exit codes for fatal errors (e.g. parse errors) and communicate verification failures in a different manner.

There are better ways we can communicate verification success or failure, we could emit YAML, JSON or XML for example which are machine readable. My preference would be YAML because XML readability sucks.

> NUnit does that. I don't particularly like it. Also, the limit of 255 only apply to posix systems; windows has 32-bits signed exit codes.

For the nunit-console runner this design decision makes complete sense. Typically all tests should be passing so all you care about is whether all tests passed or if one or more failed. If one or more failed things are bad.

Also a solution that only works on Windows is not acceptable at all. Boogie needs to be cross platform and any windows-isms do not belong in it's code base unless there's a really good reason for it.

cpitclaudel:
As much as I agree that keeping the codebase free of windows-specific tricks is a reasonable goal, I don't think our design choices should be driven by mono's bugs, both because mono's importance seems to be bound to decrease in the not-too-far future, and because mono's crashes are still fairly rare. In addition, if one is afraid of such crashes, then it is always possible to parse the output of the tool to make sure of exactly what happened.

I feel that limiting non-zero returns to parsing failures and other serious errors is a serious limitation, that will be hard to change in the future, should we want to make changes. If we assume that return codes are mostly going to be useful in scripted interactions with Boogie, then parse errors are probably going to be less frequent than verification issues. Supporting two distinct error codes for these two situations seems convenient, too, as it satisfies both those who just want to detect serious errors (check for code >= 2) and those who are interested in having an easy way to notice when verification errors occur.

I like grep's approach here: 0 if the search failed, 1 if it succeeded, and 2 if there's a problem with the query itself.

As for emitting some sort of machine-readable report: sure! But I don't think it is the real issue at hand here. This would introduce a new Boogie textual API, in addition to the existing programmatic .Net dll API; this PR, in contrast, is about leveraging an existing API, namely error codes.

zvonimir:
Team SMACK has a preference for the @delcypher's solution proposed above, which is what we also proposed. But more importantly, we would like to see some functionality of this kind implemented sooner rather than later since we can live with both solutions.

PR #23, opened 2015-11-16
Use EndCurly token for the ReturnCmd when creating unifiedExit

qunyanm:
Instead of using NoToken for the ReturnCmd, use the EndCurly when
creating unifiedExit.


delcypher:
@qunyanm I can't really give a good review here as I don't know anything about Boogie's VC. However the TravisCI build failure so you should take a look at that

https://travis-ci.org/boogie-org/boogie/jobs/91444009

Issue #29, opened 2016-01-14
Adding a Floating-Point Type


Checkmate50:
Boogie Team,

I'm Dietrich Geisler, and undergraduate student working for Dr. Zvonimir Rakamaric at the University of Utah.  Over the last year, I have worked on adding the floating point type to boogie based on the floating point update to z3.

The merge I am proposing adds the floating point type to boogie, allowing variables and constants to be declared as described below.  There are several issues with my implementation that I have listed.  I'm sure that there are more issues and bugs I haven't found, so if you find any issues, please let me know.

---
## Syntax Documentation

Declaring Variables:

The syntax for declaring a floating point variable is as follows:
var name: float(exp man);
Where exp is the size of the float exponent and man is the size of the float mantissa

It is also acceptable to use the following syntax:
var name: float();
This syntax assumes the float exponent to be size 8 and the mantissa to be size 24

example:
var x: float(11 53)
Declares a variable called x with a exponent sized 11 and mantissa sized 53

---

Declaring Constants:

All of the following syntax are viable for declaring a floating point constant:
float(dec)
float(exp_val man_val)
float(dec exp man)
float(exp_val man_val exp man)

Where dec is the decimal value of the constant,
exp_val/man_value are the integer values of their respective fields,
And exp/man are the sizes of their respective fields.
Note that when exp and man are not specified, they are given sizes 8 and 24 respectively

---

Defined Operations:

Given two floating point values x and y with the same size of exponent and mantissa
The following operations operate as defined in Operators section of this document:
http://smtlib.cs.uiowa.edu/theories-FloatingPoint.shtml
(Note that rounding mode is always assumed to be Round to Nearest Even (RNE))
(Also note that operations not listed here are undefined)

operatation     boogie syntax
neg(x)          -x
add(x, y)       x + y
sub(x, y)       x - y
mul(x, y)       x \* y
div(x, y)       x / y
leq(x, y)       x <= y
lt(x, y)        x < y
geq(x, y)       x >= y
gt(x, y)        x > y
eq(x, y)        x == y

---

Other:

The following special values can be declared with the following syntax:
Value       Declaration
NaN         fp(nan exp man)
+oo         fp(oo exp man) OR fp(INF exp man)
-oo         fp(-oo exp man) OR fp(INF exp man)
-zero       fp(-zero exp man)
Where exp and man are the sizes of the exponent and mantissa respectively

---

Known Issues:

There is currently no way to convert from a floating point to any other data type

[...]

PR #35, opened 2016-06-07
Floating Point Support

Checkmate50:
Dear Boogie Team,

I've made the changes so that floating points are supported as described here:

Note that I originally created an issue for these proposed changes which can be found here:

https://github.com/boogie-org/boogie/issues/29

It should be noted that I differed from the recommendations to my original question in two areas:

1.  I changed the syntax of declaring a floating-point constant from
float(sign, exp_value, sig_value)   OR   float<exp, sig>(bitvalue)
to
fp(sign, exp_value, sig_value)     OR   fp<exp, sig>(bitvalue)

2.  Builtin functions appear to work via such syntax as:
function {:builtin "(_ to_fp eb sb) RNE" } TO_FLOAT_BITVECTOR(bv) returns(float<eb, sb>);
Where eb is the exponent size and sb is the significand size.

If you have any questions or feedback, please let me know and I'll be glad to make changes.  Also, I had some trouble with getting the line endings to work after the line ending normalization a few months ago, so let please let me know if my line endings are incorrect.


RustanLeino:
Among these file changes, it seems the modified BoogiePL.atg is missing.

Checkmate50:
Ah, so there IS a compiler modification file.  I was wondering about that; modifying the parser by hand didn't seem right, but I haven't used an atg file before and didn't notice it.  Thanks for the tip; I'll fix that up.

PR #46, opened 2016-10-16
Fixed an issue with floating-point type expansions

Checkmate50:
As the title suggests, this fixes a bug where Boogie would crash when the floating-point type was expanded.  Also adds associated tests.


zvonimir:
@RustanLeino, @delcypher, @shazqadeer : Could one of you please review this pull request from my student?
It fixes a bug in his implementation of support for floating-point in Boogie. Thanks!

PR #51, opened 2017-02-03
Support for syntax highlighting of bitvector constants in vim.

pramodsu:
This commit makes bitvector constants like 0bv32 look pretty in vim.


delcypher:
@pramodsu LGTM. Thanks.

PR #58, opened 2017-05-21
Fix #57: propagate tokens in commutativity checks

### LINKED ISSUE #57: [civl] Commutativity check error messages contain no location information
Although the procedure names are included, tools such as the emacs mode cannot easily extract this information, which makes the the experience less enjoyable.

It would be nice if these errors contained filenames and line numbers like most other Boogie errors.

wilcoxjay:
I took a crack at fixing #57 by propagating tokens to the generated ensures clause and through return commands when cloning blocks.

I am not sure whether this was the right approach, so feedback is welcome!

### DIFF
--- a/Source/Concurrency/MoverCheck.cs
+++ b/Source/Concurrency/MoverCheck.cs
@@ -119,7 +119,10 @@ namespace Microsoft.Boogie
             }
             foreach (Block block in blocks)
             {
-                if (block.TransferCmd is ReturnCmd) continue;
+                if (block.TransferCmd is ReturnCmd) {
+                    blockMap[block].TransferCmd = new ReturnCmd(block.TransferCmd.tok);
+                    continue;
+                }
                 List<Block> otherGotoCmdLabelTargets = new List<Block>();
                 List<string> otherGotoCmdLabelNames = new List<string>();
                 GotoCmd gotoCmd = block.TransferCmd as GotoCmd;
@@ -204,7 +207,7 @@ namespace Microsoft.Boogie
                 DisjointnessExpr(program, first.thatOutParams.Union(thisInParamsFiltered), frame),
                 DisjointnessExpr(program, first.thatOutParams.Union(second.thisOutParams), frame));
             // TODO: add further disjointness expressions?
-            Ensures ensureCheck = new Ensures(false, Expr.Imp(Expr.And(linearityAssumes), transitionRelation));
+            Ensures ensureCheck = new Ensures(first.proc.tok, false, Expr.Imp(Expr.And(linearityAssumes), transitionRelation), null);
             ensureCheck.ErrorData = string.Format("Commutativity check between {0} and {1} failed", first.proc.Name, second.proc.Name);
             List<Ensures> ensures = new List<Ensures> { ensureCheck };
             


bkragl:
Good idea to improve the token propagation in the commutativity checker! However, I think there are a few things to consider and potentially improve.

There are three types of commutativity checks implemented in the following methods (your changed modified the first):
1) CreateCommutativityChecker
2) CreateGatePreservationChecker
3) CreateFailurePreservationChecker
4) CreateNonBlockingChecker
All but the last involve two atomic actions, which are currently implemented as "GOTO programs" in special assures clauses. Now the question is, which line numbers are the most relevant in each case? I think in 1) it is the beginning of both atomic action specs, in 2) and 3) it is the beginning of the atomic action spec for one action and the failing gate for the other, and in 4) it is again the beginning of the atomic action spec.

Currently some error messages point to a return in an atomic  action spec (the one at the end of a failing path). I think it would be better to have the error messages show line numbers as I suggested above and instead fix the token propagation when copying and stitching together blocks of atomic action specs, which would show up in the printed error trace.

Before merging a fix, the expected output for regression tests should be updated. I actually noticed that there is currently only one file that triggers 2) and 3). So there is a need for more negative test cases.

wilcoxjay:
Thanks for the feedback! 

> There are four types of commutativity checks

I will look into fixing the others as well.

> which line numbers are the most relevant in each case?

I agree with all your points here. But I am unsure how to implement this easily for the atomic actions, since the check is encoded as an ensures clause, and Boogie reports the path along which the ensures is violated. Do you suggest we somehow suppress this part of the error message? 

Since each check has two relevant locations, would you suggest both be reported as "related locations" in the error message?  

>  the expected output for regression tests should be updated

I will also update the regression test outputs once we converge on the desired behavior.

bkragl:
If fixed the issue on the civl branch (8e6edfb). The error reporting does not seem to easily allow much flexibility to get CIVL-specific error messages, but I hope my choices make sense and help.

In general I think the execution traces are important and should not be suppressed. This is how the error messages for the four types of mover checks now look like:

**Commutativity**
```
left-mover.bpl(8,15): Error BP5003: A postcondition might not hold on this return path.
left-mover.bpl(19,32): Related location: Commutativity check between init and inc failed
Execution trace:
    left-mover.bpl(21,5): first_anon0
```
8 is the return taken in inc (i.e., a pointer to the second action), 19 is the procedure of init (i.e., a pointer to the first action), and the execution trace shows which path through init followed by inc does not have a corresponding path through inc followed by init (in this example there is no branching).

**Gate Preservation**
```
chris2.bpl(7,15): Error BP5003: A postcondition might not hold on this return path.
chris2.bpl(23,3): Related location: Gate of atomic_p_gt2 not preserved by atomic_p_gt1_lower
Execution trace:
    chris2.bpl(7,5): second_anon0
```
7 is the return taken in atomic_p_gt1_lower, 23 is the specific gate of atomic_p_gt2 that does not hold, and the execution trace shows the path taken through atomic_p_gt1_lower.

**Gate Failure Preservation**
```
left-mover.bpl(8,15): Error BP5003: A postcondition might not hold on this return path.
left-mover.bpl(11,32): Related location: Gate failure of ass_eq_1 not preserved by inc
Execution trace:
    left-mover.bpl(8,5): second_anon0
```
8 is the return taken in inc, 11 is the procedure of ass_eq_1 (note that in this case we can not point to a specific gate; it is the negated conjunction of all gates that does not hold), and the execution trace shows the path taken through inc.

**Non-Blocking**
```
left-mover.bpl(26,30): Error: Non-blocking check for block failed
Execution trace:
    left-mover.bpl(26,30): L
```
26 is the procedure of block. The execution trace is superfluous (since the non-blocking checker contains just a single assertion), but for convenience it also points to the blocking procedure.

wilcoxjay:
Great, thanks! 

I'll close this PR and the issue.

PR #59, opened 2017-05-27
Rlimit option

qunyanm:
Add /rlimit and rlimit attribute to specify resource limits


Source/Core/CommandLineOptions.cs
```diff
@@ -2118,6 +2126,8 @@ be n (default 500)
   /timeLimit:<num>
                 Limit the number of seconds spent trying to verify
                 each procedure
+  /rlimit:<num>
```
delcypher: This is far too vague. What does `rlimit` mean? What are its units?

I presume this corresponds to Z3's "rlimit". If so you should state this.

Test/test2/Rlimitouts0.bpl.expect
```diff
@@ -0,0 +1,21 @@
+Prover error: line 105 column 10: max. resource limit exceeded
```
delcypher: This error message is very confusing. Line 105, column 10? What file does that refer to? It can't refer to the boogie program because that has 87 lines.

Test/test2/Rlimitouts0.bpl.expect
```diff
@@ -0,0 +1,21 @@
...
+Rlimitouts0.bpl(9,16): Verification inconclusive (TestTimeouts0)
```
delcypher: I'm not sure sure if your code is responsible for this but `(TestTimeouts0)` looks badly formatted.

Test/test2/Rlimitouts0.bpl
```diff
@@ -0,0 +1,87 @@
+// RUN: %boogie -rlimit:8000 "%s" > "%t"
+// RUN: %diff "%s.expect" "%t"
```
delcypher: Seeing as you're adding a new test you should probably use `%OutputCheck` rather than the incredibly fragile `%diff`.

qunyanm:
I did "git fetch origin" followed by "git rebase origin/master". Would it be sufficient?

Source/Core/CommandLineOptions.cs
qunyanm: Change it to "Z3 resource". Hopefully it makes more sense.

Test/test2/Rlimitouts0.bpl.expect
qunyanm: The line/column number were from Z3 and we printed the error message out as it is. To make it less confusing, we will reprocess the "max. resource limit exceeded" so that it will report as "OutOfResource" instead of just an error.

Test/test2/Rlimitouts0.bpl.expect
qunyanm: I started the file from a copy of test2/Timeouts0.bpl. It looks reasonably formatted in my visual studio window (and there doesn't seem to have any TAB). I am sure what other things to check.

Test/test2/Rlimitouts0.bpl
qunyanm: OK. It is updated to use %OutputCheck. I also rename the functions to TestRlimit*, and made TestRlimit1 to verify so that it is easier to check.

Test/test2/Rlimitouts0.bpl.expect
qunyanm: Sorry, I meant "I am not sure what other things to check".

PR #64, opened 2017-07-10
Extracting map values from error models.

michael-emmi:
For many purposes, including replaying counterexample traces, it would be useful to extract map-typed values from Boogie and Corral counterexample traces. This pull request addresses that by extending Corral’s `boogie_si_record` mechanism.

Consider the following program `a.bpl`:
````
procedure boogie_si_record_map(x:[int]int);

{
  var M: [int] int;
  var x: int;
  assume M[0] == 0;
  assume M[1] == 11;
  call boogie_si_record_int(x);
  call boogie_si_record_map(M);
  assert false;
}
````
Without the modifications of this pull request, we observe the following behavior:
````
$ Boogie.exe a.bpl /stratifiedInline:1
b.bpl(12,3): Error BP5001: This assertion might not hold.
Execution trace:
    b.bpl(8,3): anon0
    value = 2
    value = (_ (as-array) (k!0))

````
Note that the map-type value simply refers to the identifier `k!0` of the Z3 model. With the modifications of this pull request, we get the following instead:
````
$ Boogie.exe b.bpl /stratifiedInline:1
b.bpl(12,3): Error BP5001: This assertion might not hold.
Execution trace:
    b.bpl(8,3): anon0
    value = 2
    value = {[1]: 11, [0]: 0, *: 11}

````
The proposed format for map-type values is:
````
{ [x1, x2, …]: x, [y1, y2, …]: y, …, *: z }
````
where applications are comma separated inside braces `{…}` containing comma-separated arguments inside brackets `[…]`, and the default value is guarded with an asterisk `*`.

### DIFF
--- a/Source/VCGeneration/StratifiedVC.cs
+++ b/Source/VCGeneration/StratifiedVC.cs
@@ -2359,6 +2359,42 @@ namespace VC {
         return newCounterexample;
       }
 
+      private String GenerateTraceValue(Model model, Model.Element element) {
+        var str = new StringWriter();
+        if (element is Model.DatatypeValue) {
+          var val = (Model.DatatypeValue) element;
+          if (val.ConstructorName == "_" && val.Arguments[0].ToString() == "(as-array)") {
+            var parens = val.Arguments[1].ToString();
+            var func = model.TryGetFunc(parens.Substring(1, parens.Length - 2));
+            if (func != null) {
+              str.Write("{");
+              var appCount = 0;
+              foreach (var app in func.Apps) {
+                if (appCount++ > 0)
+                  str.Write(", ");
+                str.Write("[");
+                var argCount = 0;
+                foreach (var arg in app.Args) {
+                  if (argCount++ > 0)
+                    str.Write(", ");
+                  str.Write(GenerateTraceValue(model, arg));
+                }
+                str.Write("]: {0}", GenerateTraceValue(model, app.Result));
+              }
+              if (func.Else != null) {
+                if (func.AppCount > 0)
+                  str.Write(", ");
+                str.Write("*: {0}", GenerateTraceValue(model, func.Else));
+              }
+              str.Write("}");
+            }
+          }
+        }
+        if (str.ToString() == "")
+          str.Write(element.ToString());
+        return str.ToString();
+      }
+
       private Counterexample GenerateTraceRec(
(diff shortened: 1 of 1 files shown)


akashlal:
This looks good. Currently this change only affects "boogie /stratifiedInline" and it is a good one. I am not sure why CI is failing; the failing tests have nothing to do with this change.

michael-emmi:
It appears that CI is failing due to [a previous commit](https://github.com/boogie-org/boogie/commit/63b3602e12aa67bf9a06d3388c65745d2ad64fef) which calibrates test output to Z3 version 4.5.0; the Travis CI configuration still asks for Z3 4.4.1. I can confirm that the tests pass on my machine with Z3 4.5.0.

RustanLeino:
We were having various problems with tests failing because of different Z3 versions (in particular, one of the `livevars/daytona_bug2_*` files and `test15/CaptureState.bpl`). By stating a recommended version of Z3, I was hoping we could avoid those problems.  If there are reasons to use some different version of Z3, we could do that too, but otherwise 4.5.0, which is the current version, would seem good.

Btw, the Dafny project has a Z3 executable checked into the project (with appropriate license notices, of course).  Would that make sense for Boogie as well?  I don't know if Boogie developers tend to use Windows machines or if we'd need to check in a Z3 version for more than one platform (and, if so, how we deal with name/path issues).

  Rustan

bkragl:
I'd suggest to add a small script that downloads the recommended version of Z3 and unpacks the executable into Binaries. The script could guess the correct platform, or, perhaps more robustly, take the platform as command line parameter.
If a bash or python script is fine, I could add it.

zvonimir:
Based on my understanding, the problem here is that the Travis CI script was not updated when the recommended Z3 version was changed, which should be easy to fix. Hence, I am not really sure if writing a script as @bkragl suggested would be that helpful since the documentation already clearly says which version of Z3 is needed. Just my 2 cents.

bkragl:
Right, if we only care about the CI, then putting a fixed download URL into .travis.yml is enough. As I can see, @michael-emmi is already working on that update.

michael-emmi:
Ok pull request #65 fixes the build, after plenty of back-and-forth trial and error with Travis CI :-)

Note that this current pull request is independent — it did not make the build any worse than it was before :-)

Can this PR be merged as well now?

PR #65, opened 2017-07-11
Downloading Z3 release 4.5.0 in Travis CI.

michael-emmi:

PR #66, opened 2017-07-18
Extracting maps from prover-evaluate interface.

michael-emmi:
This is a companion to Pull Request #64 which extracts map values out of error models for display in counterexample traces. This PR enables the same feature when the prover evaluation interface is used in place of error models, i.e., via the `/useProverEvaluate` flag. This is the output corresponding to the example listed in PR #64:
````
$ Boogie.exe a.bpl /doModSetAnalysis /stratifiedInline:1 /useProverEvaluate
/Users/mje/Code/smack/a.bpl(11,3): Error BP5001: This assertion might not hold.
Execution trace:
    /Users/mje/Code/smack/a.bpl(8,5): anon0
    value = 1
    value = {[3]: 4, [2]: 3, *: 0}

````
Without this PR, the same command fails with an internal `System.NullReferenceExeption`.

### DIFF
--- a/Source/Provers/SMTLib/ProverInterface.cs
+++ b/Source/Provers/SMTLib/ProverInterface.cs
@@ -2285,6 +2285,37 @@ namespace Microsoft.Boogie.SMTLib
       }
     }
 
+    object GetArrayFromProverResponse(SExpr resp) {
+      var dict = new Dictionary<string, string>();
+      var curr = resp;
+      while (curr != null) {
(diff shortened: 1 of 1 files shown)


michael-emmi:
Note that since PR #65 has not yet been merged, this PR is expected to fail the same couple of tests that are currently failing on the `boogie-org:master` branch.

akashlal:
Thanks Mike. This only affects /stratifiedInline.

PR #67, opened 2017-07-18
JSON format form map-type values in counterexample traces.

michael-emmi:
Switching to JSON format for map-type values in counterexample traces (see PR #64 and #66) for simplified parsing. For example, given the following code in `a.bpl`
````
procedure boogie_si_record_map(x:[int,bool]int);

{
  var M: [int,bool] int;
  var x: int;
  M := M[-2,true := 3][3,false := -5];
  call boogie_si_record_int(x);
  call boogie_si_record_map(M);
  assert false;
}
````
we observe the following
````
$ Boogie.exe a.bpl /doModSetAnalysis /stratifiedInline:1
a.bpl(11,3): Error BP5001: This assertion might not hold.
Execution trace:
    a.bpl(8,5): anon0
    value = 1
    value = {"3,false":-5,"-2,true":3,"*":0}

````
and similarly with the `/useProverEvaluate` option.

### DIFF
--- a/Source/Provers/SMTLib/ProverInterface.cs
+++ b/Source/Provers/SMTLib/ProverInterface.cs
@@ -2294,7 +2294,6 @@ namespace Microsoft.Boogie.SMTLib
       var curr = resp;
       while (curr != null) {
         if (curr.Name == "store") {
-          Console.WriteLine("Got store: {0}", curr);
           var ary = curr.Arguments[0];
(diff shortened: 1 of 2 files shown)


zvonimir:
Good idea! Could we switch error traces completely to JSON?

michael-emmi:
Technically yes, but others may be replying on the current format of error traces.

PR #69, opened 2017-07-27
Handling let bindings in prover array expressions.

michael-emmi:
This PR extends `Microsoft.Boogie.SMTLib.SMTLibProcessTheoremProver.Evaluate` to handle array expressions containing let bindings. The theorem prover’s response to the evaluation of an array variable is sometimes an array expression like this:
````
(let (() (a!1 (store (store (store (() (as const (Array Int Int)) 15) 504 40) 25 26) 520 (- 103)))) (store (store (store a!1 522 (- 596)) 538 213) 31 33))
````
Previously such an expression would cause `Evaluate` to throw an exception, resulting in a null entry in a dictionary, and ultimately leading to a null pointer exception when printing error traces. Now such an expression will be reduced to this:
````
(store (store (store (store (store (store (() (as const (Array Int Int)) 15) 504 40) 25 26) 520 (- 103)) 522 (- 596)) 538 213) 31 33)
````
Then the existing code will parse this to JSON as:
````
{"31": 33,"538":213,"522":-596,"520":-103,"25": 26,"504":40,"*":15}
````

### DIFF
--- a/Source/Provers/SMTLib/ProverInterface.cs
+++ b/Source/Provers/SMTLib/ProverInterface.cs
@@ -2289,7 +2289,38 @@ namespace Microsoft.Boogie.SMTLib
       return expr.ToString().Replace(" ","").Replace("(","").Replace(")","");
     }
 
+    SExpr ReduceLet(SExpr expr) {
+      if (expr.Name != "let")
+        return expr;
+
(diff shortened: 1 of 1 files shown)


akashlal:
Looks good. Merging.

Issue #71, opened 2017-08-25
Multiple declarations of the same variable in a modifies clauses crashes boogie


pramodsu:
Here is an example program:

```
type addr_t = bv32;
type map_t = [addr_t]int;

var mem : map_t;
var thing : [int]int;

procedure {:inline 1} do_something(a : addr_t, b : int)
  returns (mem_of_thing : int)

  modifies thing;
  modifies mem;
  modifies mem;
{
  var s : int;
  
  s := mem[a];
  mem[a] := b;
  
  mem_of_thing := thing[s];
}

procedure proof()
    modifies mem;
    modifies thing;
{
  var a : addr_t;
  var b : int;

  var m : int;

  call m := do_something(a, b);
  assert (m == thing[old(mem)[a]]);
}

```

This results in the error.

```
[ERROR] FATAL UNHANDLED EXCEPTION: System.ArgumentException: An item with the same key has already been added.
  at System.ThrowHelper.ThrowArgumentException (System.ExceptionResource resource) [0x0000b] in <f712f98eb8e445c8918edaf595bbe465>:0 
  at System.Collections.Generic.Dictionary`2[TKey,TValue].Insert (TKey key, TValue value, System.Boolean add) [0x0008e] in <f712f98eb8e445c8918edaf595bbe465>:0 
  at System.Collections.Generic.Dictionary`2[TKey,TValue].Add (TKey key, TValue value) [0x00000] in <f712f98eb8e445c8918edaf595bbe465>:0 
  at Microsoft.Boogie.Inliner.BeginInline (Microsoft.Boogie.Implementation impl) [0x00311] in <646a33ef60854832a600e48efca41b9b>:0 
  at Microsoft.Boogie.Inliner.InlineCallCmd (Microsoft.Boogie.Block block, Microsoft.Boogie.CallCmd callCmd, Microsoft.Boogie.Implementation impl, System.Collections.Generic.List`1[T] newCmds, System.Collections.Generic.List`1[T] newBlocks, System.Int32 lblCount) [0x00044] in <646a33ef60854832a600e48efca41b9b>:0 
  at Microsoft.Boogie.Inliner.DoInlineBlocks (System.Collections.Generic.List`1[T] blocks, System.Boolean& inlinedSomething) [0x000c2] in <646a33ef60854832a600e48efca41b9b>:0 
[...]

PR #72, opened 2017-08-25
Fix for Issue #71: Repeated modifies clause crash.

pramodsu:
This is a fix for Issue #71 on github, which solves a boogie crash that occurs when an inlined procedure has multiple modifies clauses for the same variable. This fix silently ignores the repeated clause.


zvonimir:
I will merge this in since it is likely not the cause of the one failing regression. I will revert it if it is.

PR #76, opened 2017-11-19
Fix predication for procedure calls

jeroenk:
If a predicated procedure was called with the predicate
set to false, it would havoc the return values, because
in that case the return values were never assigned in the
predicated procedure. This lead to problems with command
sequences such as:

  p := true;
  x := foo(p);
  x := bar(!p);

where the parameter of both procedures is the predicate.
Instead preserving x, the call to bar would havoc x. To
mitigate this problem, pass the assigned variables to the
predicated procedures, and use the passed value to preserve
the assigned variables in the case the predicate was false.

The above is achieved by transforming, e.g., the procedure

procedure bar() returns (ret : T)
{
  entry:
    [some code]

  [some blocks]
}

into

procedure bar(_P : bool, _Vret : T) returns (ret : T);
  ensures !_P ==> _Vret == ret;

implementation bar(_P : bool, _Vret : T) returns (ret : T)
{
  entry:
    ret := (if _P then ret else _Vret);
    [predicated version of some code]

  [predicated version of some blocks]
}

### DIFF
--- a/Source/Predication/SmartBlockPredicator.cs
+++ b/Source/Predication/SmartBlockPredicator.cs
@@ -75,6 +75,9 @@ public class SmartBlockPredicator {
     if (cmd is CallCmd) {
       var cCmd = (CallCmd)cmd;
       Debug.Assert(useProcedurePredicates(cCmd.Proc));
+      foreach (IdentifierExpr e in cCmd.Outs) {
+        cCmd.Ins.Add(e);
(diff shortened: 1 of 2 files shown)


jeroenk:
@afd Should probably have a look at this.

Note that the predication code is only used by GPUVerify. I will add a test there.

PR #84, opened 2018-02-17
Fixed floating-point issues with Corral

liammachado:
The ToString method of the BigFloat class was not returning a syntactically correct string representation of floating-point literals. As a result, program.Emit would not print out a program that is semantically equivalent to the in-memory program, which is what Corral expects.

In addition, a VCExprEvaluationException was being thrown in the Evaluate method due to it not being capable of handling floating-point variables.

### DIFF
--- a/Source/Basetypes/BigFloat.cs
+++ b/Source/Basetypes/BigFloat.cs
@@ -157,57 +157,6 @@ namespace Microsoft.Basetypes
       return new BigFloat(isNeg, sig, exp, sigSize, expSize);
     }
 
-    [Pure]
-    public static BigFloat FromBVString(String s)
-    {
-      /*
-       * String must be either of the format (fp (_ bv* *) (_ bv* *) (_ bv* *))
-       * or of the special value format: (_ val * *)
-       * Where * indicates an integer value
-       * and val is one of the special values: NaN, +oo, -oo, zero, -zero
-       */
-
-      if (s.IndexOf('_') == 1)
-      {
-        s = s.Substring(0, s.Length - 1);
-        String[] args = s.Split(' ');
-        int sigSize = int.Parse(args[3]);
-        int expSize = int.Parse(args[2]);
-        if (sigSize <= 0 || expSize <= 0)
-          throw new FormatException("Significand and Exponent sizes must be greater than 0");
-        return new BigFloat(args[1], sigSize, expSize);
-      }
-      else
-      {
-        String[] args = s.Split('_');
-
-        int expIndex = args[2].IndexOf(' ', args[2].IndexOf(' ') + 1);
-        int sigIndex = args[3].IndexOf(' ', args[3].IndexOf(' ') + 1);
-
-        bool isNeg = args[1][3] == '1';
-        BIM exp = BIM.Parse(args[2].Substring(3, expIndex - 3));
-        int expSize = int.Parse(args[2].Substring(expIndex + 1, args[2].Length - expIndex - 4));
-        BIM sig = BIM.Parse(args[3].Substring(3, sigIndex - 3));
-        int sigSize = int.Parse(args[3].Substring(sigIndex + 1, args[3].Length - sigIndex - 3));
-        if (sigSize <= 0 || expSize <= 0)
-          throw new FormatException("Significand and Exponent sizes must be greater than 0");
-
-        if (sig < 0)
-          throw new FormatException("Significand must be greater than 0");
-
-        if (exp < 0)
-          throw new FormatException("Exponent must be greater than 0");
-
-        if (exp >= BIM.Pow(new BIM(2), expSize))
-          throw new FormatException("The given exponent " + exp + " cannot fit in the bit size " + expSize);
-
-        if (sig >= BIM.Pow(new BIM(2), sigSize))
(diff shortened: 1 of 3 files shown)


Source/Basetypes/BigFloat.cs
```diff
@@ -157,6 +157,56 @@ public static BigFloat FromBigDec(BigDec v, int significandSize, int exponentSiz
...
+    [Pure]
+    public static BigFloat FromBVString(String s)
+    {
+      /*
+       * String must be either of the format (fp (bv* *) (bv* *) (bv* *))
+       * or of the special value format: (_ val * *)
+       * Where * indicates an integer value
+       * and val is one of the special values: NaN, +oo, -oo, zero, -zero
+       */
+      BIM sig, exp;
```
zvonimir: Declare variables where they are first used. No need to declare them in the beginning of a function.

Source/Basetypes/BigFloat.cs
```diff
@@ -157,6 +157,56 @@ public static BigFloat FromBigDec(BigDec v, int significandSize, int exponentSiz
...
+
+      isNeg = args[1][3] == '1';
+      exp = BIM.Parse(args[2].Substring(3, expIndex - 3));
+      expSize = int.Parse(args[2].Substring(expIndex + 1, args[2].Length - expIndex - 4));
+      sig = BIM.Parse(args[3].Substring(3, sigIndex - 3));
+      sigSize = int.Parse(args[3].Substring(sigIndex + 1, args[3].Length - sigIndex - 3));
+      if (sigSize <= 0 || expSize <= 0)
+        throw new FormatException("Significand and Exponent sizes must be greater than 0");
+
+      sig = BIM.Abs(sig); //sig must be positive
```
zvonimir: Should you maybe check here that it is indeed positive? And throw an exception otherwise?

Source/Basetypes/BigFloat.cs
```diff
@@ -210,7 +260,7 @@ private BIM maxsignificand()
     [Pure]
     public override string/*!*/ ToString() {
       Contract.Ensures(Contract.Result<string>() != null);
-      return value=="" ? String.Format("{0}x2^{1}", significand.ToString(), Exponent.ToString()) : value;
+      return value == "" ? String.Format("{0}{1}e{2}f{3}e{4}", isNeg ? "-" : "", significand.ToString(), exponent.ToString(), significandSize.ToString(), exponentSize.ToString()) : String.Format("0{0}{1}e{2}", value, significandSize.ToString(), exponentSize.ToString());
```
zvonimir: Please break this into several lines. Maybe also think about splitting it into several statements.

Source/Basetypes/BigFloat.cs
```diff
@@ -157,6 +157,56 @@ public static BigFloat FromBigDec(BigDec v, int significandSize, int exponentSiz
...
+       * Where * indicates an integer value
+       * and val is one of the special values: NaN, +oo, -oo, zero, -zero
+       */
+      BIM sig, exp;
+      int sigSize, expSize;
+      bool isNeg;
+
+      String[] args;
+
+      if (s.IndexOf("_") == 1)
```
zvonimir: Sometimes you use `" "` and sometimes `' '` for strings. I think you mostly use `' '` and so maybe you should update this one.

Source/Basetypes/BigFloat.cs
```diff
@@ -157,6 +157,56 @@ public static BigFloat FromBigDec(BigDec v, int significandSize, int exponentSiz
...
+        s = s.Substring(0, s.Length - 1);
+        args = s.Split(' ');
+        sigSize = int.Parse(args[3]);
+        expSize = int.Parse(args[2]);
+        if (sigSize <= 0 || expSize <= 0)
+          throw new FormatException("Significand and Exponent sizes must be greater than 0");
+        return new BigFloat(args[1], sigSize, expSize);
+      }
+
+      args = s.Split('_');
```
zvonimir: Hmmm, I am confused here. In the comment above you say that format is `(fp (bv* *) (bv* *) (bv* *))`, but then here you are splitting on `_`, which is not even in this format. What am I missing here?

liammachado: The description was wrong, fixed.

Source/Basetypes/BigFloat.cs
```diff
@@ -157,6 +157,57 @@ public static BigFloat FromBigDec(BigDec v, int significandSize, int exponentSiz
...
+       * or of the special value format: (_ val * *)
+       * Where * indicates an integer value
+       * and val is one of the special values: NaN, +oo, -oo, zero, -zero
+       */
+
+      if (s.IndexOf('_') == 1)
+      {
+        s = s.Substring(0, s.Length - 1);
+        String[] args = s.Split(' ');
+        int sigSize = int.Parse(args[3]);
```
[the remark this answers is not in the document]
liammachado: Do you mean replace that line with something like this:

int sigSize = Convert.ToInt32(uint.Parse(args[3]));

The actual variable can't be of type 'uint' because the BigFloat constructor expects integers for the sigSize and expSize parameters. Should this be changed?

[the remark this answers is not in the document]
shaobo-he: Interesting. I think the current implementation is fine. There's no need to change this.

PR #86, opened 2018-03-02
Fixed sign bit not being parsed correctly from floating-point expressions.

liammachado:
The Evaluate method was not parsing out the sign bit correctly from floating-point expressions due to using an incorrect argument. For example, given the expression "(fp (_ bv0 1) (_ bv2 8) (_ bv2 23))", containing the argument "(_ bv0 1)", the "1" would be interpreted as the sign bit. However, the sign bit is actually the "0" portion of the "bv0" string.


zvonimir:
Do we have a regression that catches this problem?
If not, you should create one.

PR #88, opened 2018-03-04
Rounding mode support

liammachado:
This adds support for a type to store rounding modes as described in the **Rounding mode** section of this wiki page:

This also now allows rounding modes to be passed as arguments to the "builtin" functions described in the **SMT-LIBv2 operators via "builtin" attribute** section in the above wiki.


Source/Basetypes/RoundingMode.cs
```diff
@@ -0,0 +1,79 @@
...
+      if (s.Equals("RTN") || s.Equals("roundTowardNegative")) return RTN;
+      if (s.Equals("RTZ") || s.Equals("roundTowardZero")) return RTZ;
+      throw new FormatException(s + " is not a valid rounding mode.");
+    }
+
+    [Pure]
+    [Reads(ReadsAttribute.Reads.Nothing)]
+    public override bool Equals(object obj)
+    {
+      if (obj == null)
```
shaobo-he: I think the if statements can be reduced to the following,

``` c#
RoundingMode rm = obj as RoundingMode;
return rm != null && this == rm;
```

shaobo-he:
@liammachado I think we need more tests to demonstrate the effects of changing rounding mode. For example, for certain values, operations on these values with different rounding modes produce different results. And we'd better have tests that show operations such as `+`, `-` are not affected by rounding mode changes.

zvonimir:
This is kind of crazy, but what happens if you quantify over rounding modes using forall or exists?

zvonimir:
@shaobo-he could we translate some of the SV-COMP benchmarks that change rounding mode manually into Boogie?

shaobo-he:
@zvonimir As a sort, rounding modes can be quantified. For example,

```boogie
function {:builtin "fp.add"} ADD(rmode, float24e8, float24e8) returns (float24e8);

procedure foo()
{
  assert (forall r1: rmode, r2: rmode, f1:float24e8, f2: float24e8 :: r1 != r2 ==> ADD(r1, f1, f2) == ADD(r2, f1, f2));
}
```

Boogie reports that it does not hold.

Furthermore, I've added a couple of regression tests based on SV-COMP benchmarks.

PR #91, opened 2018-03-09
Support for "+zero" and "-zero" special values

liammachado:
This adds support for the user to use the special constant syntax for "+zero" and "-zero", as described in the **Constant declarations** section of this wiki:

### DIFF
--- a/Test/floats/SpecialValues.bpl
+++ b/Test/floats/SpecialValues.bpl
@@ -21,6 +21,11 @@ procedure foo() returns (r : float24e8) {
   assert DEQ(d,-0e2047f53e11);
   d := -200e2000f53e11;
 
+  d := 0+zero53e11;
+  assert d == 0e0f53e11;
+  d := 0-zero53e11;
+  assert d == -0e0f53e11;
+
   return;
 }
 
--- a/Test/floats/SpecialValues.bpl
+++ b/Test/floats/SpecialValues.bpl
@@ -25,6 +25,7 @@ procedure foo() returns (r : float24e8) {
   assert d == 0e0f53e11;
   d := 0-zero53e11;
   assert d == -0e0f53e11;
+  assert DEQ(0+zero53e11, 0-zero53e11);
 
   return;
 }
--- a/Source/Basetypes/BigFloat.cs
+++ b/Source/Basetypes/BigFloat.cs
@@ -118,7 +118,7 @@ namespace Microsoft.Basetypes
     public static BigFloat FromString(String s) {
       /*
        * String must be either of the format *e*f*e*
-       * or of the special value formats: 0NaN*e* 0nan*e* 0+oo*e* 0-oo*e*
+       * or of the special value formats: 0NaN*e* 0nan*e* 0+oo*e* 0-oo*e* 0+zero*e* 0-zero*e*
        * Where * indicates an integer value (digit)
        */
       BIM sig, exp;
@@ -126,9 +126,19 @@ namespace Microsoft.Basetypes
       bool isNeg;
 
       if (s.IndexOf('f') == -1) {
-        String val = s.Substring(1, 3);
-        sigSize = int.Parse(s.Substring(4, s.IndexOf('e')-4));
-        expSize = int.Parse(s.Substring(s.IndexOf('e') + 1));
+        String val;
+        int expIndex = s.IndexOf('e');
+        if (s[2] == 'z') //+zero, -zero
+        {
+          val = s.Substring(1, 5);
+          expIndex = s.IndexOf('e', expIndex + 1);
+        }
+        else //NaN, nan, +oo, -oo
+        {
+          val = s.Substring(1, 3);
+        }
+        sigSize = int.Parse(s.Substring(val.Length + 1, expIndex - (val.Length + 1)));
+        expSize = int.Parse(s.Substring(expIndex + 1));
         if (sigSize <= 0 || expSize <= 0)
           throw new FormatException("Significand and Exponent sizes must be greater than 0");
         return new BigFloat(val, sigSize, expSize);
@@ -451,7 +461,7 @@ namespace Microsoft.Basetypes
       get {
         if (value == "")
           return false;
-        return (value.Equals("NaN") || value.Equals("+oo") || value.Equals("-oo") || value.Equals("zero") || value.Equals("-zero"));
+        return (value.Equals("NaN") || value.Equals("+oo") || value.Equals("-oo") || value.Equals("+zero") || value.Equals("-zero"));
       }
     }
 
--- a/Source/Core/BoogiePL.atg
+++ b/Source/Core/BoogiePL.atg
@@ -149,7 +149,9 @@ TOKENS
 										| '0' 'N' 'a' 'N' digit {digit} 'e' digit {digit}
 										| '0' 'n' 'a' 'n' digit {digit} 'e' digit {digit}
 										| '0' '+' 'o' 'o' digit {digit} 'e' digit {digit}
-										| '0' '-' 'o' 'o' digit {digit} 'e' digit {digit} .
+										| '0' '-' 'o' 'o' digit {digit} 'e' digit {digit}
+										| '0' '+' 'z' 'e' 'r' 'o' digit {digit} 'e' digit {digit}
+										| '0' '-' 'z' 'e' 'r' 'o' digit {digit} 'e' digit {digit} .
 
 COMMENTS FROM "/*" TO "*/" NESTED
 COMMENTS FROM "//" TO lf
--- a/Source/Core/Scanner.cs
+++ b/Source/Core/Scanner.cs
@@ -267,42 +267,42 @@ public class Scanner {
 		for (int i = 65; i <= 90; ++i) start[i] = 2;
 		for (int i = 94; i <= 122; ++i) start[i] = 2;
 		for (int i = 126; i <= 126; ++i) start[i] = 2;
-		for (int i = 49; i <= 57; ++i) start[i] = 45;
+		for (int i = 49; i <= 57; ++i) start[i] = 57;
 		for (int i = 34; i <= 34; ++i) start[i] = 6;
 		start[92] = 1; 
-		start[45] = 87; 
-		start[48] = 46; 
-		start[59] = 52; 
-		start[40] = 53; 
-		start[41] = 54; 
-		start[58] = 88; 
-		start[44] = 55; 
-		start[91] = 56; 
-		start[93] = 57; 
-		start[60] = 89; 
-		start[62] = 90; 
-		start[123] = 58; 
-		start[125] = 91; 
-		start[61] = 92; 
-		start[42] = 93; 
-		start[124] = 94; 
-		start[8660] = 61; 
-		start[8658] = 63; 
-		start[8656] = 64; 
-		start[38] = 65; 
-		start[8743] = 67; 
-		start[8744] = 69; 
-		start[33] = 95; 
-		start[8800] = 73; 
-		start[8804] = 74; 
-		start[8805] = 75; 
-		start[43] = 96; 
-		start[47] = 77; 
-		start[172] = 79; 
-		start[8704] = 82; 
-		start[8707] = 83; 
-		start[955] = 84; 
-		start[8226] = 86; 
+		start[45] = 101; 
+		start[48] = 58; 
+		start[59] = 66; 
+		start[40] = 67; 
+		start[41] = 68; 
+		start[58] = 102; 
+		start[44] = 69; 
+		start[91] = 70; 
+		start[93] = 71; 
+		start[60] = 103; 
+		start[62] = 104; 
+		start[123] = 72; 
+		start[125] = 105; 
+		start[61] = 106; 
+		start[42] = 107; 
+		start[124] = 108; 
+		start[8660] = 75; 
+		start[8658] = 77; 
+		start[8656] = 78; 
+		start[38] = 79; 
+		start[8743] = 81; 
+		start[8744] = 83; 
+		start[33] = 109; 
+		start[8800] = 87; 
+		start[8804] = 88; 
+		start[8805] = 89; 
+		start[43] = 110; 
+		start[47] = 91; 
+		start[172] = 93; 
+		start[8704] = 96; 
+		start[8707] = 97; 
+		start[955] = 98; 
+		start[8226] = 100; 
 		start[Buffer.EOF] = -1;
 
 	}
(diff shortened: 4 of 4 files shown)


Test/floats/SpecialValues.bpl
```diff
@@ -21,6 +21,11 @@ procedure foo() returns (r : float24e8) {
   assert DEQ(d,-0e2047f53e11);
   d := -200e2000f53e11;
 
+  d := 0+zero53e11;
+  assert d == 0e0f53e11;
+  d := 0-zero53e11;
+  assert d == -0e0f53e11;
+
```
shaobo-he: I think it'd better if we add another assertion `DEQ(0+zero53e11,0-zero53e11)` here to test that plus/minus zeros are equal in terms of IEEE standard.

RustanLeino:
Are these new literals just synonyms for what can already be written? For example, I note that the following assertions verify:

    assert 0e0f24e8 == 0+zero24e8;
    assert -0e0f24e8 == 0-zero24e8;

  Rustan

liammachado:
Yes, like +oo and -oo, they are just synonyms.

RustanLeino:
I'm afraid I'm now more confused than before. What is +oo a synonym for?

I don't understand why we would want +zero as a synonym. Is it that much easier for a human to read? What about other synonyms, like -twentythree?

liammachado:
Ah, sorry for the confusion. +/-oo represents positive/negative infinity. There are five special values, as described in the **Constant declarations** section of the following wiki:

[https://github.com/boogie-org/boogie/wiki/Draft-floating-point-Boogie-language-extension](https://github.com/boogie-org/boogie/wiki/Draft-floating-point-Boogie-language-extension)

The reason we have these is to handle the following value constructors defined in the [SMT-Lib floating-point standard](http://smtlib.cs.uiowa.edu/theories-FloatingPoint.shtml):

```
 ((_ +oo eb sb) (_ FloatingPoint eb sb))
 ((_ -oo eb sb) (_ FloatingPoint eb sb))
 ((_ +zero eb sb) (_ FloatingPoint eb sb))
 ((_ -zero eb sb) (_ FloatingPoint eb sb))
 ((_ NaN eb sb) (_ FloatingPoint eb sb))
```

RustanLeino:
Thank you for the clarification. I was more confused than I had thought. I just spent an hour trying to figure out what our notation actually means. For example, I had thought that we had got rid of the hidden-bit convention, but we're apparently using it. Without the hidden-bit convention, we do need the special syntax for positive and negative zero, so thanks for adding them. After I had figured things out, I found this [IEEE-754 converter](http://binary-system.base-conversion.ro/real-number-converted-from-decimal-system-to-32bit-single-precision-IEEE754-binary-floating-point.php?decimal_number_base_ten=-99&sign=1&exponent=10000101&mantissa=10001100000000000000000), which may be useful to others as well.

Let me offer some thoughts and suggestions.

## Our aim

In the design of floating-point literals in Boogie, we need to be clear about what we are trying to achieve. In particular, are we aiming for a format that's closer to the actual bits stored in the computer's memory or are we aiming for a format that's a little closer to mathematics and that a human more easily can understand?

The former makes sense if the human is not in the loop. If Boogie's floating-point input is usually generated mechanically from some bit patterns, then I'd favor aiming in this direction. The sentence "Where a minus sign is placed at the beginning to represent a sign bit of 1" in the [wiki document](https://github.com/boogie-org/boogie/wiki/Draft-floating-point-Boogie-language-extension) tries to explain what a "minus sign" is to an audience that's assumed to already know what "a sign bit of 1" means. This strongly suggests that we have our heads in the 0's and 1's of the computer and that we are trying to think in terms of bit patterns rather than in terms of more common mathematics. Indeed, the current format where `0e128f24e8` represents what we commonly know as 2, `4194304e128f24e8` represents 3, `2097152e129f24e8` represents 5, and `4587520e133f24e8` represents 99, seems aimed at machines, not humans.

On the other hand, if we think that humans will want to read the Boogie code and get a feel for the numbers, or if humans should even be able to write down some numbers with moderate ease, then I favoring aiming for a format that looks more mathematical. Boogie is an intermediate verification language (IVL), which is primarily targeting mechanically generated input. But one of the virtues that makes an IVL useful is that it is human readable and human editable. The language designer in me thinks it's important that Boogie remains somewhat readable. (For example, we're already writing the mantissa before the exponent, like in mathematics and unlike in the IEEE and SMT-LIB formats.)

In the rest of what I have to say, I will take the latter viewpoint, that is, that we're aiming for a format that's closer to mathematics and somewhat human readable and writable. To make Boogie's floating-point literals easier for a human to understand, I will proceed to suggest three guidelines.

## Guideline: No synonyms

When I previously had noticed that `0e0f24e8` and `0+zero24e8` are synonyms, I was in the mindset that we had dispensed with the hidden-bit convention. I thought the `0e0f24e8` I wrote was 0 * 2^0. I was totally mixed up. Instead, by sheer coincidence, the literal I wrote is converted by Boogie into the bit pattern that also stands for positive zero. If it's peculiar that `0e0f24e8` happens to give 0, it is outright weird that `0e255f24e8` stands for positive infinity (because `0e255f24e8` looks so finite).

We should not do this! For a literal that looks like digits-e-digits-f-digits-e-digits, let's check to make sure what's given represents a legal number and not any of the special literals.

## Guideline: Unbiased exponent

The IEEE-754 format only has bits as input. It needs to interpret a substring of bits as a signed exponent or a special value. Luckily for us, we have a richer syntax at our disposal. For example, we're using `0+oo24e8` to denote positive infinity, without requiring Boogie users to remember which special bit pattern this gives rise to. In the IEEE format, the non-special values in the exponent slot must have a way to represent both negative and positive exponents. This could be done using, for example, a sign bit and an absolute number, a 2's complement bit pattern, or a biased number. The IEEE format uses a biased number. This means that the number represented by the bits in the exponent slot are, arithmetically, `B` more than the exponent represented. The `B` is the bias and it is 2^{w-1} - 1, where w is the width of the exponent slot.

The bias makes it easy to reserve the all-0's and all-1's bit patterns for special values. There is no reason for us to have to encode the exponent in this way. By using an ordinary signed value, our human users will quickly be able to see some properties of the exponent, most prominently its sign, but also something about its magnitude.

So, I think we should change the lexical form of floating point numbers to:

        ['-'] digit {digit} 'e' ['-'] digit {digit} 'f' digit {digit} 'e' digit {digit}

and interpret the given exponent without bias. (Looking at `BoogiePL.arg`, it seems we already accept a minus sign in the exponent. I have no idea what that might mean today.)

## Guideline: Readable mantissa

For most numbers that humans deal with on a daily basis, the mantissa is the most important. Indeed, in daily life, we just write out the number--the omitted exponent is implicitly 0. It's therefore unfortunate that the floating-point format of the mantissa is so complicated to understand for a human. As it currently stands, I definitely have to pull out a calculator to start to understand that `2097152e129f24e8` (or even, with an unbiased exponent, `2097152e2f24e8`) means 5.

The mantissa is hard to understand for several reasons. One is that the numbers tend to be large, much larger than numbers we deal with on a daily basis. Another reason is that the numbers represent the fractional portion only, which makes it hard to see where between 1 and 2 the mantissa lies. (The hidden-bit convention doesn't help here, either.) One more reason is that the format is tied to binary numbers, but we're trying to represent them in decimal.

Let me lead up to my suggestion.

### Binary mantissa

We could change the format of the mantissa to be in binary, not decimal. For this to feel right, we had better prefix it with something like `0b`, lest humans would confuse it for a decimal number. We could also, following Ada, allow `_` characters between symbols for readability. The number 5 above would in this way be written (here, with an unbiased exponent)

        0b10_0000_0000_0000_0000_0000e1f24e8

### Hexadecimal mantissa

Binary numbers get too big. Hex digits are easily converted into binary by the human mind, and they are much easier to read. The number 5 is written

        0x20_0000e1f24e8

The lexical grammar for this would be:

        ['-'] '0' 'x' hexdigit { ['_'] hexdigit} 'e' ['-'] digit {digit} 'f' digit {digit} 'e' digit {digit}

### Hexadecimal mantissa without the hidden-bit convention

Speaking for myself, I'm not fond of applying the hidden-bit convention to Boogie's input. If we were to do away with it, I think the numbers would be less confusing to read. We could still insist that the mantissa be in a normal form, representing a number between 1 and (less than) 2. The grammar is:

        ['-'] '0' 'x' '1' '.' hexdigit {hexdigit} 'e' ['-'] digit {digit} 'f' digit {digit} 'e' digit {digit}

But we mustn't cling to the approach from above where we encoded the 23 non-hidden mantissa bits as a 23-bit number where leading 0's can be omitted and trailing 0's are significant! No, we should treat the 23 bits as what they are--a fractional part of the number. In the fractional part of a number we'd write on paper, leading 0's are significant and trailing 0's can be omitted. Concretely, we previously had the 23 binary digits

        01000000000000000000000

Our mistake in our "flush right" thinking was to write this number with the binary digits

        010_0000_0000_0000_0000_0000

We should instead, in the "flush left" thinking, have written it using the binary digits

        0100_0000_0000_0000_0000_000

It was our flush-right thinking that made the mantissas so large, even when we were just representing numbers like 3 and 5. Hurrah! Now we write 5 as:

        0x1.4e2f24e8

Note that the mantissa is `1.4`, not `1.2` as you might think of you just prepend "`1.`" to our flush-right thinking `20_0000` above.

Another mistake we made before was to represent the mantissa as a decimal number. This makes good sense if the base of the exponent is 10. It also makes okay sense if there is no fractional part. But the combination of using decimal digits to represent the base-2 fractional part leads to the mistaken flush-right approach.

### Hexadecimal mantissa, no hidden-bit convention, hexadecimal base

As one final twist, let's think about the base of the exponent. In the binary IEEE format, it is 2. But if we're already using a hexadecimal number for the mantissa, we might as well use 16 as the base of the exponent. This means that the digit before the hexadecimal point can be any non-zero hex digit, not just `1`. This grammar is

        ['-'] '0' 'x' hexdigit '.' hexdigit {hexdigit} 'e' ['-'] digit {digit} 'f' digit {digit} 'e' digit {digit}

and the number 5 is written

        0x5.0e0f24e8

How's that for natural!

## Summary

By the no-synonym guideline, I suggest we provide only one way to write `+oo`, `-oo`, and `NaN`. By the unbiased-exponent guideline, I suggest we allow the exponent to be given as a signed decimal number. And by the readable-mantissa guideline, I suggest we use a hexadecimal representation of the mantissa with one digit before the hexadecimal point. In other words, the numbers are represented as M * 16^E, where M is a signed hexadecimal number whose absolute value lies between 1 and (less than) 16, and E is a signed decimal number. We can choose to keep the special literals for `+zero` and `-zero`, or we can choose to represent these simply as `0x0.0e0f24e8` and `-0x0.0e0f24e8` (and quite possibly disallowing the synonyms that would use a non-zero exponent for these).

Comments welcome.

  Rustan

zvonimir:
These are great suggestions Rustan! Thanks! There are two things I would like to discuss/point out though.

First, it would be good to compare what you proposed with _hexadecimal floating constants_ as they are being introduced in C++17: http://en.cppreference.com/w/cpp/language/floating_literal
There seems to be a good amount of overlap between the two solutions, and maybe sticking with C++17 would make sense (if we are not losing a lot by doing that).

Second, as a colleague of ours pointed out, what we are still missing is support for specifying subnormal numbers: "For example, you can say that subnormal numbers are written as 0x0.03e-126f24e8. Here, the leading digit is 0 and the exponent has the minimal unbiased value. You may add a rule that whenever the first digit is 0 then the exponent must be minimal."

Thoughts?

RustanLeino:
Thanks, Zvonimir. These are both good points, and both are new to me.

I looked at C++17. The description does a good job of being concise and clear. The design, however, leaves something to be desired. Its hexadecimal floating constants use a hex mantissa with a base-2 exponent. Once you have written a number in hexadecimal and want to normalize it, you have to add 4 to the exponent for every step you move the radix point to the right. For example, `23.a8` normalizes to `2.3a8 e4` in the C++ design, not to `2.3a8 e1`.  This seems unnecessarily complex and unnatural. It's more natural to have the same base for the mantissa and as the base of exponent. So, I think we should interpret the exponent with base 16, not 2 as in C++.

C++ makes the parsing of their hex-float literals unambiguous by letting `p` delimit the mantissa from the exponent. As long as we require the radix point (`.`) to be explicit, we don't have a problem with ambiguity. The `e` that delimits our mantissa from the exponent is the second `e` from the right (for example, the mantissa is `0x2.3a8` and the exponent is `1` in the `float24e8` number `0x2.3a8e1f24e8`).

C++ and other languages also allow the mantissa to be written in decimal. This means that the compiler may need to round it. For example, the decimal `0.3` does not have an exact hexadecimal representation. For Boogie as an intermediate verification language, it seems appropriate that such conversions be done a layer above Boogie. Therefore (and also for the reasons of "flush left" explained in my previous remark), I think we should stick with only allowing hexadecimal mantissas in Boogie.

If we allow only hexadecimal mantissas, it would be unambiguous to drop the `0x` prefix. However, I suggest we keep it, because it will remind Boogie users that the mantissa is in hexadecimal. For example, one might otherwise be temporarily confused and expect `1.5e0f24e8` to mean a "one and a half".

Subnormal numbers! I had completely missed these. Yes, the rule you suggest seems fine. But let me float (pun intended) another proposal to see what you think. In my previous remark, I argued for having no synonyms where a number-looking bit-pattern happens to mean something like `NaN` or `+oo`. But suppose we allow un-normalized numbers. For example, the un-normalized `0x70.0e0f24e8` would mean  the same as the normalized `0x7.0e1f24e8`. If we allow un-normalized numbers, then subnormal numbers don't have to insist on having the smallest possible exponent. In fact, with un-normalized numbers, there is no smallest possible exponent. Users would now be allowed to choose where to put the radix point and to select an appropriate exponent, _provided_ that the number denoted is one that can be represented in the specific floating-point size. We already had such a proviso (at least implicitly)--for example, even the normalized-looking `0x8.0e0f1e1` and `0x1.0e1000000000000000f24e8` were forbidden before, because `0x8.0e0` cannot be represented in the type `float1e1` and `0x1.0e1000000000000000` cannot be represented in the type `float24e8`.

Under the proposal of allowing un-normalized numbers, we might as well also allow plus and minus zero as ordinary numbers, like `0x0.0e0f24e8` and `-0x0.0e0f24e8`, in which case we don't need the special `+zero` and `-zero` literals. In fact, whether or not we insist on literals being normalized, it seems a fine idea to use such ordinary numbers instead of the special `+zero` and `-zero` literals. (We would need to decide if the exponent has to be 0 when writing the numbers zero. If we go with un-normalized numbers, we should allow any exponent--that is, `0x0.0e0f24e8` and `0x0.0e1000000000000000f24e8` are both legal and stand for plus zero.)

What do you think of freely allowing un-normalized numbers (and a hexadecimal mantissa, radix point always explicit, and the decimal exponent interpreted as a power of 16)?

  Rustan

PR #93, opened 2018-04-23
CIVL redesign

bkragl:
Improvements to the CIVL verifier with two major changes.

**Independent atomic action specifications**
Atomic actions are now specified as separate procedures instead of code expressions attached to procedures.

**Pending asynchronous calls**
The refinement checker supports the creation and elimination of pending asynchronous calls (pending asyncs) in atomic actions specifications.


bkragl:
Unfortunately this PR was accidentally "squash merged". Since we would loose the commit history (which I think is very valuable here) I'll revert master back to before the squash commit and create a new PR to create a regular merge commit.

Hopefully this does not create any troubles.

PR #95, opened 2018-06-28
Fixed significand size calculation in Evaluate method

liammachado:
This fixes an issue where Boogie would incorrectly calculate the significand size using the exponent part of floating-point expressions instead of the significand part.


zvonimir:
Could you create a regression for this so that we don't mess it up again in the future?

PR #99, opened 2018-07-30
Yices2 interface for Boogie.

dddejan:
Use with: /proverOpt:SOLVER=Yices2 /useArrayTheory

Works for unquantified fragments, e.g. Arrays + Arithmetic + Bitvectors, and can be more efficient than other solvers.


zvonimir:
I think this would be a great addition to Boogie, and it looks good to me.
But someone else should take a look as well before we merge it.
@bkragl maybe? Or @shuvendu-lahiri?

bkragl:
Generally looks good to me and works for a minimal example, but I have a few comments.

1. What is a good set of benchmarks to try? The examples I was working with all have quantifiers.

2. I guess /useArrayTheory is needed to not get the quantified array axioms injected. However, I'm not sure what else this option does, since it is meant to enable Z3's native "combinatory array logic"/"generalized array theory" and work together with the :builtin attribute. For example, I get a crash for Test/generalizedarray/Maps.bpl.

3. Is it the case that Yices2 does not work well with some Boogie features? For example, Field/ref in Test/bitvectors/bv4.bpl and Test/bitvectors/arrays.bpl, or polymorphism in Test/bitvectors/bv5.bpl. I get a crash for those examples.

4. Test/textbook/DutchFlag.bpl has quantifiers and therefore there is rightfully a prover error in the output, but the second line seems like some incompatibility:
```
Prover error: quantifiers are not allowed in logic ALL
Advisory: Partition SKIPPED because of internal error: unexpected prover output: The given key '579' was not present in the dictionary.
Test/textbook/DutchFlag.bpl(15,11): Verification inconclusive (Partition)

```

dddejan:
We only use Yices2 on problems that reduce to bitvectors and
(non-generalized) arrays, e.g. software BMC. On such examples Yices2 can
sometimes perform much better. I can dig out some simple examples, but I'm
not sure this would be useful. Zvonimir might have a better idea.

I don't know about the extended Boogie features but anything that requires
the following will not work:
- quantifiers,
- generalized arrays
- datatypes

I looked into bv4.bpl for example and, although I don't know what the
meaning of a "Field" is, if I look into the SMT log for Z3, it reduces to
quantifiers so that is not supported.

Improving the error handling would be good, specially for crashing
examples. I don't know enough about Boogie to fix that.

Cheers, Dejan

shazqadeer:
@zvonimir, @bkragl : Shall I approve this PR?

zvonimir:
I guess we should merge it, which will then give a chance for us and others to play with it, which will hopefully also lead to improvements and fixes.

bkragl:
Yes. I think it can be merged. Maybe the README.md file could be updated to mention the experimental Yices2 support (and under which conditions it can be expected to work/not work).

PR #107, opened 2018-12-30
Updated floating-point syntax

liammachado:
This implements the changes to the floating-point syntax proposed by @RustanLeino [here](https://github.com/boogie-org/boogie/pull/91#issuecomment-378798275) and [here](https://github.com/boogie-org/boogie/pull/91#issuecomment-380968607). Read the linked comments for full details on what has been changed. To clarify some points:

- The +zero and -zero special values have been removed
- The significand may be un-normalized; this allows the same number to be written in multiple ways. For example, the number 10 may be written as `0xA0.0e-1f24e8`, `0xA.0e0f24e8`, `0x0.Ae1f24e8`, `0x0.0Ae2f24e8`, etc. (However, the +/-infinity and NaN special values can only be written in a single way: `0+oo[sigSize]e[expSize]`, `0-oo[sigSize]e[expSize]`, and `0NaN[sigSize]e[expSize]`, respectively)
- The significand must have trailing zeros such that the last nibble is fully included. For example, in order to represent a floating-point value that has a 24-bit significand with the bit pattern `1_0000_0000_0000_0000_0000_001` (including the hidden bit at the beginning) and an exponent of 0, it may be written as `0x1.000002e0f24e8`, but not `0x1.000001e0f24e8`.

The [wiki page](https://github.com/boogie-org/boogie/wiki/Draft-floating-point-Boogie-language-extension) has been updated to account for these changes.
The original floating-point syntax is specified [here](https://github.com/boogie-org/boogie/wiki/Draft-floating-point-Boogie-language-extension/10fae691dc8f7b58564dc98457c5c74fc0a27707).


zvonimir:
@RustanLeino : It would be helpful if you could review these changes as well. Thanks!

zvonimir:
@liammachado : could you please update the wiki page accordingly as well? Also, it would be great to squash all the commits I think, and use your comment on this pull request as your commit message. Thx!

classner:
Hi!

I just found the floating point syntax update for Boogie and am very interested in it! Thanks for pushing this! :)

I just checked out the liammachado:improved_fp_syntax branch to play around with the FP implementation a bit. For examples combining to floating point values, the verification takes very long (so long I usually have to kill it early), e.g.:

```
procedure Exp1(a: float24e8, b:float24e8)
{
  var res: float24e8;
  assume(a > 0x1.0e0f24e8);
  assume(a < 0x4.0e5f24e8);
  assume(b > 0x1.0e0f24e8);
  assume(b < 0x4.0e5f24e8);
  res := a + b;
  assert(res > a);
  return;
}
```

Is that normal and intended? When using z3 directly to verify this using its `FP` syntax, it terminates very quickly.

zvonimir:
Thanks for trying it out @classner and for your feedback! I think it would be good to create a separate issue with your observations, so that we can continue this discussion there. Such a large discrepancy in runtimes is not something I would expect to see, but I do have a hunch why it is happening. More on that after you create this issue. Thanks!

zvonimir:
@liammachado : just realized one thing. Is there a regression that tests for and explains all the special values and corner cases? For example, `0x1.0e32f24e8` should throw a syntax error. Maybe you already have that, but if not, it would be good to add it.

zvonimir:
Hey @RustanLeino, @bkragl, @shazqadeer. Could one of you do an additional review of @liammachado's changes to the implementation of floating-point constants? Thanks!

shazqadeer:
I think Rustan should have the final word on approving this pull request.

RustanLeino:
Sorry to reenter the party so late. Thanks for the exciting changes!

PR #108, opened 2019-01-07
simplify CIVL logic using C# 7.0 pattern matching

bkragl:
This avoids a lot of casting in the CIVL code, which makes it cleaner and easier to read.

There is no change in functionality, but can somebody check if it compiles on other setups? On my Ubuntu machine running Mono 5.18 everything is fine.

### DIFF
--- a/Source/Concurrency/CivlRefinement.cs
+++ b/Source/Concurrency/CivlRefinement.cs
@@ -70,10 +70,10 @@ namespace Microsoft.Boogie
 
                 if (layerNum > yieldingProc.upperLayer)
                 {
-                    if (yieldingProc is ActionProc)
+                    if (yieldingProc is ActionProc actionProc)
                     {
                         // yielding procedure already transformed to atomic action
-                        var action = ((ActionProc)yieldingProc).refinedAction;
+                        var action = actionProc.refinedAction;
                         if (action.layerRange.Contains(layerNum))
                             return action.layerToActionCopy[layerNum].proc;
                         else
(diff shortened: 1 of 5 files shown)


michael-emmi:
I guess you mean that the C# 7.0 features are supported by .NET 4.5 runtime, right?

But does that imply that C# 7.0 code can be *compiled* by toolchains that support the 4.5 target framework?

For instance, looking at the Mono release notes, it seems that C# 7 support starts with [5.x](http://www.mono-project.com/docs/about-mono/releases/5.0.0/), but Ubuntu 14.04 comes with an older version of Mono — 4.2 I believe — which supports .NET target framework 4.5. While you’re probably right that this older version can run code compiled with C# 7, it seems that it can’t actually compile the C# 7 code.

Given that there is no Boogie binary distribution, perhaps pushing the requirement to C# 7.0 should be more carefully considered, with breaking Travis builds based on Mono 4.2 in mind?

bkragl:
Aah, it's about compilation. Sure, a compiler cannot compile syntactic sugar it does not know.
Bottom line: I'm sorry if this commit caused any trouble and if the consensus is that C# 7 should not be used it can be reverted. However, I have a few thoughts:

1. Which build is it that breaks? It seems that the Travis build of Boogie itself runs fine on Ubuntu 14.04. In particular, I'm seeing Mono 5.20 in the build log. Can this other build perhaps be upgraded?
2. Boogie still uses the xbuild tool which was deprecated in [Mono 5](https://www.mono-project.com/docs/about-mono/releases/5.0.0/#msbuild) and will be removed in future updates in favor of MSbuild. I think switching away from deprecated tools is a good idea. For example, Dafny now uses MSbuild and thus requires a sufficiently recent Mono version.
3. Why are there no Boogie binary releases?

michael-emmi:
I’m not suggesting that Boogie should *not* require C# 7!

I simply wondered whether an explicit (collective) decision was made — seems not.

I don’t know how many this may bite eventually, but @dddejan noticed this with a Travis CI build on Ubuntu 14.04 system from a C++-language image, which apparently has Mono < 5.x — he’s using Boogie in a much larger (mainly C++) project, so the C#-language Travis image with Mono 5.20 is not relevant there. The fix involves installing a newer version of Mono, which seems feasible in his case, but perhaps not for everyone.

Anyhow, I agree with you, that ultimately this whole obstacle seems silly, since at this point, i.e., after moving to .NET Core, Boogie could easily distribute a [.NET Core global tool](https://docs.microsoft.com/en-us/dotnet/core/tools/global-tools) with NuGet. Then getting Boogie would boil down to fetching a pre-compiled package via `dotnet tool install -g boogie --version X`. This is the best solution.

Issue #81, opened 2017-12-12
Switching Boogie build to .NET Standard/Core


zvonimir:
Recently I started looking into how to improve the portability of Boogie across platforms, and in particular whether it can be compiled using Microsoft's dotnet core distribution for Linux. After reading a bit about various .NET implementations, it seems to me that it would be good to switch Boogie from .NET Framework v4.5 (which it currently requires) to something that is more portable, such as .NET standard/core. (Note that .NET Framework is not available on Linux from Microsoft, while .NET Core is - see https://www.microsoft.com/net/download/linux)

Based on my understanding, switching to .NET Standard/Core ought to be possible (and hopefully not too hard) since Boogie is a command line application that does not use GUI.

Note that I was quite confused about what the heck .NET Framework, Core, and Standard are, and this is a good place to start learning about this: https://msdn.microsoft.com/en-us/magazine/mt842506.aspx

Anyhow, I would like to start a discussion regarding this switch, and maybe get opinions and/or suggestions from people who understand the potential issues and how to go about this better.

RustanLeino:
Boogie is also used by Dafny, including from within the Dafny plug-in for Visual Studio. Will that still work with .NET Core?

zvonimir:
I suspect that in this scenario Boogie is being invoked as a library. My understanding is that should still work. In fact, .NET Core is mentioned as the best option for implementing libraries if one is interested in portability. If there are GUI aspects within Boogie, then that would probably create problems.
Having said that, I am certainly not an expert on these .NET portability issues/frameworks, and it would be good to get some input from someone who has a better understanding of this landscape and options. Which is why I created this issue.

shazqadeer:
I believe this issue has been addressed by the work done by @michael-emmi and @bkragl.  Boogie now builds with .NET Core and Standard.  The travis build is .NET core.  So closing the issue now.  Please reopen if necessary explaining what remains unaddressed.

PR #124, opened 2019-05-09
Dotnet core tool

michael-emmi:
This patch packages up Boogie as a [.NET Global Tool](https://docs.microsoft.com/en-us/dotnet/core/tools/global-tools) which runs on .NET Core. This also addresses Issue #81.

The tool is packaged like so:
```bash
$ dotnet pack Source/BoogieDriver/BoogieDriver-NetCore.csproj --version-suffix=netcore-alpha
```

If published with NuGet, you’d install the tool so:
```bash
$ dotnet tool install -g Boogie --version 2.4.1-netcore-alpha
```

Or install from a local build, like so:
```bash
$ dotnet tool install -g --add-source Source/BoogieDriver/nupkg Boogie --version 2.4.1-netcore-alpha
```

Once installed, you’d run Boogie like so:
```bash
$ boogie /z3exe:$(which z3) a.bpl
```

Note that the packaging scheme above does not include Z3, which is platform dependent, and thus the Z3 executable must be provided explicitly on the command line.


shazqadeer:
All changes appear to be new files.  Excellent improvement in Boogie portability.  Thanks a lot, @michael-emmi.

emptylambda:
Hello Michael, 

thanks for the .Net core support! 
just to ask a naive clarification question:
with the new dotnet cli, the procedure for a complete build will be: 
1. `dotnet build` (in this case `BoogieDriver-NetCore.csproj`)
2. `dotnet pack`
3. `dotnet update` (provided already invoked `dotnet install` before)

michael-emmi:
I’m not familiar with the `dotnet update` command, but I do know that [`dotnet pack` includes building the project(s) by default](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-pack?tabs=netcore2x), and [`dotnet build` can generate packages as well](https://github.com/boogie-org/boogie/blob/21d9cd6a9bee7ad103defbd248ab37fb7529cbd8/Source/BoogieDriver/BoogieDriver-NetCore.csproj#L16); see [the documentation](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-build).

bkragl:
I'm wondering what the recommended/standard way of executing Boogie under .NET Core is for developers. I ran the regression tests using two variants:

1. Running `dotnet run` on the C# project (I supply `--no-build` because otherwise a rebuild is triggered occasionally, causing extra output to fail tests):
`dotnet run --no-build --project $REPO_ROOT/Source/BoogieDriver/BoogieDriver-NetCore.csproj`
2. Running `dotnet` directly on the built DLL:
`dotnet $REPO_ROOT/Source/BoogieDriver/bin/Debug/netcoreapp2.2/BoogieDriver.dll`

I think variant 1 would be nice. However, I noticed that it is consistently slower (~140 seconds) compared to both variant 2 and the Mono build (~80 seconds). These numbers come from an 8-core Ubuntu 18.04 machine. Any idea why this is the case?

Can variant 2 be done without explicitly writing `bin/Debug/netcoreapp2.2`? Or is there perhaps a different way of running .NET Core apps that I'm not aware of?

Also, two regression tests fail:
- `test15/CaptureState.bpl` (two lines in the model output are swapped, which I guess is benign)
- `snapshots/runtest.snapshot` (it seems that not all snapshots are processed)

michael-emmi:
I don’t know what kinds of actions might happen in `dotnet run --no-build` before actually executing the assembly, but I can imagine that extra actions could happen. I would think that invoking the generated assembly directly (e.g., Option 2) would be the best. What bothers you about Option 2 — is it the `dotnet` interpreter part, the path to assembly, or something else? Personally, I would run installed tool, i.e., `~/.dotnet/tools/boogie`, which is in the path, assuming .NET Core is correctly set up, i.e., `~/.dotnet/tools` is in the path. You can install from a locally-built package — it doesn’t have to come from NuGet.

bkragl:
Yes, it's the path to assembly that bothers me in Option 2. I wouldn't want to write it like this into `Test/lit.site.cfg` because it's specific to the .NET Core version.

I was thinking that, as a developer, I want to avoid the installation step for running the tests (and not rely on a correct system setup). Also, I might want to have a stable Boogie version installed that should not be overwritten while coding.

Perhaps a solution could be to have the build output of the BoogieDriver project also written to the `Binaries` directory, as it is the case for the Mono build?

michael-emmi:
Oh I see — and agree completely.

Seems like it is all possible too. When I comment out [this line](https://github.com/boogie-org/boogie/blob/21d9cd6a9bee7ad103defbd248ab37fb7529cbd8/Source/BoogieDriver/BoogieDriver-NetCore.csproj#L18) and run the following:
```
$ dotnet build -r osx.10.14-x64 Source/BoogieDriver/BoogieDriver-NetCore.csproj
```
a native executable gets generated:
```
$ file ./Source/BoogieDriver/bin/Debug/netcoreapp2.2/osx.10.14-x64/BoogieDriver
./Source/BoogieDriver/bin/Debug/netcoreapp2.2/osx.10.14-x64/BoogieDriver: Mach-O 64-bit executable x86_64
```
Of course, the executable will depend on the target system’s runtime identifier (RID), and here is the [RID catalog](https://docs.microsoft.com/en-us/dotnet/core/rid-catalog).

Issue #125, opened 2019-05-09
Inlining and verification in spec mode


hajduakos:
I would like inlined methods to be also verified "on their own". The help displayed by the `/attrHelp` flag states that "methods with the attribute {:inline N} are not verified". However, using `/inline:spec` seems to both inline the method and verify it on its own.

Consider the following example:

```
var x : int;

procedure {:inline 1} inlinedproc()
    requires x >= 0;
    ensures x >= 0;
{
    x := x + 1;
}

procedure normalproc()
{
    x := 0;
    call inlinedproc();
    assert(x >= 1);
}
```

Running Boogie with `/doModSetAnalysis /inline:spec /trace` outputs the following:
```
Verifying inlinedproc ...
  [0.146 s, 1 proof obligation]  verified

Verifying normalproc ...
  [0.010 s, 3 proof obligations]  verified
```

There must be inlining, otherwise the assertion in `normalproc` could not be proved. However, the output indicates that `inlinedproc` was also verified on its own.

Is this a feature that can be relied on, or rather a bug? If it's a bug, creating a copy of the function without inlining could work, but it would be still nice to have it as a feature.

shazqadeer:
I looked at the code and it looks like there are three inlining options: /inline:{none, assert, assume, spec}.  Verification is skipped only in the case of /inline:{assert, assume}. See the method SkipVerification in Absy.cs.  Your understanding of /inline:spec appears to be correct.

hajduakos:
Issue solved by PR #126

PR #126, opened 2019-05-10
Update command line help to match implementation

hajduakos:
Based on the command line help, inlined procedures are not verified in `assert`, `assume` and `spec` modes. However, based on the implementation, they are not verified in only `assert` and `assume` modes (see #125).


RustanLeino:
Yes, the updated documentation seems correct, and I think it makes sense. Essentially, with the assert or assume modes, you're saying you always want the procedure to be inlined, but in case the inlining depth N is reach, you want to know about it (for assert) or you want to ignore it (for assume). In contrast, the spec mode is sound and does some inlining just to be more precise. Soundness of the spec mode requires the procedure body to be checked itself, too.

However, there is still a discrepancy between the documentation and the code, namely when N is 0. In that case, it seems procedure body is always checked. I consider that to be a bug. We could just document that bug as well, but I'd rather it get fixed. I didn't look at the code just now, but I suspect some code that decides whether or not to try verifying a procedure implementation treats {:inline 0} the same as having no inline attribute at all (whereas on the call site, {:inline 0} is properly distinguished from having no inline attribute at all). Can you please find this place and fix it?

hajduakos:
@RustanLeino I think I found and fixed that condition in 9f83404. The procedure was skipped from verification (in assume/assert modes) if there was an `{:inline n}` attribute with `n>0`. As you suspected, this caused `{:inline 0}` to be equivalent to having no attribute, i.e., always verifying the procedure. Simply removing the `n>0` part of the check seems to make Boogie behave as you described (see also the test case).

PR #128, opened 2019-05-24
Fixes bugs and cleans code in BigFloat.cs

liammachado:
This pull request does the following:
-Removes many unused properties, methods, and constructors in `BigFloat.cs`
-Fixes the implementation of the `+, *, ==, !=, <=, >=, <, >` operators, the `IsZero` property, and the `FloorCeiling` and `CompareTo` methods
-The two_n method was removed and replaced with calls to `BigInteger.Pow(2, n)`
-The entire class was reformatted to have a consistent code style; changes
include added braces in if/else blocks and inserted spaces after casts
-Added comment in front of `significandSize` field
-Modified comment above `BigFloat` struct to reflect that it can have arbitrary
exponent and significand sizes
-Changed name of `isNeg` field to `isSignBitSet`

### DIFF
--- a/Source/Basetypes/BigFloat.cs
+++ b/Source/Basetypes/BigFloat.cs
@@ -317,27 +317,20 @@ namespace Microsoft.Basetypes
     // Basic arithmetic operations
 
     [Pure]
-    public BigFloat Negate {
-      get {
-        if (value != "") {
-          if (value[0] == '-') {
-            return new BigFloat("+oo", significandSize, exponentSize);
-          }
-
-          if (value[0] == '+') {
-            return new BigFloat("-oo", significandSize, exponentSize);
-          }
-
-          return new BigFloat("NaN", significandSize, exponentSize);
+    public static BigFloat operator -(BigFloat x) {
+      if (x.value != "") {
+        if (x.value[0] == '-') {
+          return new BigFloat("+oo", x.significandSize, x.exponentSize);
         }
 
-        return new BigFloat(!isSignBitSet, significand, exponent, significandSize, exponentSize);
-      }
-    }
+        if (x.value[0] == '+') {
+          return new BigFloat("-oo", x.significandSize, x.exponentSize);
+        }
 
-    [Pure]
-    public static BigFloat operator -(BigFloat x) {
-      return x.Negate;
+        return new BigFloat("NaN", x.significandSize, x.exponentSize);
+      }
+
+      return new BigFloat(!x.isSignBitSet, x.significand, x.exponent, x.significandSize, x.exponentSize);
     }
 
     [Pure]
@@ -429,7 +422,7 @@ namespace Microsoft.Basetypes
 
     [Pure]
     public static BigFloat operator -(BigFloat x, BigFloat y) {
-      return x + y.Negate;
+      return x + -y;
     }
 
     [Pure]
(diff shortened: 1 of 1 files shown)


Source/Basetypes/BigFloat.cs
```diff
@@ -523,73 +434,158 @@ public void FloorCeiling(out BIM floor, out BIM ceiling)
...
+
+        return cmpThis;
+      }
+
+      if (value == that.value) {
         return 0;
-      return this.significand > that.significand ? 1 : -1;
+      }
+
+      if (value == "NaN" || that.value == "+oo" || value == "-oo" && that.value != "NaN") {
```
zvonimir: Wait, what if `that.value == "NaN"`?

Source/Basetypes/BigFloat.cs
```diff
@@ -523,73 +434,158 @@ public void FloorCeiling(out BIM floor, out BIM ceiling)
...
+        return -1;
+      }
+
+      return 1;
     }
 
     [Pure]
     public static bool operator ==(BigFloat x, BigFloat y) {
+      if (x.value == "NaN" || y.value == "NaN") {
```
zvonimir: Why do you need these checks for `NaN` here?
To me it seems that your intention was to handle special cases in the `compareTo` method as well.

Source/Basetypes/BigFloat.cs
liammachado: If we've reached this if statement and that.value == "NaN", then we should return 1. All values, including -infinity, are greater than NaN in the CompareTo method. However, note that the expression `-infinity > NaN` should return false. See my other comment for more details.

Source/Basetypes/BigFloat.cs
liammachado: `CompareTo` isn't supposed to return results that are consistent with what the `==, !=, >, <, >=, <=` operators return. Specifically, if any of the 6 operators are given a `NaN` value for at least one of their operands, they should always return false. However, I think `BigFloat.CompareTo` should follow the specification for `Single.CompareTo`, described here:

https://docs.microsoft.com/en-us/dotnet/api/system.single.compareto?view=netframework-4.8

As described in that document, `NaNs` are treated differently than for the 6 operators. For example, the expression `(0.0f / 0.0f).CompareTo(0.0f / 0.0f)` should return true, whereas the expression `(0.0f / 0.0f) == (0.0f / 0.0f)` should return false.

liammachado: Except `!=`, which should always return true if `NaN` is passed.

zvonimir: Ah, great, I get it, thanks for clarifying this. It would be useful to add what you just wrote here (a shorter summary of it) as a comment on the `CompareTo` method.

liammachado: Alright, I've added this.

shaobo-he:
@liammachado I have a general question: how do you decide which methods should be provided or not?

liammachado:
I decided that simply based on whether each method was used or not. Every method remaining in the class is used somewhere in the Boogie project. Of the things I removed, the vast majority were not used at all. The remaining removed items were used, but either had trivial implementations (e.g. the `IsSpecialType` property) or could be replaced with methods built into C# (e.g. `two_n`).

shaobo-he:
LGTM.

Issue #110, opened 2019-01-17
Error when parsing models with maps of floating point values


classner:
This is another issue related to PR #107 concerning models for maps with floats. It seems boogie can't parse model outputs (at least from) z3 for this case. For example, this program obviously fails to verify:

```
procedure Exp2(a:[int]float24e8)
{
  assert a[0] > 0x0.0e0f24e8;
}
```

When running boogie just as `%boogie Exp2.bpl`, it correctly states that the assertion might not hold and terminates. However, when adding `/printModel:4`, it raises an error, stating

```
Prover error: Model parsing error: Invalid model: unidentified line, at line 2 (a -> |T@[Int](_ FloatingPoint 8 24)!val!0|)
Prover error: Could not parse any models
```

Adding `/printModelToFile:out.txt` also doesn't help, since it apparently wants to parse the model first.

zvonimir:
@liammachado : please take a look at this one. Maybe also create a regression for it.

liammachado:
I don't think that this error is being caused by PR #107, since I am receiving the same error with the master branch.

classner:
Sorry @liammachado, I thought only #107 was integrating the floating point syntax (I'm new to the project).

zvonimir:
@liammachado Please take a look at this problem even if it was not caused by your recent pull request. We should try to fix it no matter what. It is unclear to me whether parsing of models is not working for floating-points at all, or is it something specific about this benchmark.

shazqadeer:
@zvonimir : Any updates?

zvonimir:
I'll look into this more carefully next week.

zvonimir:
@shaobo-he : Please take a look into this and report back your findings. We should get this fixed.

shazqadeer:
Neither @liammachado nor @shaobo-he has responded.

@zvonimir : Should we close this issue going on record that we won't fix the problem now?

shaobo-he:
> Neither @liammachado nor @shaobo-he has responded.
> 
> @zvonimir : Should we close this issue going on record that we won't fix the problem now?

I recalled that I concluded it's not going to be a trivial fix. How about letting me try to fix it this week? If I can't come up with anything by Monday, we'll close this issue.

PR #134, opened 2019-06-25
Evaluate array expressions from Z3 4.8.5.

michael-emmi:
This PR updates support for parsing array expressions from Z3 4.8.5 models to generate counterexample-trace values. For instance, 

```
$ cat test.bpl && boogie /stratifiedInline:1 /useProverEvaluate test.bpl 
type Ref;

procedure boogie_si_record_ref(x:Ref);
procedure boogie_si_record_map(x:[int]int);

{
   var x, y: int;
   var w, z: Ref;
   var M: [int]int;

   x := 5;
   assume x == y;
   M[x] := y;

   assume w != z;

   call {:cexpr "x"} boogie_si_record_int(x);
   call {:cexpr "y"} boogie_si_record_int(y);
   call {:cexpr "w"} boogie_si_record_ref(w);
   call {:cexpr "z"} boogie_si_record_ref(z);

   call {:cexpr "M"} boogie_si_record_map(M);

   assert false;
}

Boogie program verifier version 2.4.1.10503, Copyright (c) 2003-2014, Microsoft.
test.bpl(28,4): Error BP5001: This assertion might not hold.
Execution trace:
    test.bpl(14,6): anon0
    value = 5
    value = 5
    value = T@Ref!val!0
    value = T@Ref!val!1
    value = {"5":5,"*":0}

```

Currently this only handles the basic case of single-index non-nested arrays, which are presented by Z3 like this: `(lambda (() (x!1 Int) (x!2 Int)) (ite (= x!1 5) 5 0))`. TODOs have been left for the other cases.

### DIFF
--- a/Source/Provers/SMTLib/ProverInterface.cs
+++ b/Source/Provers/SMTLib/ProverInterface.cs
@@ -2337,6 +2337,64 @@ namespace Microsoft.Boogie.SMTLib
 
     object GetArrayFromProverResponse(SExpr resp) {
       resp = ReduceLet(resp);
+      var dict = GetArrayFromArrayExpr(resp);
+      if (dict == null) dict = GetArrayFromProverLambdaExpr(resp);
+      if (dict == null) return null;
+      var str = new StringWriter();
+      str.Write("{");
+      foreach (var entry in dict)
+        if (entry.Key != "*")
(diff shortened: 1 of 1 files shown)


zvonimir:
Does this PR maybe fix this https://github.com/boogie-org/boogie/issues/110?

michael-emmi:
I would not think so.

PR #135, opened 2019-06-29
Fixed BigFloat construction of the returned FP constants

shaobo-he:
It turns out the the FP constants returned by the solver may have the following syntax: `(fp (_ BitVec 1) (_ BitVec eb) (_ BitVec i) (_ FloatingPoint eb sb))`, where `i=sb-1`. This PR sets the correct significand sizes.

### DIFF
--- a/Source/Provers/SMTLib/ProverInterface.cs
+++ b/Source/Provers/SMTLib/ProverInterface.cs
@@ -2456,7 +2456,7 @@ namespace Microsoft.Boogie.SMTLib
             BigInteger exp = getBvVal(expExpr);
             int expSize = getBvSize(expExpr);
             BigInteger sig = getBvVal(sigExpr);
-            int sigSize = getBvSize(sigExpr);
+            int sigSize = getBvSize(sigExpr)+1;
             return new Basetypes.BigFloat(isNeg, sig, exp, sigSize, expSize);
         }
         if (resp.Name == "_" && resp.ArgCount == 3)


zvonimir:
Does this pull request for this issue: https://github.com/boogie-org/boogie/issues/110 ??
If not, then can you take a look at it quickly?

shaobo-he:
> Does this pull request for this issue: #110 ??
> If not, then can you take a look at it quickly?

No, it does not solve that issue. I think I looked at it and concluded the fix won't be easy.

zvonimir:
Could you add/create a regression that catches this problem? Based on your fix it is not clear to me how come this worked in the first place.

liammachado:
I think it worked until I made this change in a pull request a few months ago:

https://github.com/boogie-org/boogie/pull/107/files#diff-d9b743913715f1b8d4710b53060422ddL174

I forgot to update ProverInterface.cs accordingly. Thanks for catching this.

zvonimir:
@liammachado : Does the fix look ok to you? Also, could we add a regression that catches this? Or amend an existing regression?

liammachado:
Yes, this fix looks good to me.

zvonimir:
What about suggesting a new regression or changing an existing one that would catch this?

liammachado:
Unfortunately, I'm currently not sure under what circumstances this code is called by Boogie. I remember Corral erroring when this section of code was written incorrectly a couple years ago. A suitable regression for Corral would be any one that includes a floating-point constant such that the significand fits in the old significand size, but not the new one. For example, writing `0x0.FFFFFFe0f24e8` in the regression would be valid, but this method would change it to `0x0.FFFFFFe0f23e8`, which would result in an error. However, while I know that Corral runs this section of code, I don't know how to get Boogie by itself to do so.

PR #136, opened 2019-07-02
Printing a Boogie program after the lambda-lifting phase

amaurremi:
This pull requests allows one to print out a Boogie program after it has undergone lambda lifting using the `/printLambdaLifting` option, which can be useful for debugging. Use as follows:
```
boogie <file>.bpl /printLambdaLifting /print:<outputfile>.bpl
```

### DIFF
--- a/Source/ExecutionEngine/ExecutionEngine.cs
+++ b/Source/ExecutionEngine/ExecutionEngine.cs
@@ -858,10 +858,12 @@ namespace Microsoft.Boogie
       #region Do some pre-abstract-interpretation preprocessing on the program
       // Doing lambda expansion before abstract interpretation means that the abstract interpreter
       // never needs to see any lambda expressions.  (On the other hand, if it were useful for it
-      // to see lambdas, then it would be better to more lambda expansion until after infererence.)
+      // to see lambdas, then it would be better to more lambda expansion until after inference.)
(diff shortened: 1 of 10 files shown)


Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -858,10 +858,13 @@ public static void Inline(Program program)
       #region Do some pre-abstract-interpretation preprocessing on the program
       // Doing lambda expansion before abstract interpretation means that the abstract interpreter
       // never needs to see any lambda expressions.  (On the other hand, if it were useful for it
-      // to see lambdas, then it would be better to more lambda expansion until after infererence.)
+      // to see lambdas, then it would be better to more lambda expansion until after inference.)
       if (CommandLineOptions.Clo.ExpandLambdas) {
         LambdaHelper.ExpandLambdas(program);
-        //PrintBplFile ("-", program, true);
+        if (CommandLineOptions.Clo.PrintFile != null && CommandLineOptions.Clo.PrintLambdaLifting)
+        {
```
RustanLeino: Please use the same brace formatting style as on line 862, that is, with the curly brace at the end of the `if` line.

amaurremi: Done.

PR #140, opened 2019-07-23
Lambda lifting based on maximally large bound-var-free subexpressions

amaurremi:
Currently, Boogie transforms anonymous functions into function calls that return maps. The transformation abstracts over the free variables of a lambda, replacing free variables with bound ones, and passing the original free variables as arguments to the generated lifted function. This commit generalizes this approach by abstracting over larger subexpressions of a lambda that do not contain free variables. Specifically, this lambda lifting algorithm lifts out of the lambda maximally large subexpressions that do not contain any of the lambda's bound variables.

The original lambda-lifting algorithm can be invoked using the `/freeVarLambdaLifting` option.

### DIFF
--- a/Source/Core/AbsyQuant.cs
+++ b/Source/Core/AbsyQuant.cs
@@ -575,8 +575,15 @@ namespace Microsoft.Boogie {
       }
     }
 
+    // todo (MR) this was implemented to throw an error; is this bad?
     public override int GetHashCode() {
-      throw new NotImplementedException();
+      int hash = 17;
+      hash = hash * 23 + tr.GetHashCode();
+      hash = hash * 23 + Pos.GetHashCode();
+      if (Next != null) {
+        hash = hash * 23 + Next.GetHashCode();
+      }
+      return hash;
     }
   }
 
--- a/Source/Core/CommandLineOptions.cs
+++ b/Source/Core/CommandLineOptions.cs
@@ -416,6 +416,7 @@ namespace Microsoft.Boogie {
     public int DoomStrategy = -1;
     public bool DoomRestartTP = false;
     public bool PrintDesugarings = false;
+    public bool FreeVarLambdaLifting = false;
     public string SimplifyLogFilePath = null;
     public bool PrintInstrumented = false;
     public bool InstrumentWithAsserts = false;
@@ -1650,7 +1651,8 @@ namespace Microsoft.Boogie {
               ps.CheckBooleanFlag("verifySeparately", ref VerifySeparately) ||
               ps.CheckBooleanFlag("trustAtomicityTypes", ref TrustAtomicityTypes) ||
               ps.CheckBooleanFlag("trustNonInterference", ref TrustNonInterference) ||
-              ps.CheckBooleanFlag("useBaseNameForFileName", ref UseBaseNameForFileName)
+              ps.CheckBooleanFlag("useBaseNameForFileName", ref UseBaseNameForFileName) ||
+              ps.CheckBooleanFlag("freeVarLambdaLifting", ref FreeVarLambdaLifting)
               ) {
             // one of the boolean flags matched
             return true;
@@ -1910,6 +1912,13 @@ namespace Microsoft.Boogie {
                    identifies variables
   /printUnstructured : with /print option, desugars all structured statements
   /printDesugared : with /print option, desugars calls
+
+  /freeVarLambdaLifting : Boogie's lambda lifting transforms the bodies of lambda
+                         expressions into templates with holes. By default, holes
+                         are maximally large subexpressions that do not contain
+                         bound variables (see Section 8 of Boogie 3 paper).
+                         This option performs a form of lambda lifting
+                         in which holes are the lambda's free variables. 
 
   /overlookTypeErrors : skip any implementation with resolution or type
                         checking errors
--- a/Source/Core/LambdaHelper.cs
+++ b/Source/Core/LambdaHelper.cs
@@ -3,13 +3,13 @@
 // Copyright (C) Microsoft Corporation.  All Rights Reserved.
 //
 //-----------------------------------------------------------------------------
+
+using Core;
+
 namespace Microsoft.Boogie {
 
   using System;
-  using System.IO;
-  using System.Collections;
   using System.Collections.Generic;
-  using System.Diagnostics;
   using System.Diagnostics.Contracts;
   using Set = GSet<object>;  // for the purposes here, "object" really means "either Variable or TypeVariable"
 
@@ -37,6 +37,38 @@ namespace Microsoft.Boogie {
       return program;
     }
 
+    /// <summary>
+    /// Performs lambda lifting on all anonymous functions in a program.
+    /// Lambda lifting transforms a lambda abstraction <c>(lambda x : T :: t)</c> of type <c>T -> U</c>
+    /// into a Boogie map of type <c>[T] U</c> as follows:
+    /// <list type="number">
+    /// <item>
+    ///   Certain subexpressions <c>e1</c>, ..., <c>e_n</c> (which have types <c>T1</c>, ..., <c>T_n</c>)
+    ///   of the lambda body [t] are replaced with bound variables <c>b1</c>, ..., <c>b_n</c> of the same types,
+    ///   yielding a new lambda body [u], which represents a lifted lambda <c>(lambda x : T :: u)</c>.
+    /// </item>
+    /// <item>
+    ///   The original lambda is replaced with a function call <c>f(e1, ..., en)</c> where <c>f</c> is a function
+    ///   with <c>n</c> parameters and return type <c>[T] U</c> (i.e. the function returns a map).
+    /// </item>
+    /// <item>
+    ///   The implementation of the lambda is defined through an axiom that states that
+    ///   <c>f(b1, ..., b_n)[x] == u</c>.
+    /// </item>
+    /// </list>
+    /// Lambda lifting algorithms differ based on what kind of subexpressions <c>E_i</c> are "lifted" out
+    /// of the lambda:
+    /// <list type="bullet"></list>
+    /// <item>
+    ///   <see cref="LambdaVisitor.LambdaLiftingFreeVars"/> lifts all free variables of a lambda;
+    /// </item>
+    /// <item>
+    ///   <c>LambdaVisitor.LambdaLifterMaxHoles()</c> lifts a lambda's maximally large subexpressions that are
+    ///   free of bound variables.
+    /// </item>
+    /// <see cref="LambdaVisitor.LambdaLifterMaxHoles"/> is used by default whereas <c>LambdaLiftingFreeVars</c>
+    /// is used with the command-line option <c>/freeVarLambdaLifting</c>.
+    /// </summary>
     public static void ExpandLambdas(Program prog) {
       Contract.Requires(prog != null);
       List<Expr/*!*/>/*!*/ axioms;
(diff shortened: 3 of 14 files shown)


Source/Core/CommandLineOptions.cs
```diff
@@ -1911,6 +1913,13 @@ identifies variables
   /printUnstructured : with /print option, desugars all structured statements
   /printDesugared : with /print option, desugars calls
 
+  /freeVarLambdaLifting : Boogie's lambda lifting transforms the bodies of lambda
+                         expressions into templates with holes. By default, holes
+                         are maximally large subexpressions that do not contain
+                         bound variables (see Section 8 of Boogie 3 paper).
```
RustanLeino: Leave out the parenthetic remark, since the paper you're referring to is not available.

Source/Core/AbsyQuant.cs
```diff
@@ -575,8 +575,15 @@ public Trigger(IToken/*!*/ tok, bool pos, IEnumerable<Expr>/*!*/ tr, Trigger nex
       }
     }
 
+    // todo (MR) this was implemented to throw an error; is this bad?
```
RustanLeino: Your new code is Good. It appears that it was not needed before, because then there had been no need for a map from "holes" to expressions. (A possible alternative, but I'm not pressing for it, might have been to have iterated through the triggers (via their `.Next` field) and to have visited their expressions lists `.tr`. That might have avoided the need for putting the triggers themselves into the hashtable. But the way you changed it is fine.)

Source/Core/LambdaHelper.cs
```diff
@@ -52,6 +84,7 @@ public static class LambdaHelper {
     }
 
     private class LambdaVisitor : StandardVisitor {
+      // todo (MR) why are we not using this map as a cache for lifted lambdas? why are we recomputing them?
```
RustanLeino: Has this question been resolved?

Source/Core/LambdaHelper.cs
```diff
@@ -77,6 +110,32 @@ string FreshLambdaFunctionName()
...
 
+        return CommandLineOptions.Clo.FreeVarLambdaLifting ? LiftLambdaFreeVars(lambda) : LiftLambdaMaxHoles(lambda);
+      }
+      
+      /// <summary>
+      /// Performs lambda lifting (see <see cref="LambdaHelper.ExpandLambdas"/>) by replacing the lambda's
+      /// free variables with bound ones.
+      /// </summary>
+      /// <param name="lambda">A lambda expression
+      ///   <code>(lambda x1: T1 ... x_n: T_n => t)</code>
```
RustanLeino: The `=>` should be `::`.

Source/Core/LambdaHelper.cs
```diff
@@ -219,6 +278,60 @@ string FreshLambdaFunctionName()
 
         return call;
       }
+      
+      /// <summary>
+      /// Performs lambda lifting (see <see cref="LambdaHelper.ExpandLambdas"/>) by replacing with bound variables
+      /// maximally large subexpressions of a lambda that do not contain any of the lambda's bound variables.
+      /// </summary>
+      /// <param name="lambda">A lambda expression
+      ///   <code>(lambda x1: T1 ... x_n: T_n => t)</code>
```
RustanLeino: `=>` should be `::`

Source/Core/LambdaHelper.cs
```diff
@@ -219,6 +278,60 @@ string FreshLambdaFunctionName()
...
+      ///     replacing the expressions <c>e1</c>, ..., <c>e_m</c> in <c>t</c> with bound variables
+      ///     <c>b1</c>, ..., <c>b_m</c>.
+      ///   </item>
+      ///   <item>
+      ///     Adds a definition and axiom for <c>f</c> to <see cref="lambdaFunctions"/> and <see cref="lambdaAxioms"/>.
+      ///     Memoizes <c>f</c> as the lifted lambda for <para>lambda</para>.
+      ///   </item>
+      /// </list>
+      /// </returns>
+      // todo (MR) move old handling into MaxHolesLambdaLifter? Maybe try merging MaxHolesLambdaLifter with LambdaVisitor
```
RustanLeino: Address or remove these todo's

Source/Core/LambdaHelper.cs
```diff
@@ -219,6 +278,60 @@ string FreshLambdaFunctionName()
...
+      // todo (MR) do I need to handle a more general temporal modifier (e.g. labels)?
+      private Expr LiftLambdaMaxHoles(LambdaExpr lambda) {
+        
+        // We start by getting rid of `old` expressions. Instead, we replace the free variables `x_i` that are
+        // nested inside of `old` expressions with `old(x_i)` expressions.
+        var oldFinder = new OldFinder();
+        oldFinder.Visit(lambda);
```
RustanLeino: It seems to me a more straightforward way of dealing with `old` expressions would be to both "find" and "substitute" them at the same time. Was `OldFinder` already a class before? If not, I suggest writing a new visitor that keeps track of the "old context" and then distributes it inwards.

Test/test2/LambdaExt.bpl
```diff
@@ -67,9 +67,6 @@ procedure Quantifiers() {
 }
 
 procedure FreeVariables() {
```
RustanLeino: Here is another test you could add. It checks that free variables in triggers are substituted appropriately.

```
function g(int,int) : int;

procedure Triggers'(w: int, w': int) {
  var a,b : [int]bool;
  a := (lambda x:int :: (forall u:int :: { g(u,w) } x == g(u,w)));
  b := (lambda y:int :: (forall v:int :: { g(v,w') } y == g(v,w')));
  assert w == w' ==> a == b;
  b := (lambda y:int :: (forall v:int :: y == g(v,w')));
  assert w == w' ==> a == b; // should fail because triggers are different
}
```

Source/Core/MaxHolesLambdaLifter.cs
```diff
@@ -0,0 +1,392 @@
...
+using System.Diagnostics.Contracts;
+using System.IO;
+using System.Linq;
+using Microsoft.Boogie;
+using Type = Microsoft.Boogie.Type;
+
+namespace Core {
+using Set = GSet<object>;
+
+// todo (MR) It would be better if this class were a private class within LambdaHelper because
```
RustanLeino: I think it's okay to keep it where it is. That makes it fit nicely in a file.

Source/Core/MaxHolesLambdaLifter.cs
```diff
@@ -0,0 +1,392 @@
...
+//           this class relies on the lambda not having old expressions any more (only old free variables).
+//           But this makes LambdaHelper very long so I'm leaving this class in a separate file for now.
+
+// todo (MR) test more complicated repl expressions, global variables
```
RustanLeino: Address or remove these todo's.

Source/Core/MaxHolesLambdaLifter.cs
```diff
@@ -0,0 +1,392 @@
...
+        var nodeArgs = node.Args;
+        if (nodeArgs.Any(arg => _templates[arg].ContainsBoundVariables())) {
+            var replacements = new List<Expr>();
+            foreach (Expr arg in nodeArgs) {
+                replacements.AddRange(_templates[arg].GetReplacements());
+            }
+
+            _templates[node] = new TemplateWithBoundVariables(replacements);
+        } else {
+            // todo (MR) Could we avoid the casting? Could we encode in type system that all children of hole nodes are holes?
```
RustanLeino: No, C#'s type system does not allow you to do this, unless you made a complete copy of the AST classes (which I don't recommend). But this is a good invariant to be stated in the AST class for hole nodes.

Source/Core/MaxHolesLambdaLifter.cs
```diff
@@ -0,0 +1,392 @@
...
+        }
+        else {
+            var newRhss = from arg in varBodies select ((TemplateNoBoundVariables) _templates[arg]).GetReplacement();
+            LambdaLiftingTemplate template = new TemplateNoBoundVariables(
+                new LetExpr(node.tok, node.Dummies, newRhss.ToList(),
+                    node.Attributes, // todo (MR) document says attributes should be dropped?
+                    ((TemplateNoBoundVariables) _templates[node.Body])
+                    .GetReplacement()));
+            _templates[node] = template;
+            // todo (MR) explicitly set type of let expr here?
```
RustanLeino: If the resulting expr is supposed to be resolved, then you'd need to set it yourself here. (How was this done in the previous lambda lifter?)

Source/Core/MaxHolesLambdaLifter.cs
```diff
@@ -0,0 +1,392 @@
...
+            }
+
+            replacements.AddRange(bodyTemplate.GetReplacements());
+            _templates[node] = new TemplateWithBoundVariables(replacements);
+        }
+        else {
+            var newRhss = from arg in varBodies select ((TemplateNoBoundVariables) _templates[arg]).GetReplacement();
+            LambdaLiftingTemplate template = new TemplateNoBoundVariables(
+                new LetExpr(node.tok, node.Dummies, newRhss.ToList(),
+                    node.Attributes, // todo (MR) document says attributes should be dropped?
```
RustanLeino: Yes, I think you can drop the attributes (adding a comment that you're doing so). If you do keep them, I suppose you'd need to do substitutions in them as well, just as for triggers.

Source/Core/MaxHolesLambdaLifter.cs
```diff
@@ -0,0 +1,392 @@
...
+
+    public override Expr VisitLambdaExpr(LambdaExpr node) {
+        // Phase I of the lambda lifting
+        base.VisitLambdaExpr(node);
+        if (node.Attributes != null) {
+            VisitQKeyValue(node.Attributes);
+        }
+
+        // Phase II of the lambda lifting. 
+        // todo (MR) this code should not be in a VisitLambdaExpr but in a separate function I think so that we don't confuse Phase I and Phase II.
```
RustanLeino: I think it's okay to have it here

Source/Core/MaxHolesLambdaLifter.cs
```diff
@@ -0,0 +1,392 @@
...
+        
+        // todo (MR) more efficient to do in loop but I find this more clear
```
RustanLeino: fine, keep it

Source/Core/MaxHolesLambdaLifter.cs
```diff
@@ -0,0 +1,392 @@
...
+        }
+
+        NAryExpr call = new NAryExpr(tok, fcall, allReplacementExprs.ToList()) {
+            Type = res.TypedIdent.Type, TypeParameters = SimpleTypeParamInstantiation.From(freeTypeVars, freeVarActuals)
+        };
+
+        return call;
+    }
+
+    // todo (MR) will this take care of all types of variables?
```
RustanLeino: Address or remove todo's

Source/Core/LambdaHelper.cs
amaurremi: > Was `OldFinder` already a class before?

Yes, it was already used by the free-variable-lifting algorithm and I reused it. Should I then keep it as is?

Source/Core/MaxHolesLambdaLifter.cs
amaurremi: The previous lambda lifter just replaced one type of variables (free variables) with another type of variables (bound variables), so it didn't need to do anything special for let expressions. I.e., it could reuse the original let expression and its type. I think it is better to set the type manually so I did that now.

Source/Core/MaxHolesLambdaLifter.cs
amaurremi: Yes, I'm doing substitutions for them. The original tests require the attributes to be present so I let's just keep them.

Test/test2/LambdaExt.bpl
amaurremi: Thanks!

Issue #104, opened 2018-10-11
Printing error model with floating point (real) values


dhruvturakhia:
I tried using the /printModel <n> functionality, but it ended up not being able to print the error model but throws a particular error

Prover error: Model parsing error: Invalid model: invalid element name 719.0, at line 2 (b -> 719.0)
Here is a test code to check:

[test.txt](https://github.com/boogie-org/boogie/files/2470942/test.txt)

zvonimir:
Fixed by pull request #142.

PR #142, opened 2019-08-05
Fixing crash when parsing model with real values

marcoeilers:
Adds support for parsing literals of the form "ddd.ddd" (where d are all digits) to the model parser, thereby fixing https://github.com/boogie-org/boogie/issues/104

### DIFF
--- a/Source/Model/Model.cs
+++ b/Source/Model/Model.cs
@@ -1,4 +1,4 @@
-﻿//-----------------------------------------------------------------------------
+//-----------------------------------------------------------------------------
 //
 // Copyright (C) Microsoft Corporation.  All Rights Reserved.
 //
@@ -45,6 +45,7 @@ namespace Microsoft.Boogie
     public enum ElementKind
     {
       Integer,
+      Real,
       BitVector,
       Boolean,
       Uninterpreted,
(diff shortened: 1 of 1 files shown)


zvonimir:
Thanks for this fix!

Just to clarify this pull request a little bit. So Boogie has real type and floating-point type, and those are two distinct types - the former gets encoded using the theory of reals and the latter using the theory of floating-points, as expected. Now, this pull request only introduces `Real` into `ElementKind` enum (and the respective class).
Does that mean that reals and floating-points are handled uniformly when it comes to models? Could this lead to any problems?

@liammachado It would be great if you could take a look at this pull request too.

marcoeilers:
The example in the referenced bug only used reals, not floating point values, and this PR fixes the problem only for reals. 
I thought that there are no floating point literals in SMTLib but that seems to be incorrect; I just tried an example and the analogous problem occurs with floating point variables:
Example:
```
procedure main (a: real, b: real, y: float24e8, z: float24e8) returns (c: real)
ensures(z == y);
{
}
```
leads to the error `Prover error: Model parsing error: Invalid model: invalid element name -zero, at line 2 (z -> (_ -zero 8 24))`.

marcoeilers:
I checked, and it seems that floating point literals are currently parsed as `ElementKind.DatatypeValue`. As an example, the literal `(fp bv0[1] bv0[8] bv4194304[23])` is parsed as a datatype value with constructor `fp` (I guess that's as expected), and special cases like `(_ +zero 8 24)` are parsed as a datatype value with constructor `_` (I'm not sure whether that's expected, but it does not throw an error). 

My new commit fixes the previous case that failed because of the `-zero`, the problem was that because of the leading minus, Boogie tried to parse this as a numeric value.

That means that if this PR is merged, reals are always parsed as `ElementKind.Real`, and floats are always parsed as `ElementKind.DatatypeValue`.

PR #143, opened 2019-08-06
Adding checksum lambda attribute to lambda lifting to fix Dafny tests

amaurremi:
The new lambda-lifting broke the following tests in the Dafny suite:
```
    Dafny :: dafny0/snapshots/Snapshots8.run.dfy
    Dafny :: server/git-issue223.transcript
    Dafny :: server/simple-session.transcript
```
This PR fixes the tests which occurred because the new lambda lifting did not handle a `checksum` lambda.

### DIFF
--- a/Source/Core/MaxHolesLambdaLifter.cs
+++ b/Source/Core/MaxHolesLambdaLifter.cs
@@ -267,8 +267,20 @@ class MaxHolesLambdaLifter : StandardVisitor {
         var dummies = new List<Variable>(_lambda.Dummies);
         dummies.AddRange(replDummies);
 
+        
+        var lambdaAttrs = _lambda.Attributes;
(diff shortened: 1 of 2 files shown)


Source/Core/MaxHolesLambdaLifter.cs
```diff
@@ -267,8 +267,21 @@ class MaxHolesLambdaLifter : StandardVisitor {
...
 
+        
+        var lambdaAttrs = _lambda.Attributes;
+        if (0 < CommandLineOptions.Clo.VerifySnapshots && QKeyValue.FindStringAttribute(lambdaAttrs, "checksum") == null) {
+            // Attach a dummy checksum to avoid issues in the dependency analysis.
+            var checksumAttr = new QKeyValue(_lambda.tok, "checksum", new List<object> { "lambda expression" }, null);
+            if (lambdaAttrs == null) {
+                lambdaAttrs = checksumAttr;
+            }
+            else {
```
RustanLeino: Move this `else` up to the previous line.

PR #145, opened 2019-08-17
Updates for CVC4 compatibility

barrettcw:
The parsing of CVC4 models was incorrect or out of date in several places.  This fixes those problems.  At least one testcase goes through using CVC4 now.


shazqadeer:
Which version of CVC4 are these changes consistent with?

barrettcw:
These changes are consistent with the latest CVC4 master.

PR #146, opened 2019-08-19
 Fix bug in NativeIntervallDomain.Node.Create

xurongchen:
The intersection of two constraints should have a upperbound equaling to the larger origin upperbound.


shazqadeer:
@RustanLeino : Do you know who wrote this code originally?  It would be great if the original author could take a quick look.  It is a one line change.

RustanLeino:
I'm the original author and that does appear to be a bug (and a good fix). @xurongchen , please include a test case with your pull request (something that used to be done incorrectly and that your fix corrects). Thanks!

shazqadeer:
@xurongchen : Could you provide a simple test so we can close this PR?

shazqadeer:
The modification looks simple and has been vetted by @RustanLeino .  I am going ahead and merging this PR.

PR #152, opened 2019-10-15
refactoring of CIVL refinement code

shazqadeer:
This PR refactors the file CivlRefinement.cs into a collection of smaller files.  
- split out the instrumentation for refinement checking and yield checking into separate classes.
- split out the transformation of an implementation into a separate file.
Extraneous "using" clauses are deleted in a few other files.


shazqadeer:
I have done the renaming.

PR #156, opened 2019-10-24
CIVL: new transition relation computation

bkragl:
This PR replaces the current transition relation computation (which relies on several heuristics) with a new one that is more principled in addressing the problem of avoiding internally introduced existential quantifiers in transition relations.

The following new language features are introduced:
* Backward assignments in atomic actions
* Witness functions for commutativity checking

Thanks to @SoroushE for this work!


shazqadeer:
@SoroushE , @bkragl :  Thanks a lot for the great work on this PR.  This work provides a good foundation for the treatment of transition relations in CIVL.

shazqadeer:
I noticed that the merge history is somewhat messed up.  Perhaps "git merge" rather than "git rebase" was used.  In future, let us use "git rebase" and also squash commits into a single commit.

bkragl:
> I noticed that the merge history is somewhat messed up. Perhaps "git merge" rather than "git rebase" was used. In future, let us use "git rebase" and also squash commits into a single commit.

I fixed this issue. I had to resolve some conflicts locally and accidentally created two merge commits. One proper merge commit that actually merged two streams, but then one other nonsense merge commit when moving to master.
I removed the second one and now you should see a clean history.

I'm somewhat against rebasing and squashing, because I think there is some value in the development history.

shazqadeer:
It sounds your objection is to squashing but not to rebasing.  Is my understanding correct?

bkragl:
No, I'm against rebasing because it rewrites the history. I don't see a reason why it should artificially be made linear, when it really was parallel.

I guess squashing makes sense to eliminate "exploratory" commits.

PR #161, opened 2019-11-29
Add Coco/R and modified frame files under new ThirdParty directory

### LINKED ISSUE #132: Errors in generated parser.
The parser generated using Boogie’s (Modified) Coco generates thousands of compile errors — due to inconsistencies with the names `_x` and `x` or `_T` and `t`. Here are the steps to reproduce the error, with the current head of boogie-org/boogie-partners:

```
(cd Source/Core && mono ../../../boogie-partners/CocoR/Modified/Coco.exe BoogiePL.atg -namespace Microsoft.Boogie -frames ../../../boogie-partners/CocoR/Modified)
msbuild Source/Boogie.sln
```

Here’s a sample of the errors:
```
Parser.cs(2214,217): error CS0103: The name 'T' does not exist in the current context [/Users/mje/Code/boogie-mje/Source/Core/Core-NetCore.csproj]
Parser.cs(2214,220): error CS0103: The name 'T' does not exist in the current context [/Users/mje/Code/boogie-mje/Source/Core/Core-NetCore.csproj]
Parser.cs(2214,222): error CS0103: The name 'x' does not exist in the current context [/Users/mje/Code/boogie-mje/Source/Core/Core-NetCore.csproj]
Parser.cs(2214,224): error CS0103: The name 'x' does not exist in the current context [/Users/mje/Code/boogie-mje/Source/Core/Core-NetCore.csproj]
```

bkragl:
This PR adds Coco/R (binary and sources) and the modified frame files necessary to (re)generate Boogie's scanner and parser directly to the repository. This should make it easier to get somebody going that needs to modify this part of Boogie (e.g., this resolves #132).

Since Coco/R has a different license than Boogie, it is located in a new ThirdParty subdirectory and a NOTICES.md file explains the license(s) that apply to the content in this subdirectory. This is similar to the Dafny repository, so I hope this is properly done. Please somebody double-check: @RustanLeino, @zvonimir, @shazqadeer, @michael-emmi.


bkragl:
I get the concern regarding licensing. That's why I noted that Dafny already does exactly the same thing and asked if somebody can confirm for sure that it is fine.

For me it would be also fine to update and link the boogie-partners repo (I'm not a big fan of Git submodules though). However, I don't understand the concern about confusion and complicating maintainability. The expectation is that the Coro/R version we use will be updated very rarely (overall I saw two updates in the log). Updating the grammar to support new features (floating-point arithmetic, strings, concurrency, ...) is much more frequent. And for someone who wants to update the grammar nothing seems to be more straightforward than having everything directly in the Boogie repo.

What are the complications you have in mind?

michael-emmi:
I worry that the licensing issue transcends the current users of Boogie and Dafny, and could prevent future uses. We certainly don’t want to *add* constraints, which including another license almost certainly does.

Fandom of submodules aside — I only suggested it as an alternative to including Coco/R directly — I’m no expert of software engineering, but I believe the general issues are code duplication, and the tying together of version histories which should be independent. Imagine somebody commits updates to the Coco/R grammar while somebody else commits updates to proper Boogie code, and the latter needs to be reverted. Now simply rolling back will undo both. Will somebody be in a position to cherry pick commits at that point? It‘s not clear that there is or will be a qualified owner. As I said, I'm no expert, but I suspect that seasoned software engineers would advise against these two pieces of software. Perhaps a second opinion is in order.

Also, note that Coco/R is neither a runtime nor build-time dependency, and is seldom used here. Simply including a pointer to the currently-compatible sources/binaries is adequate.

zvonimir:
Stupid question - but why would Coco be a submodule instead of a completely independent repo under boogie-org?
I think we should move it from boogie-partners repo into its own coco repo under boogie-org.
For a developer who would like to use Coco to generate the Boogie parser, to me it doesn't seem like a large overhead to checkout coco repo manually. But I guess I am missing something obvious here.

zvonimir:
Ah, I see, you are invoking Coco directly from the make file. Makes sense.
Ok, I am not a software engineering expert, but I view Coco to be an external dependency and I would not include it directly into the repo. I would prefer to have it be included as a submodule. It should be in its own repo in the boogie-org I think (and it should be removed from boogie-partners).

bkragl:
I revised my PR.

I created a repository named coco under boogie-org which contains the code of Coco/R, and a project file to generate a NuGet package I'm now hosting on [nuget.org](https://www.nuget.org/packages/CocoR/).

Now we can use Coco/R in Boogie as a .NET Core local tool. There is a configuration file called `dotnet-tools.json`. Running `dotnet tool restore` installs Coco/R and then it can be invoked from anywhere within the Boogie code base as `dotnet tool run coco` (this is what the Makefile now does).

The only question left is regarding the "frame files" `Parser.frame` and `Scanner.frame`. These are modifications of the original Coco/R ones, which are under GPL (since Coco/R is self-histing). However, the Coco/R license explicitly says the following.

> If not otherwise stated, any source code generated by Coco/R (other than Coco/R itself) does not fall under the GNU General Public License.

So I think it is perfectly fine to keep this files in the Boogie repo (in this PR I put them under `Source/Core`, next to the grammar). They can stay under GPL without "infecting" Boogie, since the created `Parser.cs` and `Scanner.cs` which are compiled in Boogie don't fall under GPL.

Does this sound good?

.gitignore
```diff
@@ -21,6 +21,10 @@ Binaries/*.pdb
 Binaries/*.config
 Binaries/*.vshost.exe.manifest
 
+# Coco/R
+Scanner.cs.old
```
zvonimir: What are these `.old` files?

bkragl: Backup copies that Coco creates when the output files (`Parser.cs` and `Scanner.cs`) already exist.

zvonimir:
Btw, one final thing to do is to remove Coco from here: https://github.com/boogie-org/boogie-partners

shazqadeer:
It all looks fine to me except I don't understand the point of creating a copy of the Coco source code under boogie-org.  Please explain.

bkragl:
> It all looks fine to me except I don't understand the point of creating a copy of the Coco source code under boogie-org. Please explain.

It contains the configuration files to compile under .NET Core and create a NuGet package. In case this should be ever updated, I thought it is better to put it here where other people have also write access, as opposed to for example under my personal GitHub account.

shazqadeer:
So, the coco repository does not contain a copy of the source code of Coco?  I am asking because the source code might get out of sync with the source code maintained by the developers of Coco.

zvonimir:
@shazqadeer As far as I know, Coco does not have a source code repo - it is distributed as a zip file. It is also mostly a dormant project with very few code changes being issued occasionally. So I think it would be nice to have a "mirror" on our end.

bkragl:
> @shazqadeer As far as I know, Coco does not have a source code repo - it is distributed as a zip file. It is also mostly a dormant project with very few code changes being issued occasionally. So I think it would be nice to have a "mirror" on our end.

I agree.  Actually I was almost done writing the same :)

shazqadeer:
Having a mirror is fine.  But we should indicate to the Coco team that we are interested in maintaining a mirror so they at least notify us when changes are made at their end.  Perhaps they would even consider making our Coco repository the official one.  We could give the Coco team admin/write privileges on our repo if they want it.

I suggest we go ahead with this PR and start communicating with the Coco developers in parallel.

zvonimir:
@shazqadeer Agreed, good idea!

bkragl:
I already had an email exchange with the Coco/R developer (Hanspeter Mössenböck) to check if they are OK with me creating this NuGet package and what the version numbering scheme should be. Boogie and the NuGet package are already mentioned on their webiste: http://www.ssw.uni-linz.ac.at/coco/

michael-emmi:
@bkragl excellent!

PR #166, opened 2019-12-07
Improvements of the .NET Core build

bkragl:
**Edit:** This PR evolved quite a bit from its original version. Points 1 and 3 below do not apply anymore, and point 5 was added.

This PR makes the following (in my opinion) improvements:

~1. The build output of the BoogieDriver project is set to the `Binaries` directory, as for the old build. Thus the compiled executable is written to `Binaries/Boogie.dll`. The option `AppendTargetFrameworkToOutputPath` is set to `false` because otherwise the location would depend on the target framework (e.g., `Binaries/netcoreapp2.2/Boogie.dll`). Also, the created NuGet package is placed into `Binaries/nupkg/`.~
2. The lit test tool can now be called with `-D dotnet` to run the regression tests with the .NET Core build.
~3. A wrapper script `Binaries/boogie-dotnet` is added. This makes it more convenient to call on the command line or to configure the Emacs Flycheck mode for Boogie.~
4. The build instructions in `README.md` are cleaned up and emphasize that .NET Core is from now on the preferred way to build/run Boogie. I think this was the consensus from previous discussions. Also, more details for the three supported SMT solvers are given.
5. **Added**: The Travis CI build is switched from Mono to .NET Core and configured to automatically deploy tagged releases.

@michael-emmi, please let me know if you agree with the above changes, or if anything should be updated.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,27 +1,37 @@
 # vim: set sw=2 ts=2 softtabstop=2 expandtab:
 language: csharp
-dist: trusty
+dist: bionic
 sudo: true
-solution: "Source/Boogie.sln"
+dotnet: 3.1
+mono: none
 env:
   global:
+    - SOLUTION=Source/Boogie-NetCore.sln
     - Z3URL=https://github.com/Z3Prover/z3/releases/download/z3-4.8.4/z3-4.8.4.d6df51951f4c-x64-ubuntu-14.04.zip
   matrix:
-    - BOOGIE_CONFIG=Debug
-    - BOOGIE_CONFIG=Release
+    - CONFIGURATION=Debug
+    - CONFIGURATION=Release
 install:
   # Download a Z3 release
   - wget ${Z3URL}
   - unzip z3*.zip
-  # NuGet is a little flakey in legacy TravisCI, use travis_retry command to retry the command if it fails
-  - travis_retry nuget restore ${TRAVIS_SOLUTION}
-  - travis_retry nuget install -Version 3.10.0 NUnit.Console -OutputDirectory Source/packages/
+  - export PATH="$(find $PWD/z3* -name bin -type d):$PATH"
   # Install needed python tools
   - sudo pip install lit OutputCheck pyyaml
 script:
-  - msbuild /p:Configuration=${BOOGIE_CONFIG} ${TRAVIS_SOLUTION}
-  # Run unit tests
-  - python Source/UnitTests/run-unittests.py ${BOOGIE_CONFIG}
-  # Run driver tests
-  - ln -s $(find $PWD/z3* -name z3 -type f) Binaries/z3.exe
-  - lit -v Test/
+  - dotnet build -c ${CONFIGURATION} ${SOLUTION}
+  - dotnet test --no-build -c ${CONFIGURATION} ${SOLUTION}
+  - lit -v -D dotnet -D configuration=${CONFIGURATION} Test
+deploy:
+  skip_cleanup: true
+  on:
+    branch: master
+    tags: true
+    condition: ${CONFIGURATION}=Release
+  - provider: script
+    script :
+      - dotnet nuget push Source/BoogieDriver/bin/${CONFIGURATION}/Boogie*.nupkg -k ${NUGET_API_KEY} -s https://api.nuget.org/v3/index.json
+  - provider: releases
+    api_key: TODO
+    file_glob: true
+    file: Source/BoogieDriver/bin/${CONFIGURATION}/Boogie*.nupkg
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,8 @@ dist: bionic
 sudo: true
 dotnet: 3.1
 mono: none
+git:
+  depth: false
 env:
   global:
     - SOLUTION=Source/Boogie-NetCore.sln
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,8 @@ deploy:
       tags: true
       condition: ${CONFIGURATION}=Release
   - provider: releases
-    api_key: TODO
+    api_key:
+      secure: ZjKhOiIpC6R+Xfp1iJX/1a2DD1o+tYhUefZDqRjUfM4rDZqzvOBvY7mA/1BcqNs4gXJIk3p11Kud72cPSSS8iW2EVlRm2UlfdVOf2wmGys/TILvHNDWUoVFSxhVgxbzMVULp6fIrqDypaZ0PAYZVg2loLkVI5AZ/P35ZRVaa9oE=
     file_glob: true
     file: Source/BoogieDriver/bin/${CONFIGURATION}/Boogie*.nupkg
     skip_cleanup: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,15 +23,20 @@ script:
   - dotnet test --no-build -c ${CONFIGURATION} ${SOLUTION}
   - lit -v -D dotnet -D configuration=${CONFIGURATION} Test
 deploy:
-  skip_cleanup: true
-  on:
-    branch: master
-    tags: true
-    condition: ${CONFIGURATION}=Release
   - provider: script
-    script :
+    script:
       - dotnet nuget push Source/BoogieDriver/bin/${CONFIGURATION}/Boogie*.nupkg -k ${NUGET_API_KEY} -s https://api.nuget.org/v3/index.json
+    skip_cleanup: true
+    on:
+      branch: master
+      tags: true
+      condition: ${CONFIGURATION}=Release
   - provider: releases
     api_key: TODO
     file_glob: true
     file: Source/BoogieDriver/bin/${CONFIGURATION}/Boogie*.nupkg
+    skip_cleanup: true
+    on:
+      branch: master
+      tags: true
+      condition: ${CONFIGURATION}=Release
--- a/Source/BoogieDriver/BoogieDriver-NetCore.csproj
+++ b/Source/BoogieDriver/BoogieDriver-NetCore.csproj
@@ -13,7 +13,6 @@
     <AssemblyName>BoogieDriver</AssemblyName>
     <TargetFramework>netcoreapp3.1</TargetFramework>
     <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
-    <PackageOutputPath>nupkg</PackageOutputPath>
     <PackAsTool>true</PackAsTool>
     <ToolCommandName>boogie</ToolCommandName>
   </PropertyGroup>
--- a/.gitignore
+++ b/.gitignore
@@ -1,25 +1,20 @@
 # Nuget
-Source/packages/*
-Source/**/nupkg
+Source/packages/
+
+# Binaries
+bin/
+obj/
+Binaries/**/*.dll
+Binaries/**/*.mdb
+Binaries/**/*.exe
+Binaries/**/*.pdb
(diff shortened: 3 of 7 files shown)


michael-emmi:
Just curious, why the wrapper script and all that?

According to [the docs](https://docs.microsoft.com/en-us/dotnet/core/tools/global-tools-how-to-create#setup-the-global-tool) you can just install the tool from a package:
```console
dotnet tool install --global --add-source ./nupkg botsay
```
i.e., the package does not need to exist (yet) in the online NuGet repository.

Also, it seems that local tools are now possible too, avoiding a global install: https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-tool-install

But maybe creating a script is preferable for some reason?

Source/BoogieDriver/BoogieDriver-NetCore.csproj
```diff
@@ -9,12 +9,14 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <AssemblyName>BoogieDriver</AssemblyName>
+    <AssemblyName>Boogie</AssemblyName>
```
michael-emmi: Why this change? Maybe better to maintain the legacy structure as much as possible?

bkragl: I thought it's nice to call the "main" executable DLL just `Boogie.dll`, but I understand the desire to name it the same as the project. I'll rename it back to `BoogieDriver`.
But you agree with `Binaries/` in the repository root as the output path?

bkragl:
The wrapper script is intended for developers. During (iterative) development and testing it does not seem right to always do an installation step.

But we should add instructions on how to install the global tool from nuget.org for people who are just interested in using Boogie.

Source/BoogieDriver/BoogieDriver-NetCore.csproj
michael-emmi: Well, since that is how Boogie was built before the `-NetCore.csproj` variations, yes.

michael-emmi:
> During (iterative) development and testing it does not seem right to always do an installation step.

Agreed, though the step I suggest would be an “install” that is *local* to the workspace, i.e., after generating `BoogieDriver.dll` (which is anyhow done in iterative development) generate the package then have dotnet “install” the tool in the workspace directory, which the tests can use — I’m supposing those steps don’t add significant overhead to iterative development.

I think either approach could work, and I suggested the above to avoid having a script that could be generated automatically, and avoid divergence with what’s actually on regular (global) installs.

Anyhow I won’t insist on this point, since I’m not so frequently involved in the Boogie development cycle :-)

michael-emmi:
@bkragl how about updating `.travis.yml` to build for dotnet core (also)?

If you do that, then we can almost immediately automate publishing versioned dotnet global tools to NuGet.org with #174.

bkragl:
Sure, we can add a .NET Core build to Travis.

However, I was rethinking my proposed updates in this PR. Specifically, how to run Boogie as a developer and from the Lit tests. I don't particularly like the output of the BoogieDriver project to be written to `./Binaries/` and having that script (actually I would hope that `Binaries` can be deleted at some point), but the installation you mentioned above is also not completely satisfactory.

I tried `dotnet tool install` and found that there are three options where exactly one can be given:

* `--local`: This is meant for tools that should be installed and available "locally". It relies on a tool manifest file that specifies the desired tools and versions. However, the actual installation is not done locally, but somewhere in the HOME directory (such that it can be shared in case different projects want to have the same local tool). According to my understanding this would be a perfect use case for Coco/R for us. My plan is to create a Coco/R package on nuget.org that we can include as a local tool into Boogie.

* `--global`: This makes a global installation for the current user. This would be the right thing for a user of Boogie, but not for development/testing.

* `--tool-path`: This takes a directory as target for the installation (which would be truly local). Here we could say `--tool-path test-installation`. However, it seems that there is no way to explicitly tell dotnet that the installation should be performed from the local package. You have to specify `--add-source Source/BoogieDriver/bin/$CONFIGURATION`, but I'm not sure if there could be ambiguity if there is already a published version on nuget.org with the same version number (because during development the version number is not yet bumped).

Do you think we should try the last option, or for now call `dotnet Source/BoogieDriver/bin/Debug/netcoreapp2.2/BoogieDriver.dll` for testing?

bkragl:
BTW, we should perhaps upgrade to .NET Core 3.1 which is a long term support (LTS) release.

michael-emmi:
> Do you think we should try the last option, or for now call dotnet Source/BoogieDriver/bin/Debug/netcoreapp2.2/BoogieDriver.dll for testing?

The would avoid ad-host invocation scripts, as long as the potential caveat you mention can be resolved:

> but I'm not sure if there could be ambiguity if there is already a published version on nuget.org with the same version number

I’m supposing that the local installation would take precedence? Is there an easy way to verify this?

michael-emmi:
> BTW, we should perhaps upgrade to .NET Core 3.1 which is a long term support (LTS) release.

Absolutely.

michael-emmi:
Hey @bkragl just noticed something

> A wrapper script Binaries/boogie-dotnet is added. This makes it more convenient to call on the command line or to configure the Emacs Flycheck mode for Boogie.

The current build seems to generate `./Source/BoogieDriver/bin/Debug/netcoreapp3.1/BoogieDriver`, which can be invoked directly.

michael-emmi:
Also see #180 for improvements to the -NetCore.csproj files themselves.

michael-emmi:
@bkragl how’s this one coming? need any assist?

bkragl:
> @bkragl how’s this one coming? need any assist?

I'll post a proposal soon.

bkragl:
@michael-emmi: I basically restarted this PR, pulling in the recent changes we did. I switched the Travis build to .NET Core and added configurations to deploy to nuget.org and GitHub releases (we'll have to see if this works as intended). Unfortunately there seems to be a problem during compilation related to GitVersion.

This is the error message I saw in the build log:
```
/home/travis/.nuget/packages/gitversiontask/5.1.2/build/GitVersionTask.targets(10,9): error : InvalidOperationException: Could not find a 'develop' or 'master' branch, neither locally nor remotely. [/home/travis/build/boogie-org/boogie/Source/CodeContractsExtender/CodeContractsExtender-NetCore.csproj]
```

Perhaps the git metadata is not available in the Travis environment? Could you look into that?

If this is fixed, we also have to add a GitHub oauth token, if we want to have files attached to the releases on GitHub.

michael-emmi:
> Perhaps the git metadata is not available in the Travis environment? Could you look into that?

Looks like you have to update increase the clone depth:
https://docs.travis-ci.com/user/customizing-the-build#git-clone-depth

Source/BoogieDriver/BoogieDriver-NetCore.csproj
```diff
@@ -13,7 +13,6 @@
     <AssemblyName>BoogieDriver</AssemblyName>
     <TargetFramework>netcoreapp3.1</TargetFramework>
     <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
-    <PackageOutputPath>nupkg</PackageOutputPath>
```
michael-emmi: Why are we deleting this?

.travis.yml
```diff
@@ -1,27 +1,44 @@
...
-  # Run driver tests
-  - ln -s $(find $PWD/z3* -name z3 -type f) Binaries/z3.exe
-  - lit -v Test/
+  - dotnet build -c ${CONFIGURATION} ${SOLUTION}
+  - dotnet test --no-build -c ${CONFIGURATION} ${SOLUTION}
+  - lit -v -D dotnet -D configuration=${CONFIGURATION} Test
+deploy:
+  - provider: script
+    script:
+      - dotnet nuget push Source/BoogieDriver/bin/${CONFIGURATION}/Boogie*.nupkg -k ${NUGET_API_KEY} -s https://api.nuget.org/v3/index.json
```
michael-emmi: Wouldn't the path `Source/BoogieDriver/nupkg/Boogie.2.4.2.nupkg` be better here?

Source/BoogieDriver/BoogieDriver-NetCore.csproj
bkragl: I wanted to see where the package is placed by default and found it cleaner to have no overlap between the build output of different configurations, i.e., everything concerning a particular configuration is placed in `bin/${CONFIGURATION}/`. I'll add it back if you prefer.

michael-emmi: We’re considering builds with multiple configurations?

bkragl: Yes. Currently Travis is configured to build and test the Debug and Release configurations. The deployment should only happen for Release.

michael-emmi: But those are two separate builds, one configuration per build.

Wouldn’t it be nice to find the package in the same place independently of configuration?

bkragl: I was thinking of a local clone of the repository. The same argument could be made for the other build output. Putting everything into `bin` instead of `bin/Debug/netcore3.1` would make it easier to find the binaries (e.g., in test scripts). Still, I'm against that because it might be useful to have multiple configurations compiled simultaneously.

That being said, for the package I don't care.

michael-emmi: Right. For `bin` I agree, because, like you say, you might want to build both locally.

But these two lines are just about the package and related configuration for Travis, which only builds one configuration, and can be simplified accordingly.

michael-emmi: Then again, maybe deleting the line above is better than simplifying the line below.

Let’s keep your version :-).

.travis.yml
bkragl: Any idea for a clean way to get rid of the wildcard?

michael-emmi: No, but this seems clean already. Are you worried that this could match multiple files in a single travis build?

bkragl: It's more about knowing that there is a single file and still using a wildcard. But you are right, this is probably more stable than some way of trying to compute the exact file name. It's similar to the code that puts Z3 into PATH. We won't have to change it when we switch to a newer version.

.travis.yml
```diff
@@ -1,27 +1,44 @@
...
+    skip_cleanup: true
+    on:
+      branch: master
+      tags: true
+      condition: ${CONFIGURATION}=Release
+  - provider: releases
+    api_key: TODO
```
bkragl: @michael-emmi, do you know how to fill this? I can look into it, but I'm not sure if I have the necessary privileges since I'm not an owner of the Boogie repo.

Or do we want to eliminate the deployment to GitHub?

michael-emmi: I’m not clear on the utility of GitHub Releases given NuGet releases, but I’ve added an API key anyhow :-)

michael-emmi:
@bkragl is this ready to merge?

bkragl:
The readme file should be updated with a pointer to the NuGet package for people that are only interested in using Boogie, and some more compilation instructions for .NET Core. I can do that tomorrow.

Otherwise I think we can merge.

michael-emmi:
@bkragl cool let me know when you add those and I can merge when ready.

bkragl:
> @bkragl cool let me know when you add those and I can merge when ready.

Done. I also updated the original description of this PR.

PR #174, opened 2019-12-14
Using GitVersion for NetCore CLI and package.

michael-emmi:
**NOTE**: please do not merge this yet, this is meant to be discussed with @bkragl.

This PR leverages [GitVersion](https://gitversion.readthedocs.io) to generate version numbers for the CLI and NuGet package automatically; version numbers are generated from git history, e.g., by looking at branch names and tags.

### DIFF
--- a/Source/AIFramework/AIFramework-NetCore.csproj
+++ b/Source/AIFramework/AIFramework-NetCore.csproj
@@ -5,6 +5,7 @@
     <AssemblyName>AIFramework</AssemblyName>
     <TargetFramework>netcoreapp3.1</TargetFramework>
     <DefineConstants>COREFX_SUBSET</DefineConstants>
+    <GenerateGitVersionInformation>false</GenerateGitVersionInformation>
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
   </PropertyGroup>
 
@@ -15,6 +16,13 @@
 
   <ItemGroup>
     <Compile Remove="**\cce.cs" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="GitVersionTask" Version="5.1.2">
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
+      <PrivateAssets>all</PrivateAssets>
+    </PackageReference>
   </ItemGroup>
 
 </Project>
--- a/Source/AIFramework/AIFramework-NetCore.csproj
+++ b/Source/AIFramework/AIFramework-NetCore.csproj
@@ -5,7 +5,6 @@
     <AssemblyName>AIFramework</AssemblyName>
     <TargetFramework>netcoreapp3.1</TargetFramework>
     <DefineConstants>COREFX_SUBSET</DefineConstants>
-    <GenerateGitVersionInformation>false</GenerateGitVersionInformation>
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
   </PropertyGroup>
 
--- a/Source/AIFramework/AIFramework-NetCore.csproj
+++ b/Source/AIFramework/AIFramework-NetCore.csproj
@@ -14,7 +14,6 @@
   </ItemGroup>
 
   <ItemGroup>
-    <Compile Include="..\version.cs" />
     <Compile Remove="**\cce.cs" />
   </ItemGroup>
 
--- a/Source/AbsInt/AbsInt-NetCore.csproj
+++ b/Source/AbsInt/AbsInt-NetCore.csproj
@@ -5,7 +5,6 @@
     <AssemblyName>BoogieAbsInt</AssemblyName>
     <TargetFramework>netcoreapp3.1</TargetFramework>
     <DefineConstants>COREFX_SUBSET</DefineConstants>
-    <GenerateGitVersionInformation>false</GenerateGitVersionInformation>
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
   </PropertyGroup>
 
--- a/Source/AbsInt/AbsInt-NetCore.csproj
+++ b/Source/AbsInt/AbsInt-NetCore.csproj
@@ -5,6 +5,7 @@
     <AssemblyName>BoogieAbsInt</AssemblyName>
     <TargetFramework>netcoreapp3.1</TargetFramework>
     <DefineConstants>COREFX_SUBSET</DefineConstants>
+    <GenerateGitVersionInformation>false</GenerateGitVersionInformation>
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
   </PropertyGroup>
 
@@ -17,6 +18,13 @@
 
   <ItemGroup>
     <Compile Remove="**\cce.cs" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="GitVersionTask" Version="5.1.2">
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
+      <PrivateAssets>all</PrivateAssets>
+    </PackageReference>
   </ItemGroup>
 
 </Project>
--- a/Source/AbsInt/AbsInt-NetCore.csproj
+++ b/Source/AbsInt/AbsInt-NetCore.csproj
@@ -16,9 +16,6 @@
   </ItemGroup>
 
   <ItemGroup>
-    <Compile Include="..\version.cs">
-      <Link>version.cs</Link>
-    </Compile>
     <Compile Remove="**\cce.cs" />
   </ItemGroup>
 
--- a/Source/Basetypes/Basetypes-NetCore.csproj
+++ b/Source/Basetypes/Basetypes-NetCore.csproj
@@ -5,6 +5,7 @@
     <AssemblyName>BoogieBasetypes</AssemblyName>
     <TargetFramework>netcoreapp3.1</TargetFramework>
     <DefineConstants>COREFX_SUBSET</DefineConstants>
+    <GenerateGitVersionInformation>false</GenerateGitVersionInformation>
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
   </PropertyGroup>
 
@@ -14,6 +15,13 @@
 
   <ItemGroup>
     <Compile Remove="**\cce.cs" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="GitVersionTask" Version="5.1.2">
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
+      <PrivateAssets>all</PrivateAssets>
+    </PackageReference>
   </ItemGroup>
 
 </Project>
--- a/Source/Basetypes/Basetypes-NetCore.csproj
+++ b/Source/Basetypes/Basetypes-NetCore.csproj
@@ -5,7 +5,6 @@
     <AssemblyName>BoogieBasetypes</AssemblyName>
     <TargetFramework>netcoreapp3.1</TargetFramework>
     <DefineConstants>COREFX_SUBSET</DefineConstants>
-    <GenerateGitVersionInformation>false</GenerateGitVersionInformation>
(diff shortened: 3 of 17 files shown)


michael-emmi:
Hey @bkragl @zvonimir what are your thoughts on this one?

bkragl:
In general I like the concept, but I don't quite understand how this thing works and if it is stable. Also I'm not sure if I understand what these fancy versioning concepts like "Semantic Versioning" are.

Anyway, I tried to build and found the following: I get a package that has "2.4.2" in the name. Probably because "2.4.1" was the last tag and then it just incremented by 1. Running Boogie on the command line I get the logo message:
```
Boogie program verifier version 2.4.2.0, Copyright (c) 2003-2014, Microsoft.
```
Will this trailing ".0" will always be 0?

Now the problematic part. If I do some git operation, like adding a commit, I get the following error if I want to rebuild.
```
/home/bkragl/.nuget/packages/gitversiontask/5.1.3-beta1.61/build/GitVersionTask.targets(19,9): warning : File contains assembly version attributes which conflict with the attributes generated by GitVersion obj/Debug/netcoreapp2.2/AssemblyInfo_Core-NetCore_uhevcwtw.e1o.g.cs [/home/bkragl/Code/boogie/Source/Core/Core-NetCore.csproj]
obj/Debug/netcoreapp2.2/GitVersionInformation_Core-NetCore_c2es2cnt.vwb.g.cs(12,14): error CS0101: The namespace '<global namespace>' already contains a definition for 'GitVersionInformation' [/home/bkragl/Code/boogie/Source/Core/Core-NetCore.csproj]
obj/Debug/netcoreapp2.2/GitVersionInformation_Core-NetCore_c2es2cnt.vwb.g.cs(11,2): error CS0579: Duplicate 'global::System.Runtime.CompilerServices.CompilerGenerated' attribute [/home/bkragl/Code/boogie/Source/Core/Core-NetCore.csproj]
```

zvonimir:
@michael-emmi : could you explain the steps for pushing out a release using this?

In SMACK, I manually go to Releases on GitHub, I click on "Create a new release", and I type in my desired version number manually as well as other stuff, such as release notes. Then the master branch gets tagged with the entered release number, and the release gets created (which currently only amounts to source code being zipped, meaning no packages get created).

It is not completely clear to me how this pull request influences/changes the above steps. It would be good to document the Boogie release steps somewhere anyways.

michael-emmi:
for @bkragl’s concerns:
>  I don't quite understand how this thing works…

I won’t repeat the [GitVersion docs](https://gitversion.readthedocs.io/), but the basic idea is to
1. determine a version number for a given build by looking at Git history, e.g., tags, branches, etc; and
2. insert that version number into build artifacts, e.g., assemblies and packages. 

In our case, the version number `2.4.2.0` is derived from the last tag having been `v2.4.1`, and there having been commits since then. If you were to tag a commit with `v2.4.3`, e.g., then the version number derived by that commit would be `2.4.3.0`, and the version number of subsequent commits would be `2.4.4.0`. Looking at the generated file `Source/BoogieDriver/obj/netcoreapp2.2/GitVersionInformation_BoogieDrvier-NetCore_xxxx.bxe.g.cs` is somewhat informative:
```csharp
static class GitVersionInformation
{
   public static string Major = "2";
   public static string Minor = "4";
   public static string Patch = "2";
   public static string PreReleaseTag = "gitversion.1";
   public static string PreReleaseTagWithDash = "-gitversion.1";
   public static string PreReleaseLabel = "gitversion";
   public static string PreReleaseNumber = "1";
   public static string WeightedPreReleaseNumber = "1";
   public static string BuildMetaData = "51";
   public static string BuildMetaDataPadded = "0051";
   public static string FullBuildMetaData = "51.Branch.gitversion.Sha.9d70fb097f00a9e6244440b8ac3210f080e49c69";
   public static string MajorMinorPatch = "2.4.2";
   public static string SemVer = "2.4.2-gitversion.1";
   public static string LegacySemVer = "2.4.2-gitversion1";
   …
```

> …and if it is stable

There are millions of downloads of its CLI, and nearly a million of this task:

Screen Shot 2019-12-17 at 10 35 07 AM

> what these fancy versioning concepts like "Semantic Versioning" are

see [Semantic Versioning](https://semver.org)

> Will this trailing ".0" will always be 0?

Seems like this is some legacy Visual Studio thing, which GitVersion will always fill with zero.

> Now the problematic part. If I do some git operation, like adding a commit, I get the following error if I want to rebuild.

Yes, there is a slight kink I still have to work out, due to an extra `AssemblyInfo_xxx.cs` file being generated on each build. I will fix this if we agree to use this GitVersion approach, but for now, you can always do a fresh build by running `git clean -fxd` first.

michael-emmi:
For @zvonimir:
> could you explain the steps for pushing out a release using this?

The nice thing about this is that it is completely separate from pushing out releases.

Essentially, when you want to increment the version number, you just tag a commit with, e.g., `v2.4.2` or `v.2.5.0`.

The artifacts (assembles, packages) that are built from that commit with carry the corresponding version numbers. You can push those artifacts as GitHub releases, NuGet packages, etc.

For instance, a travis build could publish the generated NuGet package to nuget.org when the build is a merge to master. Likewise, it could generate a GitHub release. You could also use Github Actions do do this.

In summary, the steps would be (1) tag, (2) build, (3) push. I assume that steps 2 and 3 would be automated. I also think it would be cool to automate step 1, but that’s another conversation altogether which I'd like to have in the future.

> It would be good to document the Boogie release steps somewhere anyways.

I will absolutely document this if we agree on the GitVersion approach.

zvonimir:
Ok, this makes sense to me now and seems like something we should move forward with.

michael-emmi:
> Now the problematic part. If I do some git operation, like adding a commit, I get the following error if I want to rebuild.

@bkragl solved with the latest commit. We won’t see the GitVersionInformation*.cs files anymore, but we aren’t using them anyways.

bkragl:
Hi @michael-emmi. I found the actual culprit of this issue. It is the line
```
<Compile Include="**\*.cs" />
```
which includes (multiple) stale versions of the generated GitVersionInformation*.cs files. My suggestion would be to remove the line and use the default compilation items, i.e., also remove the following lines from all project files:
```
<PropertyGroup>
  <EnableDefaultCompileItems>false</EnableDefaultCompileItems>
</PropertyGroup>
```

Then we can enable `GenerateGitVersionInformation` again.

On a related note, could you explain why we have the following:
```
<DefineConstants>COREFX_SUBSET</DefineConstants>
<GenerateAssemblyInfo>false</GenerateAssemblyInfo>
```

bkragl:
Three more points:

1. I think we should configure GitVersion for all projects. Otherwise the generated DLLs have assembly version `0.0.0.0`.
2. My understanding is that the non-.NET Core build is going to be deprecated, right? So eventually all `*NetCore*` configuration files should replace the old ones. In the meantime, should `Source/version.cs` still be maintained manually? I think it is not worth the effort to set up GitVersion for the legacy build.
3. I was working on adding a .NET Core build to Travis in PR #166. Currently it builds the Release configuration under Mono and .NET Core. Should I remove the Mono build from Travis?

michael-emmi:
Hey @bkragl !

> My suggestion would be to remove the line and use the default compilation items…

I’m not against this change, although I would say it is orthogonal to this PR, and is not strictly necessary due to the following:

> … Then we can enable GenerateGitVersionInformation again.

It seems there is no need to enable this, since we do not use it. Do you have a potential use?

> could you explain why we have the following:
> `<DefineConstants>COREFX_SUBSET</DefineConstants>`

Seems like this is only used here:
https://github.com/boogie-org/boogie/blob/master/Source/Core/CommandLineOptions.cs#L811-L816

As I recall, this block contains some functionality that is available in .NET Framework but not .NET Core. We can try removing this line in all projects besides Core (since others do not use). But again, seems fairly orthogonal to this PR.

> `<GenerateAssemblyInfo>false</GenerateAssemblyInfo>`

I believe this was added to prevent duplicate/conflicting symbols in assemblies generated from projects including other (Boogie) projects. I do not recall details.

Anyhow, since all of these issues seem orthogonal to this PR, I think we should address them separately, if you deem them important. They might fit into #166 for instance, or separate PRs on their own.

michael-emmi:
Hey @bkragl !

> I think we should configure GitVersion for all projects.

Agreed! I will add commits for that.

> My understanding is that the non-.NET Core build is going to be deprecated, right? So eventually all *NetCore* configuration files should replace the old ones. In the meantime, should Source/version.cs still be maintained manually? I think it is not worth the effort to set up GitVersion for the legacy build.

I am not against deprecating the non-.NET Core build. (Thought I wonder whether some current Boogie users might object — particularly those using Visual Studio, if they exist?) I agree that its probably not worth the effort to set up GitVersion for the legacy build. At the same time, maintaining `version.cs` would be a pain. (Also note that it hasn’t been maintained for years now.)

Maybe it is fine for Visual Studio and Mono users to go on their merry way with the same version number (according to `version.cs`) that has not changed in a long time, and the rest of us .NET Core users can use increasing version numbers.

> I was working on adding a .NET Core build to Travis in PR #166. Currently it builds the Release configuration under Mono and .NET Core. Should I remove the Mono build from Travis?

The more I think about it, the less I see a problem with deprecating Mono support. (Though I’m less clear about deprecating Visual Studio support.) Surely anybody who is willing to install Mono to run Boogie would be happy to use Dotnet Core instead, no?

I’m not against removing the Mono build from Travis. How about @zvonimir and @delcypher ?

bkragl:
> It seems there is no need to enable GenerateGitVersionInformation, since we do not use it. Do you have a potential use?

I don't have a use case now, but I would stick with the default configuration if possible. If somebody wants to use it in the future, they need to look for this option and perhaps don't know that it can be enabled without problems.

bkragl:
> Maybe it is fine for Visual Studio and Mono users to go on their merry way with the same version number (according to `version.cs`) that has not changed in a long time, and the rest of us .NET Core users can use increasing version numbers.

I agree with that. Maintaining `version.cs` would mainly be relevant if we wanted to include Standard .NET and Mono binaries in the official releases we are starting now. I'm against that.

> The more I think about it, the less I see a problem with deprecating Mono support. (Though I’m less clear about deprecating Visual Studio support.)

What do you mean by Visual Studio support? AFAIK, Visual Studio has .NET Core support and thus can open and compile `Source/Boogie-NetCore.sln`.

michael-emmi:
> What do you mean by Visual Studio support? AFAIK, Visual Studio has .NET Core support and thus can open and compile `Source/Boogie-NetCore.sln`.

I mean any user who is happily using the existing .Net Framework builds with Visual Studio. While I agree that they should be able to use our .Net Core builds in principle, as far as I am aware, nobody is testing that, right?

michael-emmi:
> > It seems there is no need to enable GenerateGitVersionInformation, since we do not use it. Do you have a potential use?
> 
> I don't have a use case now, but I would stick with the default configuration if possible. If somebody wants to use it in the future, they need to look for this option and perhaps don't know that it can be enabled without problems.

If you want to insist on enabling the default compile items (in another PR), I am not against rebasing this on to enable generating git version info.

bkragl:
> If you want to insist on enabling the default compile items (in another PR), I am not against rebasing this on to enable generating git version info.

I'll simultaneously upgrade to .NET Core 3.1, OK?

michael-emmi:
> > If you want to insist on enabling the default compile items (in another PR), I am not against rebasing this on to enable generating git version info.
> 
> I'll simultaneously upgrade to .NET Core 3.1, OK?

👍🏼

michael-emmi:
Hey @bkragl how’s this looking now?

I think it also makes sense to look into the following (in some other PR):
* Enabling assembly-info generation
* Removing unused `COREFX_SUBSET` constant from projects other than Core.
* Looking into whether these `cce.cs` files should sill exist (and be ignored).

michael-emmi:
@bkragl I’ve addressed the following in #180 to bring us closer to minimalist bliss.

> I think it also makes sense to look into the following (in some other PR):
> 
> * Enabling assembly-info generation
> * Removing unused `COREFX_SUBSET` constant from projects other than Core.
> * Looking into whether these `cce.cs` files should sill exist (and be ignored).

Issue #153, opened 2019-10-15
CIVL introduction procedures can have a layer range


dricketts:
I think introduction procedures are only supposed to have a single layer number, but the implementation currently allows a layer range:
```
var {:layer 0, 1} x: int;

procedure {:layer 0, 1} intro_x()
modifies x;
{
  x ≔ x + 1;
}

procedure {:yields} {:layer 1} update_x() 
{
  yield;
  call intro_x();
  yield;
}
```
If this is not a bug, then what does it mean for introduction procedures to have a layer range?

shazqadeer:
My conjecture is that an introduction procedure with a side-effect cannot be called at two different layers.  Do you have a counterexample that violates my conjecture? Please attach if you do.  I will look into the code in the meantime.

dricketts:
I think this is a counterexample:
```
var {:layer 0, 1} x: int;

procedure {:layer 0, 1} intro_x()
modifies x;
ensures x == old(x) + 1;
{
  x := x + 1;
}

procedure {:yields} {:layer 1} layer1() 
{
  yield;
  call intro_x();
  yield;
}

procedure {:yields} {:layer 0} {:refines "atomic_layer_0"} layer0()
{
  yield;
  call intro_x();
  yield;
}

procedure {:atomic} {:layer 1} atomic_layer_0()
modifies x;
{
  x := x + 1;
} 
```

shazqadeer:
@bkragl and I had a discussion about this issue.  It is indeed a bug which will be fixed as follows:
- require introduction procedures to be annotated by a single layer rather than a layer range
- requiring that if a yielding procedure P calls an introduction procedure Q, then the disappearing layer of P matches the layer annotation of Q

The side effect of an introduction procedure will be introduced into a caller only at the layer of the introduction procedures; at other (lower) layers the call will be erased.

I will try to get this bug fixed soon.  In the meantime, please obey the restrictions mentioned above manually.  Thank you for pointing out the issue.

shazqadeer:
My recent PR provides a fix for this issue.  However, the fix is not quite as I described above.  An introduction procedure will be allowed to have a layer range in general but if it modifies a global variable then the singleton requirement will be applied.

PR #175, opened 2019-12-15
[CIVL] Fix for introduction procedures.

shazqadeer:
An introduction procedure introduces computation into a CIVL program.  An introduction procedure  that does not modify any global variable and has no output variables is called a lemma introduction procedure.  A call to such a lemma procedure is used to inject trusted or untrusted facts into the verification condition.

A non-lemma introduction procedures may only be called at the disappearing layer of a yielding procedure.  A lemma procedure may be called at many layers but the call command must be annotated with the layer/s at which the call is happening.

This PR should fix https://github.com/boogie-org/boogie/issues/153.

### DIFF
--- a/Source/Concurrency/CivlCoreTypes.cs
+++ b/Source/Concurrency/CivlCoreTypes.cs
@@ -221,14 +221,14 @@ namespace Microsoft.Boogie
     {
         public Procedure proc;
         public LayerRange layerRange;
-        public bool isLeaky;
 
-        public IntroductionProc(Procedure proc, LayerRange layerRange, bool isLeaking)
+        public IntroductionProc(Procedure proc, LayerRange layerRange)
         {
             this.proc = proc;
-            this.isLeaky = isLeaking;
             this.layerRange = layerRange;
         }
+
+        public bool IsLemma => proc.Modifies.Count + proc.OutParams.Count == 0;
     }
 
     /// <summary>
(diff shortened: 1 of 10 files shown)


Source/Concurrency/CivlTypeChecker.cs
```diff
@@ -1284,35 +1291,67 @@ private void VisitYieldingProcCallCmd(CallCmd call, YieldingProc callerProc, Yie
                 }
             }
 
-            private void VisitIntroductionProcCallCmd(CallCmd call, YieldingProc callerProc, IntroductionProc calleeProc)
+            private void VisitIntroductionProcCallCmd(CallCmd call, YieldingProc callerProc,
+                IntroductionProc calleeProc)
             {
                 if (!calleeProc.layerRange.Contains(callerProc.upperLayer))
```
bkragl: I think this first if block can be removed. It only applies to the case `!calleeProc.IsLemma` where the same check is already done.

Source/Concurrency/CivlCoreTypes.cs
```diff
@@ -221,14 +221,17 @@ public class IntroductionProc
...
             this.proc = proc;
-            this.isLeaky = isLeaking;
             this.layerRange = layerRange;
         }
+
+        public bool IsLemma()
+        {
+            return proc.Modifies.Count + proc.OutParams.Count == 0;
+        } 
     }
```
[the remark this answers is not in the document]
shazqadeer: Done.

Source/Concurrency/CivlTypeChecker.cs
shazqadeer: Done.

PR #178, opened 2019-12-18
Upgrade unit tests to NUnit 3 and make them work under .NET Core

### LINKED ISSUE #165: Boogie tests are incompatible with NUnit versions 3.0 onwards
Various test annotation features (e.g. Assert.IsNullOrEmpty, ExpectedExceptionAttribute) were removed from NUnit quite some versions ago. When a version of NUnit comes with Visual Studio (on Windows), this is tricky to work around (at least, in the tool and not on the command-line). It might be good to migrate to a newer NUnit.

See [this page](https://github.com/nunit/docs/wiki/Breaking-Changes) for details of the removed features.

bkragl:
To run the unit tests under .NET Core, execute `dotnet test Source/Boogie-NetCore.sln`.

Also, I removed all the `AssemblyInfo.cs` files that were auto-generated for some projects but never maintained. Instead, in the legacy build all projects now consistently link `Source/version.cs`.

Fixes #165


.travis.yml
```diff
@@ -15,10 +15,9 @@ install:
   - unzip z3*.zip
   # NuGet is a little flakey in legacy TravisCI, use travis_retry command to retry the command if it fails
   - travis_retry nuget restore ${TRAVIS_SOLUTION}
+  - travis_retry nuget install -Version 3.10.0 NUnit.Console -OutputDirectory Source/packages/
```
michael-emmi: Why is it that this should be installed manually, rather than as a dependency that is restored by the previous line?

bkragl: This only concerns the Mono build and was already like this before. I just replaced the obsolete NUnit.Runners with NUnit.Console.

I guess the reason is that the previous line only restores the NUnit package needed to compile the unit test projects. This line installs the command line tool that runs the tests.

michael-emmi: Oh I see, and it’s right there in Travis’s own instructions too: https://docs.travis-ci.com/user/languages/csharp/#nunit

Just documenting for future reference then :-)

PR #180, opened 2019-12-19
Simplify .Net Core project files.

michael-emmi:


michael-emmi:
Easily undone already :-)

PR #181, opened 2019-12-20
Add implementations of GetHashCode to classes that override Equals

bkragl:
The following classes are affected:
* TypeCoercion
* QuantifierExpr
* BasicType
* CtorType

### DIFF
--- a/Source/Core/AbsyExpr.cs
+++ b/Source/Core/AbsyExpr.cs
@@ -2249,7 +2249,9 @@ namespace Microsoft.Boogie {
       }
     }
 
-
+    public override int GetHashCode() {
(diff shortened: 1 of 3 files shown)


michael-emmi:
@bkragl is this ready to merge?

bkragl:
I think so, but please double-check.

For QuantifierExpr and CtorType the behavior should be the same as before, just explicitly calling ` base.GetHashCode()`. For TypeCoercion and BasicType this is actually a fix that establishes the property that equal objects (according to method `Equals`) must have the same hash code.

Mainly, this eliminates compiler warnings :)

michael-emmi:
lgtm

PR #182, opened 2019-12-24
Adding workflow to bump version tags.

michael-emmi:
This workflow will create and push a new tag each time commits are pushed to the master branch (including PR merges).

By default, the created tag increments the patch version number from the previous tag. For example, if the last tagged commit were `v2.4.3`, then pushing to master would tag the latest commit with `v2.4.4`.

If incrementing minor or major number is desired instead of patch, simply add #minor or #major to the first line of the commit message. For instance:
```
Adding the next greatest feature. #minor
```
If the last tagged commit were `v2.4.3`, then pushing this commit would generate the tag `v2.5.0`.

For pull-request merges, if minor or major version increments are desired, a squash merge is required so that the first line of the merge commit message can be changed to include #minor or #major.

Note that on each push to `master`, the following would happen:
* A travis build for `master` is triggered.
* This GitHub workflow is also triggered.
* Once this workflow pushes a new tag `vX.Y.Z`, another travis build for `vX.Y.Z` is triggered.
* The travis build for `vX.Y.Z` in `Release` configuration publishes releases to GitHub and NuGet.org.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,18 @@
+name: Bump
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
+    - name: Bump version and push tag
+      uses: anothrNick/github-tag-action@master
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        WITH_V: true
+        DEFAULT_BUMP: patch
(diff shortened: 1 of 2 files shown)


shazqadeer:
@michael-emmi : You could add the documentation in the PR summary to README.md.

michael-emmi:
@zvonimir @shazqadeer good idea; done!

bkragl:
This looks nice! The only thing I find quite unfortunate is that squash merges are required to increment minor/major. I usually like to preserve the development history of a new feature.

Although, doing a regular merge on the command line would still work, because git will ask for a commit message for the merge commit, right?

michael-emmi:
> Although, doing a regular merge on the command line would still work, because git will ask for a commit message for the merge commit, right?

Should work, so long as you can control the commit message for the merge commit and add #minor or #major on the first line.

bkragl:
Actually the GitHub interface also let's you edit the commit message for regular merges. When you select "Create a merge commit" and click "Merge pull request" you'll see something like this:
image
So squashing is not a requirement.

michael-emmi:
Oooh good point @bkragl I missed that!

Updates readme to reflect that.

PR #183, opened 2019-12-27
Restore TotalUserTime field removed by 714d5a0.

michael-emmi:
The `TotalUserTime` field was removed from `Microsoft.Boogie.SMTLib.SMTLibProcess` by 714d5a0 because it was causing a crash. However, other projects, [e.g., Corral](https://github.com/boogie-org/corral/blob/master/source/Driver.cs#L444), were using this field, so that commit blocks them from updating Boogie.

This commit solves the original issue, by not invoking the getter for `UserProcessorTime` if the given process has already exited.

### DIFF
--- a/Source/Provers/SMTLib/SMTLibProcess.cs
+++ b/Source/Provers/SMTLib/SMTLibProcess.cs
@@ -190,8 +190,21 @@ namespace Microsoft.Boogie.SMTLib
       }
     }
 
+
+    // NOTE: this field is used by Corral.
+    // https://github.com/boogie-org/corral/blob/master/source/Driver.cs
+    public static System.TimeSpan TotalUserTime = System.TimeSpan.Zero;
+
     public void Close()
     {
+      try {
+        TotalUserTime += prover.UserProcessorTime;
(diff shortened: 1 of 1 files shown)


bkragl:
Could there be a race condition? I.e., that the process exited after checking HasExited, but before accessing UserProcessorTime.

michael-emmi:
> Could there be a race condition? I.e., that the process exited after checking HasExited, but before accessing UserProcessorTime.

Good point; I hadn’t seen it happen, but seems like it could in principle.

Just pushed another approach: try it and catch the exception if it occurs.

michael-emmi:
@bkragl are you ok with this now?

michael-emmi:
It’s not clear to me whether there’s a way to prevent this exception, since it seems to be possible so long as the child process exits before the call happens — though I haven’t actually looked into the details — which is not up to the parent process. Plus, it’s probably not worth the time to look into. My intent here is just to patch up breakage due to this missing field.

So I‘ll just merge this and we can open another issue if we wish to pursue a better fix.

Issue #73, opened 2017-09-11
Discussing (and cleaning up) default options passed to Z3


zvonimir:
By default, Boogie passes the following options to Z3 through queries:
```
(set-option :print-success false)
(set-info :smt-lib-version 2.0)
(set-option :AUTO_CONFIG false)
(set-option :pp.bv_literals false)
(set-option :MODEL.V2 true)
(set-option :smt.PHASE_SELECTION 0)
(set-option :smt.RESTART_STRATEGY 0)
(set-option :smt.RESTART_FACTOR |1.5|)
(set-option :smt.ARITH.RANDOM_INITIAL_VALUE true)
(set-option :smt.CASE_SPLIT 3)
(set-option :smt.DELAY_UNITS true)
(set-option :NNF.SK_HACK true)
(set-option :smt.MBQI false)
(set-option :smt.QI.EAGER_THRESHOLD 100)
(set-option :TYPE_CHECK true)
(set-option :smt.BV.REFLECT true)
```
I first used Boogie probably 10 years ago or so, and some of these options if I remember correctly were already there :). I wonder if it is time to revisit these given that Z3 changed a lot since then. Ideally, it would be good to let Z3 choose what the optimal setup should be by enabling `AUTO_CONFIG`.
Of course, such a change might break some regressions by causing time outs, but it might be a good move in the long run.
I should also add that these options are I think geared towards particular kinds of benchmarks (quantifier-heavy, etc.) and they are likely not a good setup for other usage scenarios. So I think that going with default Z3 options would be a good thing to do, while users can on top of that customize by adding options that are good for their benchmarks.
Thoughts?

shazqadeer:
@zvonimir :  Looks like there is this old issue that has resurfaced.  Let us discuss here and do the necessary cleanup.

@bkragl : FYI.

zvonimir:
Boogie is currently used in more scenarios than a decade ago, and various usage scenarios might require different Z3 parameters to be set. So I wonder if the code that sets the parameters should be refactored to reflect this. One idea I has in this direction was that by default no Z3 parameters get set except for the absolutely necessary ones (such as model generation), which would leave Z3 to select its default parameters based on the generated query. Then, we could have configuration files from which other parameters would be loaded as needed (Daisy would have one, Corral one, maybe CIVL one).

[...]

PR #186, opened 2019-12-30
Updating Z3 model_compress parameter.

michael-emmi:
* Z3 4.8.7 consolidates model.compact and model_compress.


michael-emmi:
@bkragl any ideas what should be done when test output varies from system to system? I’m getting different test failures on my own machine…

bkragl:
On my system (Ubuntu / .NET Core) with Z3 4.8.4 I see the same four tests failing as in Travis:
```
test15/CaptureState.bpl
test15/IntInModel.bpl
test15/ModelTest.bpl
test15/NullInModel.bpl
```
All of them look benign. Just some small difference in the model output.

With Z3 4.8.7 I see the following tests fail:
```
aitest9/TestIntervals.bpl
extractloops/t3.bpl
livevars/bla1.bpl
livevars/stack_overflow.bpl
test15/CaptureState.bpl
test15/ModelTest.bpl
test2/Rlimitouts0.bpl
test7/MultipleErrors.bpl
```

@michael-emmi: Could this be the difference that you see? In any case, also there the failures are due to slightly different models/error traces (except `test2/Rlimitouts0.bpl` which tests different resource limits). I would say we could adapt the expected output for these tests and switch the testing in Travis to Z3 4.8.7.

In general I don't have a good strategy to make testing with different configurations (in particular different Z3 versions) stable.

@michael-emmi: In case you did not know, there is #73 where we started to discuss a cleanup of the Z3 options that Boogie sets by default. There is a branch `z3-option-cleanup` where I removed all but some minimal set of necessary options for testing (I'll rebase it onto the latest master to get the current .NET Core build).

michael-emmi:
@bkragl I wasn’t aware of #73 but glad I am now — thanks!

> Could this be the difference that you see?

Yes, those are precisely the differences I see.

So should this PR fix the expected test output so that they all pass on Travis with Z3 4.8.7?

bkragl:
> So should this PR fix the expected test output so that they all pass on Travis with Z3 4.8.7?

I think so. We could also fix the tests for Z3 4.8.4, but to me upgrading right away makes more sense.

michael-emmi:
Oh crap I merged by accident…

PR #187, opened 2020-01-04
Extend builtin support to user-defined types.

cbarrettfb:
This commit extends support for the builtin attribute to user-defined types.

It also adds two new attributes.  The builtindecl attribute allows the user
to specify their own SMT-LIB type declaration for a user-defined type.
The dependson attribute is used to indicate when a datatype depends on another
datatype when Boogie cannot figure this out itself (which may happen if
builtin's are used).

Examples of how to use these attributes are in Test/sequence.

### DIFF
--- a/Source/Core/AbsyType.cs
+++ b/Source/Core/AbsyType.cs
@@ -3238,9 +3238,10 @@ Contract.Requires(that != null);
       return QKeyValue.FindBoolAttribute(Decl.Attributes, "datatype");
     }
 
-    // This attribute is used to provide a user-specified SMT-LIB declaration for the type.
-    public string GetBuiltinDecl() {
-      return this.Decl.FindStringAttribute("builtindecl");
+    // This attribute is used to tell Boogie that this type is built into SMT-LIB and should
+    // be represented using the provided string (and also does not need to be explicitly declared).
+    public string GetBuiltin() {
+      return this.Decl.FindStringAttribute("builtin");
     }
 
     // This attribute can be used to tell Boogie that a datatype depends on another datatype
@@ -3396,7 +3397,7 @@ Contract.Requires(that != null);
       //Contract.Requires(stream != null);
       stream.SetToken(this);
       // If this type has a "builtin" attribute, use the corresponding user-provided string to represent the type.
-      string builtin = this.Decl.FindStringAttribute("builtin");
+      string builtin = GetBuiltin();
       if (builtin != null) {
         stream.Write(builtin);
       } else {
--- a/Source/Core/AbsyType.cs
+++ b/Source/Core/AbsyType.cs
@@ -3238,6 +3238,18 @@ Contract.Requires(that != null);
       return QKeyValue.FindBoolAttribute(Decl.Attributes, "datatype");
     }
 
+    // This attribute is used to provide a user-specified SMT-LIB declaration for the type.
+    public string getBuiltinDecl() {
+      return this.Decl.FindStringAttribute("builtindecl");
+    }
+
+    // This attribute can be used to tell Boogie that a datatype depends on another datatype
+    // in case Boogie can't figure this out itself (as may happen, for example when a type
+    // has the ":builtin" attribute).
+    public string getTypeDependency() {
+      return this.Decl.FindStringAttribute("dependson");
+    }
+
     //-----------  Cloning  ----------------------------------
     // We implement our own clone-method, because bound type variables
     // have to be created in the right way. It is /not/ ok to just clone
@@ -3383,7 +3395,13 @@ Contract.Requires(that != null);
     public override void Emit(TokenTextWriter stream, int contextBindingStrength) {
       //Contract.Requires(stream != null);
       stream.SetToken(this);
-      EmitCtorType(this.Decl.Name, Arguments, stream, contextBindingStrength);
+      // If this type has a "builtin" attribute, use the corresponding user-provided string to represent the type.
+      string builtin = this.Decl.FindStringAttribute("builtin");
+      if (builtin != null) {
+        stream.Write(builtin);
+      } else {
+        EmitCtorType(this.Decl.Name, Arguments, stream, contextBindingStrength);
+      }
     }
 
     internal static void EmitCtorType(string name, List<Type> args, TokenTextWriter stream, int contextBindingStrength) {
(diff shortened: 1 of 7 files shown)


Source/Core/AbsyType.cs
```diff
@@ -3160,6 +3160,18 @@ public CtorType(IToken/*!*/ token, TypeCtorDecl/*!*/ decl, List<Type>/*!*/ argum
       return QKeyValue.FindBoolAttribute(Decl.Attributes, "datatype");
     }
 
+    // This attribute is used to provide a user-specified SMT-LIB declaration for the type.
+    public string getBuiltinDecl() {
+      return this.Decl.FindStringAttribute("builtindecl");
```
shazqadeer: Boogie code follows the naming convention of starting method names with Capital letters.

Source/Core/AbsyType.cs
```diff
@@ -3160,6 +3160,18 @@ public CtorType(IToken/*!*/ token, TypeCtorDecl/*!*/ decl, List<Type>/*!*/ argum
...
+    }
+
+    // This attribute can be used to tell Boogie that a datatype depends on another datatype
+    // in case Boogie can't figure this out itself (as may happen, for example when a type
+    // has the ":builtin" attribute).
+    public string getTypeDependency() {
```
shazqadeer: Same here.

Test/sequences/intseq.bpl
```diff
@@ -0,0 +1,22 @@
...
+procedure test()
+{
+  var s: IntSeq;
+  
+  s := IntSeqConcat(EmptyIntSeq(), IntSeqUnit(0));
+  s := IntSeqConcat(s, IntSeqUnit(1));
+  s := IntSeqConcat(s, IntSeqUnit(2));
+  assert IntSeqLen(s) == 3;
+  assert IntSeqExtract(s, 1, 2) == IntSeqConcat(IntSeqUnit(1), IntSeqUnit(2));
+  assume IntSeqNth(s, 1) == 1;
```
shazqadeer: Did you intend the "assume" to be an "assert"?

Test/sequences/intseq_datatype.bpl
```diff
@@ -0,0 +1,57 @@
+// RUN: %boogie "%s" > "%t"
+// RUN: %diff "%s.expect" "%t"
+
+type {:datatype} Value;
+
+function {:constructor} Integer(i: int): Value;
+function {:constructor} Vector(v: ValueArray): Value;
+
+type {:builtindecl ""} {:builtin "(Seq T@Value)"} ValueSeq;
```
shazqadeer: What is the empty string for?

Source/Core/AbsyType.cs
cbarrettfb: Fixed

Source/Core/AbsyType.cs
cbarrettfb: Fixed

Test/sequences/intseq.bpl
cbarrettfb: Good catch - fixed.

Test/sequences/intseq_datatype.bpl
cbarrettfb: This is to tell Boogie to use the provided string in place of its usual declaration for this type.  In this case, we don't want any declaration.  Without this, Boogie produces:
(declare-sort (Seq T@Value))
which of course makes no sense to z3.

Test/sequences/intseq.bpl
```diff
@@ -0,0 +1,23 @@
+// RUN: %boogie "%s" > "%t"
+// RUN: %diff "%s.expect" "%t"
+
+type {:builtindecl ""} {:builtin "(Seq Int)"} IntSeq;
```
shazqadeer: What is the purpose of the empty string in the "builtindecl" attribute?  Under what circumstances will this string be nonempty?

shazqadeer: Upon more thinking, I don't even understand the purpose of "builtindecl".  You already have a "builtin" attribute on the type declaration.  So it is already clear that the uninterpreted type IntSeq is representing the concrete type (Seq Int) obtained from the string associated with the "builtin" attribute.

Test/sequences/intseq_datatype.bpl
shazqadeer: I see.  The parameterized sort Seq is "pre-declared".  So Z3 will scream if we try to re-declare it.  Please see my comment above that "builtindecl" itself might be redundant.

shazqadeer: I would have attempted something like the following declaration for ValueSeq.

type {:builtin "Seq", "Value"} ValueSeq;

or in general

type {:builtin "SortName", "SortArg0", "SortArg1", ...}.

This will allow you to eliminate the "dependson" attribute as well because the dependency of ValueSeq on Value can now be calculated.

Test/sequences/intseq.bpl
cbarrettfb: My thinking was as follows.  A type declaration in Boogie has two effects on the generated SMT-LIB.  First, it creates a declaration for the type, something like:
(declare-sort IntSeq)
Second, it allows IntSeq to appear in other declarations and formulas.

So, the ":builtindecl" attribute tells Boogie how to declare this built-in sort.  In this case, we don't need anything, but you could imagine that in other cases you might need a particular declaration command.  The ":builtin" attribute tells Boogie how to refer to the sort when it is used in other formulas.

This seemed like a clean way to do it, but I'm open to other suggestions.

cbarrettfb: Note that without the ":builtindecl" change, Boogie will go ahead and generate:
(declare-sort (Seq Int))
which of course make no sense and makes z3 unhappy.

An obvious alternative would be to just omit declarations for any type that has a ":builtin" attribute.  That would work fine here - I just wasn't sure that would always be the right call for a ":builtin" type.

shazqadeer: Your last comment was helpful to me.  It seems to me that a "builtin" type is naturally built in and therefore does not require a declaration of its own.  So omitting the declaration when "builtin" is used on a type makes sense.  I suggest dropping "builtin-decl" from this PR.  We can add it later if we need it.  I will go ahead and approve the PR assuming you will make this change.  Thanks a lot for this valuable addition to Boogie.

cbarrettfb: OK - this has been done.  I think this is a fine way to proceed for now.

PR #189, opened 2020-01-20
Using old Z3 model_compress parameter for Dafny.

michael-emmi:


michael-emmi:
@RustanLeino this ought to fix the failing Dafny test you mentioned.

RustanLeino:
Confirmed. Thanks, @michael-emmi !

PR #196, opened 2020-02-15
Add support for SMT-LIB function definitions

cbarrettfb:
This PR adds support for a function attribute ":define".  Adding this attribute to a function generates a SMT-LIB "define-fun" command instead of a definition via universal quantification.  Note that the function must not be recursive.

Example in Test/functiondefine.


zvonimir:
This looks like a great addition! I do have a quick question though: how is this better/different than the `:inline` attribute that already exists?
I guess I understand how it is different at the level of the VC, but I am curious if solvers somehow handle `define-fun` more efficiently internally?
Thanks!

cbarrettfb:
Good question Zvonimir.  At the level of the VC, using :inline eagerly substitutes the function everywhere in the SMT formula.  Using :define creates a define-fun and lets the solver decide how to handle things.  Somewhat surprisingly, Z3 slows down quite a bit (3-4X) in some cases when using :inline, but the same examples are the same or even faster with :define.  I'm not sure exactly why - I suspect Z3 may be doing some lazy definition expansion, but in any case, using definitions seems to give the solver more flexibility.

Source/Core/Parser.cs
```diff
@@ -428,6 +428,8 @@ public override int ComputeHashCode()
 		 // generate either an axiom or a function body
 		 if (QKeyValue.FindBoolAttribute(kv, "inline")) {
 		   func.Body = definition;
+		 } else if (QKeyValue.FindBoolAttribute(kv, "define")) {
+		   ds.Add(func.CreateFunctionDefinition(definition, kv));
```
shazqadeer: @bkragl : I recall that you explained that Parser.cs is a generated file and should not be modified.  Could you please add the appropriate steps here when modifying BoogiePL.atg?

Source/Core/Absy.cs
```diff
@@ -2866,11 +2868,17 @@ public Function(IToken tok, string name, List<TypeVariable> typeParams, List<Var
...
       }
       DefinitionAxiom = new Axiom(tok, def);
       return DefinitionAxiom;
     }
+
+    public Axiom CreateFunctionDefinition(Expr definition, QKeyValue kv = null) {
+      CreateDefinitionAxiom(definition, kv, true);
+      return DefinitionAxiom;
+    }
+
```
shazqadeer: It seems we are creating an axiom even when the programmer is indicating a function definition using :define.  Is there some peculiarity of Boogie that is forcing this design?

Source/VCExpr/Boogie2VCExpr.cs
```diff
@@ -397,14 +397,15 @@ private class VariableMapping<VarKind> : ICloneable {
...
               flipContextForArg0 = true;
-          else if (oper.Op == BinaryOperator.Opcode.Iff) {
-              Expr one = new NAryExpr(node.tok, new BinaryOperator(node.tok, BinaryOperator.Opcode.Imp), new List<Expr> { node.Args[0], node.Args[1] });
-              Expr two = new NAryExpr(node.tok, new BinaryOperator(node.tok, BinaryOperator.Opcode.Imp), new List<Expr> { node.Args[1], node.Args[0] });
-              NAryExpr cmpd = new NAryExpr(node.tok, new BinaryOperator(node.tok, BinaryOperator.Opcode.And), new List<Expr> { one, two });
-              TypecheckingContext tc = new TypecheckingContext(null);
-              cmpd.Typecheck(tc);
-              return TranslateNAryExpr(cmpd);
-          }
+          // This code is no longer needed - left over from when SMT-LIB didn't support Boolean equality
```
shazqadeer: Just delete it.

shazqadeer:
@RustanLeino : It would be helpful if you could take a look at this PR.  @cbarrettfb is adding a very useful Boogie feature.

Test/functiondefine/fundef.bpl
```diff
@@ -0,0 +1,19 @@
...
+procedure test(x:int) returns (r:int)
+requires x > 0;
+ensures r > 0;
+{
+  if (foo(x)) {
+    r := foo2(x);
+  } else {
+    r := 1;
+  }
+}
```
shazqadeer: It is complaining about no newline at end of file.

Source/Core/Parser.cs
bkragl: `Parser.cs` is generated, but we still keep it in the repository (I guess to avoid parser generation as part of the build process). I see that this change comes from `BoogiePL.atg` so everything is good.

shazqadeer:
If the summary provided by @bkragl  is correct, wouldn't it be better to have the field isFunctionDefinition in the axiom?

Source/Provers/SMTLib/ProverInterface.cs
```diff
@@ -347,15 +388,80 @@ private void PrepareCommon()
...
+          axiomList.Add(axioms);
+
+        // Separate function definitions from other axioms.  Function definitions must be processed
+        // first; otherwise, processing an axiom that uses the function definition can create a declaration
+        // resulting in the symbol being introduced twice (a definition and a declaration).
+        Dictionary<Function, VCExpr> functionDefinitionMap = new Dictionary<Function, VCExpr>();
+        Stack<Function> functionDefs = new Stack<Function>();
+        Stack<VCExpr> otherAxioms = new Stack<VCExpr>();
+        for (int i = axiomList.Count-1; i >= 0; i--) {
+          Function f = IsFunctionDef(axiomList[i]);
```
RustanLeino: Following the similar C# operators, I would expect `is` to return a boolean (`true` if it is, and `false` if it ain't) and `as` to return a reference (the original value if it is, and `null` if it ain't). Therefore, I suggest renaming this function to `AsFunctionDef`.

Source/Provers/SMTLib/ProverInterface.cs
```diff
@@ -347,15 +388,80 @@ private void PrepareCommon()
...
+            dependenciesComputed.Add(f);
+          }
+        }
+
+	// Now process the rest of the axioms.
+        while (otherAxioms.Count > 0) {
+          VCExpr expr = otherAxioms.Pop();
+          var str = VCExpr2String(expr, -1);
+          if (str != "true")
+            AddAxiom(str);
```
RustanLeino: Coding style: always use curly braces around the `if` branches

Source/Provers/SMTLib/ProverInterface.cs
```diff
@@ -382,7 +488,9 @@ private void FlushAxioms()
       TypeDecls.Clear();
       foreach (string s in Axioms) {
         Contract.Assert(s != null);
-        if (s != "true")
+	if (s.StartsWith("(define-fun"))
+	  SendCommon(s);
+        else if (s != "true")
           SendCommon("(assert " + s + ")");
```
RustanLeino: Use braces around the `if` alternatives.

Source/Provers/SMTLib/SMTLibLineariser.cs
```diff
@@ -322,8 +322,21 @@ public bool Visit(VCExprQuantifier node, LineariserOptions options)
...
-        string kind = node.Quan == Quantifier.ALL ? "forall" : "exists";
-        wr.Write("({0} (", kind);
+        VCQuantifierInfos infos = node.Infos;
+        if (infos.isFunctionDefinition) {
+	  wr.Write("(define-fun ");
+	  var naryExpr = node.Body as VCExprNAry;
+	  Contract.Assert(naryExpr != null);
+	  var funCall = naryExpr[0] as VCExprNAry;
+	  Contract.Assert(funCall != null);
+	  VCExprBoogieFunctionOp op = (VCExprBoogieFunctionOp)funCall.Op;
```
RustanLeino: indent

Source/Provers/SMTLib/SMTLibLineariser.cs
```diff
@@ -322,8 +322,21 @@ public bool Visit(VCExprQuantifier node, LineariserOptions options)
...
+          Contract.Assert(op != null);
+          wr.Write(Namer.GetQuotedName(op.Func, op.Func.Name));
+	  wr.Write(" (");
```
RustanLeino: indent

Source/Provers/SMTLib/SMTLibLineariser.cs
```diff
@@ -335,7 +348,15 @@ public bool Visit(VCExprQuantifier node, LineariserOptions options)
 
         wr.Write(") ");
 
-        VCQuantifierInfos infos = node.Infos;
+	if (infos.isFunctionDefinition) {
+	  var naryExpr = node.Body as VCExprNAry;
+	  Contract.Assert(naryExpr != null);
+	  wr.Write(" {0} ", TypeToString(naryExpr[0].Type));
+          Linearise(naryExpr[1], options);
+          wr.Write(")");
```
RustanLeino: Indentation is messed about from L351

Test/functiondefine/fundef.bpl
```diff
@@ -0,0 +1,19 @@
...
+// RUN: %diff "%s.expect" "%t2"
+
+function {:define} foo(x:int) returns(bool)
+  { foo2(x) > 0 }
+function {:define} foo2(x:int) returns(int)
+  { x + 1 }
+
+procedure test(x:int) returns (r:int)
+requires x > 0;
+ensures r > 0;
```
RustanLeino: indent

RustanLeino:
I suggest considering the design of this feature from the Boogie perspective. Is the following a good way to present it to the Boogie user?

* If a function is declared without a body, it is just an uninterpreted function. Of course, you can add axioms about it if you wish.

* If the function is given a body, then Boogie gives you three options:

  - Macro: The definition can be inlined in each place the function occurs, like a macro. This is illegal to do for recursively defined functions. Because the function definition is inlined, the function symbol itself is not part of the VC; this means that any use of the function in a trigger will also be replaced by the definition (which may produce an illegal trigger, where the function form looked okay as a trigger).

  - Axiom: The definition can be turned into a universally quantified axiom. By giving the definition in this way, it looks nice and localized in the Boogie program, and it saves you the trouble of generating the parameters once for the function and once for the axiom.

  - Lazy Expansion: If the underlying prover supports it, then the responsibility of equating a use of a function with its definition can be done during proving. This is similar to Axiom, but allows the prover to use different heuristics for when to bring the function definition into play. Like Macro, this option is illegal for recursively defined functions. (Does the function symbol stay around? In particular, what happens with a Lazy Expanded function that's used in a trigger?)

What is the best way for a Boogie user to select between these options? Currently, we have the {:inline} attribute, the new {:define} attribute, and probably the former also looks at some command-line options. It would be nice to have a clean and understandable story here.

shazqadeer:
>   * Axiom: The definition can be turned into a universally quantified axiom. By giving the definition in this way, it looks nice and localized in the Boogie program, and it saves you the trouble of generating the parameters once for the function and once for the axiom.

@RustanLeino : If the function definition is recursive and this axiom option is applied, it seems to me that it is possible for an inconsistent axiom to be generated.  Have you considered outlawing recursion altogether in function definitions, at least until termination check for functions has been implemented in Boogie?

Source/Core/Absy.cs
cbarrettfb: Yes.  At first, I tried an implementation that just stores the function definition as a special field in the Function class.  However, such functions must ultimately be translated to VCExprs. 
 This requires adding a new type of VCExpr for function definitions.  While conceptually clean, this ends up being a very heavy and error-prone change, as it requires changing all of the VCExpr visitors (there are like 10-20 of them).  I thought it was maybe not worth it when I could instead recover the function from the axiom quite easily and not have to touch the VCExpr class heirarchy.  I'm open to ideas if you think there is a better way.

Source/VCExpr/Boogie2VCExpr.cs
cbarrettfb: OK - for clarity I've put this in a separate PR.

Test/functiondefine/fundef.bpl
cbarrettfb: Fixed

Source/Core/Absy.cs
```diff
@@ -1527,6 +1527,7 @@ public abstract class DeclarationContracts : Declaration {
 
   public class Axiom : Declaration {
     private Expr/*!*/ expression;
+    public bool isFunctionDefinition;
```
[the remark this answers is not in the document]
cbarrettfb: Done

Source/Core/Absy.cs
```diff
@@ -2838,7 +2840,7 @@ public Function(IToken tok, string name, List<TypeVariable> typeParams, List<Var
       return visitor.VisitFunction(this);
     }
 
-    public Axiom CreateDefinitionAxiom(Expr definition, QKeyValue kv = null) {
+    public Axiom CreateDefinitionAxiom(Expr definition, QKeyValue kv = null, bool isFuncDef=false) {
```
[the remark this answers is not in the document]
cbarrettfb: Done

Source/Core/AbsyQuant.cs
```diff
@@ -593,30 +593,30 @@ public Trigger(IToken/*!*/ tok, bool pos, IEnumerable<Expr>/*!*/ tr, Trigger nex
 
   public class ForallExpr : QuantifierExpr {
     public ForallExpr(IToken/*!*/ tok, List<TypeVariable>/*!*/ typeParams,
-                      List<Variable>/*!*/ dummies, QKeyValue kv, Trigger triggers, Expr/*!*/ body, bool immutable=false)
-      : base(tok, typeParams, dummies, kv, triggers, body, immutable) {
+                      List<Variable>/*!*/ dummies, QKeyValue kv, Trigger triggers, Expr/*!*/ body, bool immutable=false, bool isFuncDef=false)
```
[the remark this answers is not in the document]
cbarrettfb: Fixed

Source/Provers/SMTLib/ProverInterface.cs
```diff
@@ -24,6 +24,31 @@
...
+
+    public List<Function> Collect(VCExpr expr) {
+      Contract.Requires(expr != null);
+      functionList = new List<Function>();
+      Traverse(expr, true);
+      return functionList;
+    }
+
+    public override bool Visit(VCExprNAry node, bool arg) {
+      Contract.Requires(node != null);
```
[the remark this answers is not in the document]
cbarrettfb: Ah ok - I copied this from TypeDeclCollector.  Removed.

At least the Contracts are good documentation!

Source/Provers/SMTLib/ProverInterface.cs
cbarrettfb: Done

Source/Provers/SMTLib/ProverInterface.cs
cbarrettfb: Fixed here and in a few other places.

Source/Provers/SMTLib/ProverInterface.cs
cbarrettfb: Fixed

Source/Provers/SMTLib/SMTLibLineariser.cs
cbarrettfb: Oops - my emacs was set up to insert tabs instead of spaces.  Fixed.

Source/Provers/SMTLib/SMTLibLineariser.cs
cbarrettfb: Fixed

Source/Provers/SMTLib/SMTLibLineariser.cs
cbarrettfb: Fixed

Test/functiondefine/fundef.bpl
cbarrettfb: Done

cbarrettfb:
@bkragl 
> First, let me try to summarize what is going on to see if my understanding is correct. When a function is decorated with `{:define}` there is still a quantified axiom generated as for "regular" functions. However, the quantifier expression (not the axiom) is marked as a function definition, such that it can be treated differently (and actually not be emitted as quantified formula to the SMT solver) in the VC generation.

Yes, that is correct.

> 1. My first thought was that it would be nice if no axiom is generated for function definitions (what @shazqadeer also noted above), but now I think it is actually good because there could be prover backends that do not support function definitions. However, what is the reason that the quantifier expression has to be marked and not the axiom? In class `Axiom` I see the field `isFunctionDefinition`, but it is unused. Can it be removed?

That is a potential benefit I did not consider, yes.  The main reason is what I said in my response to Shaz.  Adding another VCExpr subclass seems unnecessarily complicated and error-prone.  The IsFuncionDefinition field *is* used.  It's used by the QuantifierExpr constructor in AbsyQuant.cs and in fact, this is the only way I could find to communicate downstream the information that this should be treated as a definition.

> 2. I think it would be better if `isFunctionDefinition` would not be a field of `QuantifierExpr` and thus appear as optional parameter in all the constructors of it and `ForallExpr`. Instead, can we have a set that stores all `ForallExpr` objects related to function definitions, such that `Boogie2VCExprTranslator.GenerateQuantifierInfos` can query it? @shazqadeer, what would be a good place to keep such a set?

I've removed it from the constructors where it is not needed.  I suppose that such a set would be another possible implementation, though it may be a bit more cumbersome to ensure that both the AST and the translator have access to it.  What is the rationale?  Is it that we don't want to take up unnecessary space in the AST?  It seems to me that one flag is not too bad, but I'm open to changing it.

> 3. Parameter `isFuncDef` of `VCExpressionGenerator.Forall` (twice) and `VCExpressionGenerator.Exists` can be removed.

OK I think I've removed this where possible.

> Finally, could you please pull in the current state of the master branch? This will help us ensure that all regression tests pass.

Done.

cbarrettfb:
@shazqadeer 
> If the summary provided by @bkragl is correct, wouldn't it be better to have the field isFunctionDefinition in the axiom?

Hmm...unless I'm missing something, I don't think it's that simple.  We need to propagate the information that it's a definition from the AST through the translator to the VCExpr.  So we need a flag in the AST node, we need to check it in the translator, and we need a flag in the VCExpr - the quantifier info seemed like a convenient place to put the flag as we already store other flags there and it required changing the least amount of code.

cbarrettfb:
@RustanLeino 
> I'm thinking that it would be better to check for the attribute :define during the VC generation stages, rather than having the parser look for it and using another boolean in the AST. Was such a design rejected for some reason?

Hmm...how would that work exactly?  Maybe i'm missing something, but right now, it seems that functions are either inlined or converted into axioms, so there's no translation done on a function directly during the translation stage.  I'm not sure how I could check a quantified axiom to see whether it is a function definition without having some flag there.  There is no back pointer from the axiom to the function it was created from.  I'm definitely open to suggestions, but I didn't see a simpler way to do this.  I'm a noob when it comes to Boogie though so feel free to let me know if I'm missing something!

> Please document :define in the /attrHelp flag.

Done.

cbarrettfb:
> I suggest considering the design of this feature from the Boogie perspective. Is the following a good way to present it to the Boogie user?
> 
> * If a function is declared without a body, it is just an uninterpreted function. Of course, you can add axioms about it if you wish.
> * If the function is given a body, then Boogie gives you three options:
>   
>   * Macro: The definition can be inlined in each place the function occurs, like a macro. This is illegal to do for recursively defined functions. Because the function definition is inlined, the function symbol itself is not part of the VC; this means that any use of the function in a trigger will also be replaced by the definition (which may produce an illegal trigger, where the function form looked okay as a trigger).
>   * Axiom: The definition can be turned into a universally quantified axiom. By giving the definition in this way, it looks nice and localized in the Boogie program, and it saves you the trouble of generating the parameters once for the function and once for the axiom.
>   * Lazy Expansion: If the underlying prover supports it, then the responsibility of equating a use of a function with its definition can be done during proving. This is similar to Axiom, but allows the prover to use different heuristics for when to bring the function definition into play. Like Macro, this option is illegal for recursively defined functions. (Does the function symbol stay around? In particular, what happens with a Lazy Expanded function that's used in a trigger?)
> 
> What is the best way for a Boogie user to select between these options? Currently, we have the {:inline} attribute, the new {:define} attribute, and probably the former also looks at some command-line options. It would be nice to have a clean and understandable story here.

Yes, I think this is an accurate picture after adding this feature.  I might say it a slightly different way.  With the first two options (macro, axiom), Boogie takes responsibility for how the function will be encoded.  It's either inlined or turned into a quantified expression.  In the third case (using :define), the solver is responsible for it.  The solver can choose to inline or not and can choose to use quantifier-like heuristics or not.  The question about whether the function symbol stays around is probably solver-dependent.  I don't know what would happen if you use it in a trigger, but my guess is that it will probably behave similar to the first option (Macro).

As for the question of how a Boogie user should select between the options, that's a bit tricky.  I would have guessed that {:inline} and {:define} would behave similarly, but I found that on our examples, {:define} is often much more efficient.  But this may be different on other examples.

For me, it really comes down to having more fine-grained control over the generated SMT file.  I always prefer using quantifier-free encodings if possible, so for me the {:define} route is a nice clean way to do that, and it's what I would personally recommend, but again I suspect others may take a different view.

zvonimir:
This pull request is subsumed by #236 and so I am closing this one.

RustanLeino:
I'm trying to figure out if the "Axiom" mode is still supported. In the experiments I've tried, it seems that a function with a body (and without the `{:define}` attribute) always causes the function's body to be inlined (whether or not `{:inline}` is explicitly part of the input). Am I missing something or is the "Axiom" functionality not supported?

shazqadeer:
I just tried Boogie master version on the following program:

```
function foo(x: int): int {
    3
}

procedure A() {
    assert foo(0) == 3;
}
```

The generated SMTLIB2 file was the following:

```
(set-info :smt-lib-version 2.6)
(set-option :smt.mbqi false)
(set-option :model.compact false)
(set-option :model.v2 true)
(set-option :pp.bv_literals false)
; done setting options


(declare-fun tickleBool (Bool) Bool)
(assert (and (tickleBool true) (tickleBool false)))
(declare-fun foo (Int) Int)
(assert (forall ((x Int) ) (! (= (foo x) 3)
 :qid |blahbpl.1:14|
 :skolemid |0|
 :pattern ( (foo x))
)))
(declare-fun ControlFlow (Int Int) Int)
(push 1)
(set-info :boogie-vc-id A)
(set-option :timeout 0)
(set-option :rlimit 0)
(assert (not
 (=> (= (ControlFlow 0 0) 66) (let ((anon0_correct  (=> (= (ControlFlow 0 41) (- 0 67)) (= (foo 0) 3))))
(let ((PreconditionGeneratedEntry_correct  (=> (= (ControlFlow 0 66) 41) anon0_correct)))
PreconditionGeneratedEntry_correct)))
))
(check-sat)
(pop 1)
; Valid
```

As you can see, the axiom is being generated.

RustanLeino:
Oh, thank you! I now see my problem. I'm using the API and wanted to have the effect of the `foo` in your example. I had set the `.Body` field, but that's not the way to do it. (There's even a comment next to that field in the Boogie sources that tells me `.Body` is to be used only with `{:inline}`.) To me delight, I see there's a method `Function.CreateDefinitionAxiom` that I should call instead of setting the `.Body`. Thanks for setting me straight.

PR #197, opened 2020-02-16
Prover option cleanup

### LINKED ISSUE #73: Discussing (and cleaning up) default options passed to Z3

bkragl:
This PR cleans up the prover-related command-line options of Boogie, and reduces the default options sent to Z3 to a minimum.

Closes #73.

# Prover library

Renamed `-prover` to `-proverDll` to indicate that it is the name of a dynamically loaded DLL that implements the interface to some prover. Currently the only included one is `SMTLib`.

Introduced `-proverHelp` to print the options available with `-proverOpt` for the used prover DLL.

# Discovery of prover executable

The code to find a prover executable was unified into one place (instead of one for every supported SMT solver). The following prover-specific options were removed: `-z3exe`, `-z3name`, `-cvc4exe`, `-yices2exe`.

Instead, the already existing `-proverOpt:PROVER_PATH` and `-proverOpt:PROVER_NAME` have to be used.

If specified, `PROVER_PATH` has priority, otherwise we look for `PROVER_NAME` next to the Boogie executable, and finally for `PROVER_NAME` in the system path. If not specified, `PROVER_NAME` defaults to the name of the used SMT solver (given with `-proverOpt:SOLVER`), and we also try with the `.exe` file extension added.

# Legacy options

There is a generic way to specify options sent to the SMT solver (i.e., as `(set-option :<name> <value>)`), namely `-proverOpt:O` (e.g., `-proverOpt:O:smt.mbqi=false`). The Z3-specific `-z3opt` was removed.

The Z3-specific `-z3multipleErrors` was removed because it is already covered by `-proverOpt:MULTI_TRACES`.

The following outdated options were removed: `-z3lets`, `-z3types`, `-useSmtOutputFormat`, `-simplifyMatchDepth`, `-proverOpt:SMTLIB2_MODEL`.

# Z3 default options

The options sent to Z3 by default were reduced to the following:
```
smt.mbqi: false        // default: true
model.compact: false   // default: false
model.v2: true         // default: false
pp.bv_literals: false  // default: true
```

Consequently, `-proverOpt:OPTIMIZE_FOR_BV` got obsolete and was removed.

In the Boogie regression tests, only very few updates were necessary, related to a different generated error trace. This might be a breaking change for Boogie clients, but they have the capability to recover the previously set options (or any other set of options they like) using `-proverOpt:O:...` as explained above.

# Disabled tests

The expected output for the following tests is very fragile and thus they were set to not run as part of the regressions: `test2/Rlimitouts0.bpl`, `test7/MultipleErrors.bpl`.


zvonimir:
I like this update, thanks!
I do have a question though. I am wondering how setting of Z3 tactics (not sure if other solvers support them) would work with the current strategy for setting options?
We encountered the need to play with tactics a couple of times, and that has been something I wanted to add to Boogie.

akashlal:
As is, the changes look fine. It will need an investigation from Corral side, in order to get back the same default behavior as before, but seems doable.

bkragl:
> I am wondering how setting of Z3 tactics (not sure if other solvers support them) would work with the current strategy for setting options?

Hi @zvonimir. I have not used Z3 tactics before, but it looks like they are quite different (in the way they are sent to the solver) compared to regular options (i.e., sending `(set-option :<name> <value>)` at the beginning of the input). So I think we should not conflate them.

Here are three suggestions you could consider:
* If just a single string per Boogie invocation is fine, you could introduce `-proverOpt:TACTIC=...`. There are other Z3-specific options under `-proverOpt`, like `MULTI_TRACES`, which is fine. We just wanted to avoid "top-level" options that are prover specific. If more solvers can understand tactics, even better.
* If the user should be able to supply multiple tactic strings, you could introduce a new "namespace" under `-proverOpt` (like `-proverOpt:O:...` for options, and `-proverOpt:C:...` for command-line arguments). For example `-proverOpt:T:...`.
* If more fine-grained control on the level of procedures, or even code locations, is needed, maybe you could introduce a new attribute.

To me it seems like the bulk of the work of adding tactic support would be getting the low-level VC generation (i.e., sending the tactic commands to the solver) right. This PR should neither help nor hinder such an effort.

shazqadeer:
I have had extensive discussions with both @bkragl and @RustanLeino about this PR.  I am quite comfortable moving ahead.  Everything that was possible before should be possible now but will require more explicit command-line parameters.

bkragl:
We should make use of the new automatic versioning mechanism to indicate
the breaking change.

@michael-emmi, do you think bumping minor version number is appropriate, or
should we go for major version? It seems that semantic versioning dictates
major version, but I expect some further upcoming cleanup and would refrain
from bumping major version too often.

michael-emmi:
> We should make use of the new automatic versioning mechanism to indicate the breaking change. @michael-emmi, do you think bumping minor version number is appropriate, or should we go for major version? It seems that semantic versioning dictates major version, but I expect some further upcoming cleanup and would refrain from bumping major version too often.

Right. Strictly speaking this might be a major version bump, but I would agree to relax the rules in this case and bump minor version — for the reasons you say.

RustanLeino:
This merged PR does not build. I get:

```
  CSC : error CS2001: Source file '~/Source/boogie/Source/Provers/SMTLib/Yices2.cs' could not be found. [~/Source/boogie/Source/Provers/SMTLib/SMTLib.csproj]
  CSC : error CS2001: Source file '~/Source/boogie/Source/Provers/SMTLib/CVC4.cs' could not be found. [~/Source/boogie/Source/Provers/SMTLib/SMTLib.csproj]
```

bkragl:
> This merged PR does not build. I get:
> 
> ```
>   CSC : error CS2001: Source file '~/Source/boogie/Source/Provers/SMTLib/Yices2.cs' could not be found. [~/Source/boogie/Source/Provers/SMTLib/SMTLib.csproj]
>   CSC : error CS2001: Source file '~/Source/boogie/Source/Provers/SMTLib/CVC4.cs' could not be found. [~/Source/boogie/Source/Provers/SMTLib/SMTLib.csproj]
> ```

Sorry, I did not realize I have to delete the entries for the deleted `Yices2.cs` and `CVC4.cs` files from the old csproj file. I pushed a commit that should fix this.

PR #199, opened 2020-02-17
Remove stratified inlining

shazqadeer:
This PR removes the option /stratifiedInline from Boogie.  In a discussion with @akashlal , we concluded that Corral is the official implementation of stratified inlining so there is little point in maintaining another implementation in Boogie.

@akashlal: So far, I am being conservative in what code I delete. I tried to delete the minimum so I can compile the code while removing all the command-line options related to stratified inlining. I suspect more code can be deleted. Please advise. 

@shuvendu-lahiri : Some of the tests in extractLoops folder depend on stratified inlining.  I don't know how to capture their essence without using stratified inlining.  Currently, I have disabled them.  Either advise how to fix them or move them to the Corral repository.


akashlal:
@shazqadeer, this is likely deleting too much. Corral needs to set the `Clo` flag `StratifiedInlining` in many places to control the behavior of Boogie methods that it calls. I can see that this flag is very pervasive, but unfortunately most of it is needed.  I would recommend abandoning this PR. Instead, open an issue and assign it to me. I'll have to do it in a way that makes sure Corral is still running.

shazqadeer:
@akashlal : I will do what you suggest.  Let me explain what I was trying to do that brought me to this task of eliminating /stratifiedInline as a Boogie option.  The main thing I want to accomplish is to eliminate labels from Boogie entirely.  Labels is a very old feature, inherited from Simplify, that is  still used and causes duplication and complexity in the VC generation.  While doing this work, I discovered that the regressions in stratifiedInline folder do not work without labels.

It seems to me that there are two separate issues on which progress can be made independently: removing labels and simplifying the SI code in Boogie so it acts purely as a library for Corral.  I can make progress on the first by disabling the tests under extractLoops and stratifiedInline.  I will assign the latter to you.

shazqadeer:
#200

shuvendu-lahiri:
Thanks @shazqadeer  for notifying. I would love to retain the tests detLoopExtract*.bpl (that SymDiff relies upon) in the regression, whose testing indeed relies on stratifiedinlining option. I will port them to Corral if SI option is entirely moved into Corral.

Issue #200, opened 2020-02-17
Cleanup of stratified inline code


shazqadeer:
Currently, Boogie implements code used as a library for implementing stratified inlining of procedures.  But it also implements the algorithm itself, without all the bells and whistles of Corral.  It would be good to keep the library but eliminate the implementation of the algorithm to reduce maintenance and increase robustness.

shazqadeer:
@akashlal : Are you planning to look into this issue?

shazqadeer:
@akashlal :  We need to make progress on this issue.  I am happy to work on it myself if you don't have the cycles. For starters, could you tell me which flags in CommandLineOptions need to be retained for Corral's benefit?

akashlal:
Can I blame the delay on the pandemic? :) Give me this week to sort this out ...

akashlal:
I need some tips. I want to locally make some changes to Boogie, and then try out those changes with Corral. The trouble is that Corral now consumes Boogie as a nuget package. What's the best way to try out my local changes?

akashlal:
I think I'm done -- on this branch: akashl/remove-si
Still need to test Corral, but unable to pull in the Boogie dlls ...

RustanLeino:
> I need some tips. I want to locally make some changes to Boogie, and then try out those changes with Corral. The trouble is that Corral now consumes Boogie as a nuget package. What's the best way to try out my local changes?

I had a tough time doing this with Dafny. I was able to do it once, but wasn't able to reproduce it after that. If you figure out the right recipe for doing this, I'd love to know.

shazqadeer:
You should be able to create a Boogie package locally as follows:

dotnet build Source/Boogie.sln
dotnet pack Source/Boogie.sln

After that, I suppose you have to tell Corral to grab the generated package rather than pulling from nuget.  That would require some investigation.

shazqadeer:
@akashlal : Could you please put up a PR so I can look at your changes?

RustanLeino:
> After that, I suppose you have to tell Corral to grab the generated package rather than pulling from nuget. That would require some investigation.

Yeah, this is where I had got stuck.

shazqadeer:
I found a pertinent discussion on stackoverflow: https://stackoverflow.com/questions/43400069/add-a-package-with-a-local-package-file-in-dotnet.

Also, @bkragl may have some insights.

akashlal:
[...]

PR #201, opened 2020-02-18
Remove labels and unused VC generation code

shazqadeer:
This PR removes labels inside the verification conditions generated by Boogie.  Labels were used to generate error traces but are obsolete now given that SMT solvers are reasonably good at generating models.  The new mechanism uses extra "control-flow" instrumentation in the verification to generate error traces.  This PR also removes a bunch of algorithms for generating verification conditions that were not being used. As a result, the option /vc has been eliminated.

The tests inside stratifiedInline and extractLoops folder have been disabled because they are dependent on implementation of stratified inlining which is going to be eliminated from Boogie (#200) and which works only with labels.

The option MULTI_TRACES worked only with labels.  It is now obsolete and has been eliminated.  If there is demand for multiple error traces, we can investigate techniques based purely on model finding to solve the problem.


akashlal:
Changes look good. Goodbye labels.

shazqadeer:
@RustanLeino : I double checked that with the current version of this branch, all tests in Dafny pass (with Z3 4.8.4) except those in comp folder and a couple others that require some .net dlls to be installed (I am working on a MacOS machine).  So I think we are good to land this PR.

shazqadeer:
I have removed /vc option entirely.  So its mention in the help message is also eliminated.

I have removed the doomed code and tests.

I tried to remove `doomedInjectedAssumptionVariables` in a couple of different ways (without really understanding the point of this variable) but the test Test/snapshots/runtest.snapshot was failing.  So I gave up on this endeavor. Perhaps @RustanLeino  can explain the purpose of this field and whether it can be eliminated.  If it cannot be eliminated, perhaps it should be renamed.

wuestholz:
@shazqadeer @bkragl `doomedInjectedAssumptionVariables` definitely doesn't have anything to do with the doomed VC generation option. I wrote that code a long time ago for the verification result caching feature that is used by the Dafny IDE, but I don't fully remember the reason for picking that name... Probably that any attempt at reusing cached results is doomed. :)

RustanLeino:
I'm joining the party only after it already ended.

@shazqadeer Yes, this looks good. Thanks for doing this spring cleaning.

I'm surprised about the deletion of stratified inlining. Doesn't Corral use it?

@bkragl A _doomed program point_ is one from which execution will inevitably fail, regardless of how control reached the doomed point. It was the subject of @martinschaef's PhD thesis.

PR #202, opened 2020-02-20
Add option /noProc:<p> to exclude procedures matching pattern <p> from checking

bkragl:
`/noProc` is a counterpart to the already existing option `/proc`. The idea is that we first consider the set of *included* procedures (either all procedures if no `/proc` argument is given, or otherwise all procedures matching any of the given `/proc` arguments), and then remove from it all procedures matching any of the given `/noProc` arguments.

### DIFF
--- a/Source/Core/CommandLineOptions.cs
+++ b/Source/Core/CommandLineOptions.cs
@@ -764,6 +764,9 @@ namespace Microsoft.Boogie {
       }
     }
 
+    // Note that procsToCheck stores all patterns <p> supplied with /proc:<p>
+    // (and similarly procsToIgnore for /noProc:<p>). Thus, if procsToCheck
+    // is empty it means that all procedures should be checked.
     private List<string/*!*/> procsToCheck = new List<string/*!*/>();
     private List<string/*!*/> procsToIgnore = new List<string/*!*/>();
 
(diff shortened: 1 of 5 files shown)


Source/Core/CommandLineOptions.cs
```diff
@@ -764,18 +764,13 @@ public enum TypeEncoding {
...
     }
 
-    public IEnumerable<string/*!*/> ProcsToCheck {
-      get {
-        Contract.Ensures(cce.NonNullElements(Contract.Result<IEnumerable<string/*!*/>>(), true));
-        return this.procsToCheck != null ? this.procsToCheck.AsEnumerable() : null;
-      }
-    }
-
-    private List<string/*!*/> procsToCheck = null;  // null means "no restriction"
```
shazqadeer: Do not delete the comment.

Also, I have a slight preference for using null to represent "no restriction".  But it is not a big deal.

bkragl: I found this better (and more uniform with `procsToIgnore`) because it safes on non-null checks in several places.

It just represents what was given on the command line, a possibly empty list of `/proc`.

bkragl: Added back a comment.

PR #206, opened 2020-02-21
Fix lit for mac

cbarrettfb:
When running the test framework "lit" on a mac, it needs to be run as an argument of "dotnet" and the name has a ".dll" suffix.  This fixes the configuration to automatically do the right thing when running
lit -D dotnet

### DIFF
--- a/Test/lit.site.cfg
+++ b/Test/lit.site.cfg
@@ -84,6 +84,9 @@ runtime = ''
 
 if 'dotnet' in lit_config.params:
     boogieBinary = 'Source/BoogieDriver/bin/{}/{}/BoogieDriver'.format(configuration, framework)
+    if platform.system() == 'Darwin':
+        runtime = 'dotnet'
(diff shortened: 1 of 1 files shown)


shazqadeer:
I didn't understand the issue this PR is fixing.  I have been running "lit . -D dotnet" on my Mac machine and it seems to do the right thing.  

@bkragl : Please accept if all looks good to you.

bkragl:
@cbarrettfb, are you running dotnet version 3.1? I think I remember some older version not generating the executable without .dll file extension.

If we merge this, I'd suggest eliminating the case distinction for Mac OS. Your change works the same way on Linux.

cbarrettfb:
Hmm..yes I seem to be running 3.1.  That seems to be the default when I go to Microsoft's website to download .NET for Mac.  Should I be using a different version?  I'm happy to upgrade.  Or can you suggest a more robust fix to the config file?

PR #211, opened 2020-03-02
Add support to inject lemmas into commutativity checks

bkragl:
This PR adds a mechanism to inject lemmas into CIVL's commutativity checks. To unify notation, the syntax for commutativity witnesses also changed.

Let `x` and `y` be a shared variables, and both `A` and `B` be atomic actions with input parameter `i`.

**Lemmas**

```
function {:lemma}{:commutativity "A", "B"}
lemma (x: int, x':int, first_i, second_i) : bool
{
  boolean_expression
}
```

With the above declaration, CIVL injects the following assumption into the commutativity check between `A` and `B`:
```
assume lemma(old(x), x, first_i, second_i);
```
This means that we have the following parameter mapping:
* `x`: pre-value of `x`
* `x'`: post-value of `x`
* `first_i`: parameter `i` of `A`
* `second_i`: parameter `i` of `B`

**Witnesses**

```
function {:witness "y"}{:commutativity "A", "B"}
witness (x: int, x':int, first_i, second_i) : type_of_y
{
  witness_expression
}
```

The commutativity check between `A` and `B` involves the computation of the transition relation `B ∘ A` (i.e., sequential composition of `B` and `A`). This transition-relation formula could require the introduction of an existential quantifier due to shared variable `y`. For example, if both `B` and `A` nondeterministically update `y`. The quantified variable represents the value of `y` after `B` but before `A`. To avoid this existential quantification, the programmer can supply a witness function like the one above. Then instead of `exists quantified_y :: ...` the expression `witness(old(x), x, first_i, second_i)` is used in the transition-relation formula.

Previously the annotation was `{:witness "y", "A", "B"}`.

A function can have at most one `lemma` or `witness` attribute, but arbitrarily many `commutativity` attributes.


shazqadeer:
Could you update the PR description to include the meaning of :witness attribute?

shazqadeer:
We should consider adding documentation of all the options and attributes supported by CIVL somewhere.  Perhaps Suha would be interested in picking up this work item.

shazqadeer:
@suhaorhun, @menesro: Just testing

bkragl:
> Could you update the PR description to include the meaning of :witness attribute?

I'll add some description. Did not put it initially, because it's not a new feature.

> We should consider adding documentation of all the options and attributes supported by CIVL somewhere. Perhaps @suhaorhun would be interested in picking up this work item.

`-help` documents some CIVL options. We should polish it and similarly add a CIVL section to `-attrHelp`.

shazqadeer:
@bkragl : We also discussed lemma procedures which are currently a sub case of introduction procedures in an artificially constrained way.  Are you planning a separate PR for cleaning up and generalizing lemma procedures?

bkragl:
> @bkragl : We also discussed lemma procedures which are currently a sub case of introduction procedures in an artificially constrained way. Are you planning a separate PR for cleaning up and generalizing lemma procedures?

Yes, this will be a separate PR. I'm not planning to do this immediately, though.

suhaorhun:
> We should consider adding documentation of all the options and attributes supported by CIVL somewhere. Perhaps @suhaorhun would be interested in picking up this work item.

It is interesting, I can try. However, I am inexperienced. I might need time, guidance and some clarifications. I checked the repo. I think I should modify source/core/CommandLineOptions.cs file and AttributeUsage method. Is it correct?

Can I directly push to this branch? Don't I need some permissions?

Thanks,

Issue #204, opened 2020-02-21
Make /noinfer by the default option


zvonimir:
I am not sure if there are users that invoke Boogie without using `/noinfer`. I've never managed to successfully leverage any of the abstract interpretation algorithms that are build into Boogie. So I think we should make `/noinfer` be the default behavior. If that is already the case, it might be good to document that.

Thoughts?

bkragl:
In my work on CIVL, I found that sometimes abstract interpretation figures out simple loop invariants when iterating over arrays, which are needed to prove the core invariant.

```
i := 0;
while(i < n)
invariant 0 <= i && i <= n; // For example this one.
{
  ...
  i := i + 1;
}
```

But we actually always strive to have our examples verified with `/noinfer`.

zvonimir:
Got it. To me this also says that `/noinfer` should probably be the default setting.

shazqadeer:
I am supportive of /noinfer being the default, especially because sometimes inference takes a long time when a significant amount of inlining has been done.  

@zvonimir : Please put up a PR.  

@RustanLeino : Heads up.

bkragl:
So the idea is that option `/noinfer` will be removed, and no abstract interpretation is performed unless requested with `/infer`, right?

zvonimir:
Yep, that is my take on this. I got busy with Corral and so I won't be able to get to this today.

zvonimir:
@rospoly This is a good first issue to try to resolve as you are learning Boogie.

alexanderjsummers:
I'd also like this to be the default behaviour - we specify it from all of our tools (and typically teach students to do it when they're new to using Boogie).

RustanLeino:
I'm late to the party. Dafny always uses -infer:j. But Dafny can turn this mode on explicitly.

Issue #31, opened 2016-02-11
polyhedra for linear inequalities not supported for abstract interpretation?


shaobo-he:
Hi,

I tried to invoke boogie with flag `/infer:p` while got an error message saying argument `'p'` is not valid. I was wondering if it is still supported while the help message is not updated.

Thanks,
Shaobo

shazqadeer:
@RustanLeino : This should be a quick one to close for you.

shazqadeer:
I just looked at the help message:

i = intervals
                   c = constant propagation
                   d = dynamic type
                   n = nullness
                   p = polyhedra for linear inequalities

I see that the above options are not supported by the parsing of /infer flag.  

@RustanLeino :  Could it be that support for the above options was dropped at some point but help was not updated?

bkragl:
It looks like there are two implementations of abstract interpretation in the Boogie codebase.

Project `AbsInt` implements the currently supported domains:
t = trivial bottom/top lattice
j = intervals (used to be called "strong intervals" to distinguish it from i)

Project `AIFramework` is currently not used anywhere, but seems to contain code for the other domains mentioned above.

PR #219 updates the help message to reflect the currently supported flags.

shazqadeer:
AIFramework is not part of the Boogie.sln.  Anybody know why?

PR #215, opened 2020-04-03
/noinfer is default behaviour

rospoly:
targeting #204

### DIFF
--- a/Source/Core/CommandLineOptions.cs
+++ b/Source/Core/CommandLineOptions.cs
@@ -573,7 +573,7 @@ namespace Microsoft.Boogie {
     }
     public bool ExpandLambdas = true; // not useful from command line, only to be set to false programatically
     public bool DoModSetAnalysis = false;
-    public bool UseAbstractInterpretation = true;          // true iff the user want to use abstract interpretation
+    public bool UseAbstractInterpretation = false;          // true iff the user want to use abstract interpretation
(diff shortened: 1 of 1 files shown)


Source/Core/CommandLineOptions.cs
```diff
@@ -1822,7 +1822,7 @@ checking errors
                    s = debug statistics
                 0..9 = number of iterations before applying a widen (default=0)
   /noinfer      turn off the default inference, and overrides the /infer
-                switch on its left
+                switch on its left (default)
```
zvonimir, line 1825: Wait, why do we need `/noinfer` if the default behavior now is to not infer?
It seems to me that we do not need this switch any more.

zvonimir, line 1825: Also, above it says `The default is /infer:i`. That is not true any more, and it should be removed.
You can say something like `If this flag is omitted, Boogie does not attempt to infer invariants.`

zvonimir:
It seems that there are regressions now failing because I would guess they relied on `/infer:i` being the default behavior. This should be fixed by adding `/infer:i` when invoking Boogie on those regressions.

shazqadeer:
Issue #31 is related to this PR.  Please take a look.

bkragl:
Hey @zvonimir & @rospoly, I already implemented this at some point locally and fixed the regressions. I'll create another PR and close this one. Hope that's fine for you.

zvonimir:
That's fine. Let me close this one then.

PR #220, opened 2020-04-14
Remove (some) obsolete options

bkragl:
@zvonimir pointed out that PR #197 made `WeakArrayTheory` obsolete. I think this is the only such prover-related option.

However, on a quick look I saw the following other unused option fields and dropped them. There might be more.
```
LoopFrameConditions
ModifiesDefault
LocalModifiesChecks
NoVerifyByDefault
OwnershipModelEncoding
CausalImplies
```

Also, I removed the obsolete methods `AddProverOption` and `RemoveAllProverOptions`, and the non-existing options `/log` and `/lazyInline` from the help message.

I seems that neither Dafny nor Corral reference anything removed by this PR, but it woulds still be good if @RustanLeino and @zvonimir could confirm that.

### DIFF
--- a/Source/Core/CommandLineOptions.cs
+++ b/Source/Core/CommandLineOptions.cs
@@ -533,8 +533,6 @@ namespace Microsoft.Boogie {
     public bool ShowVerifiedProcedureCount = true;
     [ContractInvariantMethod]
     void ObjectInvariant3() {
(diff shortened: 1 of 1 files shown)


zvonimir:
I suggest we merge this in. We can always (partially) revert it if things get broken down the line.

PR #227, opened 2020-05-05
Fix to recognize :timeLimit attribute on implementations

### LINKED ISSUE #63: timeLimit attribute on procedures appears to no longer work
Boogie allows a per-procedure timeLimit attribute:

```
procedure{:timeLimit 10} p1() {}
procedure{:timeLimit 20} p2() {}
```

This used to produce a different TIMEOUT option on each of the two SMT queries:
```
(set-info :smt-lib-version 2.0)
...
(set-option :TIMEOUT 10000)
...
(check-sat)
...
(set-option :TIMEOUT 20000)
(reset)
(set-option :TIMEOUT 20000)
(set-info :smt-lib-version 2.0)
...
(set-option :TIMEOUT 20000)
...
(check-sat)
...
```

I think this may have unintentionally been changed in commit 09a9f17b9a8642f779ea0e5097c0aab8ce9f370b.  After that commit, the TIMEOUT is 10000 in both queries.

I'm not sure what the best fix is; as the commit rightly points out, the old code was messy (e.g. TIMEOUT 20000 is printed three times in the SMT code shown above) and the timeLimit attribute handling was Z3-specific.  But it would be nice if there were a way to make the timeLimit attribute work again.

shazqadeer:
Fixes #63 .

### DIFF
--- a/Source/Provers/SMTLib/ProverInterface.cs
+++ b/Source/Provers/SMTLib/ProverInterface.cs
@@ -246,7 +246,7 @@ namespace Microsoft.Boogie.SMTLib
 
         // Set produce-unsat-cores last. It seems there's a bug in Z3 where if we set it earlier its value
         // gets reset by other set-option commands ( https://z3.codeplex.com/workitem/188 )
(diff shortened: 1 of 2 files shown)


RustanLeino:
It looks alright to me.

PR #231, opened 2020-05-17
Update z3 to the latest release v4.8.8

zvonimir:
Everything is working fine (with minor adjustments), expect that the regression `Test\civl\inductive-sequentialization\paxos\is.sh` is timing out.
@bkragl and @shazqadeer : how should we deal with this?


zvonimir:
The civl regression problem got fixed by the recent updates to CIVL. So we are ready to merge this.
@shazqadeer could you please quickly review this? Thx!

shazqadeer:
@zvonimir : I want to note that the Inductive Sequentialization proof of Paxos suffers from the butterfly effect.  The latest changes that seem to have "fixed" the problem are simply an instance of the butterfly effect.  In short, the problem can pop up again.  @bkragl and I have spent a few days on it to figure out how to make the proof more robust.  However, we are not done yet.  Having said that, I will go ahead and approve this PR.

zvonimir:
Yep, I understand, "fixed" was a wrong term to use here. Anyhow, it will take some time to figure out all the butterfly effect issues, and so, like you wrote, I'll merge this one.

PR #234, opened 2020-05-20
Refactor monomorphic

zvonimir:
This pull request:
- Disables AxiomBuilder for monomorphic encoding
- Removes Monomorphize and MonomorphicArrays flags. This means that array theory cannot be enabled on polymorphic programs, but that part did not work anyways and nobody was relying on it (to the best of my knowledge)
- Removed /typeEncoding:m from command line options. Boogie was already inferring whether a program is monomorphic automatically, and so we do not need a command like option for that.
- Boogie now reports an error if /useArrayTheory is used with a polymorphic program.

### DIFF
--- a/Source/Provers/SMTLib/ProverInterface.cs
+++ b/Source/Provers/SMTLib/ProverInterface.cs
@@ -36,7 +36,6 @@ namespace Microsoft.Boogie.SMTLib
     void ObjectInvariant()
     {
       Contract.Invariant(ctx != null);
-      Contract.Invariant(AxBuilder != null);
       Contract.Invariant(Namer != null);
       Contract.Invariant(DeclCollector != null);
       Contract.Invariant(cce.NonNullElements(Axioms));
@@ -2770,7 +2769,7 @@ namespace Microsoft.Boogie.SMTLib
 
     public override string Lookup(VCExprVar var)
     {
-      VCExprVar v = parent.AxBuilder.TryTyped2Untyped(var);
+      VCExprVar v = parent.AxBuilder?.TryTyped2Untyped(var);
       if (v != null) {
         var = v;
       }
--- a/Source/Provers/SMTLib/ProverInterface.cs
+++ b/Source/Provers/SMTLib/ProverInterface.cs
@@ -111,7 +111,7 @@ namespace Microsoft.Boogie.SMTLib
           AxBuilder.Setup();
           break;
         case CommandLineOptions.TypeEncoding.Monomorphic:
-          AxBuilder = new TypeAxiomBuilderPremisses(gen);
+          AxBuilder = null;
           break;
         default:
           AxBuilder = new TypeAxiomBuilderPremisses(gen);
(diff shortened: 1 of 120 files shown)


Source/Provers/SMTLib/ProverInterface.cs
```diff
@@ -111,7 +110,7 @@ private void SetupAxiomBuilder(VCExpressionGenerator gen)
           AxBuilder.Setup();
           break;
         case CommandLineOptions.TypeEncoding.Monomorphic:
-          AxBuilder = new TypeAxiomBuilderPremisses(gen);
+          AxBuilder = null;
```
shazqadeer, line 113: Can you provide some context for the effect of this change?  For example, why is it safe to make this change?

Source/VCExpr/TypeErasure.cs
```diff
@@ -685,7 +685,7 @@ private struct TypeCastSet {
     [Pure]
     public override bool UnchangedType(Type type) {
       //Contract.Requires(type != null);
-      return type.IsInt || type.IsReal || type.IsBool || type.IsBv || type.IsFloat || type.IsRMode || type.IsString || type.IsRegEx || (type.IsMap && CommandLineOptions.Clo.MonomorphicArrays);
+      return type.IsInt || type.IsReal || type.IsBool || type.IsBv || type.IsFloat || type.IsRMode || type.IsString || type.IsRegEx;
```
shazqadeer, line 688: I have a similar question as above for the changes in this file.

Source/VCExpr/TypeErasureArguments.cs
```diff
@@ -193,41 +193,23 @@ internal MapTypeAbstractionBuilderArguments(TypeAxiomBuilderArguments axBuilder,
...
       }
       // Fill in the map type
-      if (CommandLineOptions.Clo.MonomorphicArrays) {
-        selectTypes[i] = abstractedType;
-        storeTypes[i] = abstractedType;
-      } else {
-        selectTypes[i] = AxBuilder.U;
-        storeTypes[i] = AxBuilder.U;
-      }
+      selectTypes[i] = AxBuilder.U;
```
shazqadeer, line 196: I suppose this change is safe because we are not even constructing AxiomBuilder for monomorphic encoding?

Source/Provers/SMTLib/ProverInterface.cs
zvonimir: TypeAxiomBuilders are used to generate type axioms for polymorphic programs. Now, there used to be code in Boogie, which I removed in this pull request, that, to the best of my understanding, attempted to support polymorphic programs but have monomorphic arrays. This was controlled using Monomorphize and MonomorphicArrays flags.

Once we decided not to support that functionality, and note that when I tried it out it did not work in the first place, there was no need to be creating an axiom builder here for monomorphic programs.

Source/VCExpr/TypeErasure.cs
zvonimir: There is an invariant now that no AxiomBuilders get created/invoked for monomorphic programs.

Source/VCExpr/TypeErasureArguments.cs
zvonimir: Yes. There is an invariant now that no AxiomBuilders get created/invoked for monomorphic programs.

PR #235, opened 2020-05-21
[CIVL] Improve inductive sequentialization proof for Paxos

shazqadeer:
The Inductive Sequentialization proof for Paxos is flaky.  This PR attempts to make it more robust by adding patterns on a quantifier and removing inline attribute on a function being used in a trigger.


zvonimir:
It would be good to try it out with Z3 4.8.8.

bkragl:
The work in this PR is not quite finished. I'll cherry pick and merge two changes (splitting of IS inductiveness checks and dropping `:inline` from a function that is used in a trigger). Other improvements can be revisited later.

PR #236, opened 2020-05-28
Refactored the define attribute on functions

zvonimir:
This pull request continues where #196 left off. In particular, instead of pushing function definitions through Boogie by piggybacking on axioms, I refactored this part to instead have an explicit list of function definitions.


zvonimir:
This is still an ongoing work. There is an issue when multiple VCs are being generated.
Moreover, I will clean up the code further.

zvonimir:
Ok, I think this is ready for a round of reviewing.

Source/Core/Absy.cs
```diff
@@ -2871,6 +2892,33 @@ public Function(IToken tok, string name, List<TypeVariable> typeParams, List<Var
...
     }
+
+    // Generates function definition of the form func_call == func_body
+    // For example, for
+    // function {:define} foo(x:int) returns(int) { x + 1 }
+    // this will generate
+    // foo(x):int == x + 1
+    // We need the left hand call part later on to be able to generate
+    // the appropriate SMTlib style function definition.
+    public NAryExpr CreateFunctionDefinition(Expr body) {
```
shazqadeer, line 2903: I find it odd to generate a DefBody in the SMTLIB style Absy.cs.  I was expecting that you would keep only the body expression here and then generate the function call etc on the left-hand side as needed when you are creating the actual SMT input.

zvonimir: I agree. I invested quite a bit of time trying to make this cleaner. The tricky part is resolution, since you have to create the "function call" from function parameters that did not go through the resolution process, and that still have to be connected to the appropriate variables in the function body. At the point of VC generation, where function definition gets created, I could not figure out how to do this reliably.

Source/Core/Absy.cs
```diff
@@ -2671,8 +2671,10 @@ public DatatypeMembership(Function constructor)
   public class Function : DeclWithFormals {
     public string Comment;
 
-    // the body is only set if the function is declared with {:inline}
+    // Body is only set if the function is declared with {:inline}
     public Expr Body;
+    // DefBody is only set if the function is declared with {:define}
+    public NAryExpr DefBody;
```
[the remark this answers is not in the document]
zvonimir: I resolved both of these. We are not allowing now both define and inline to be set. The precedence idea seems too complicated to me (in terms of user experience) and probably not really needed.

Source/Core/Absy.cs
shazqadeer: ok, understood.  Let us go ahead and land this PR.  We will investigate and improve gradually.

zvonimir:
@RustanLeino Any chance you could also review it before it gets merged? Thanks!

Source/Core/Absy.cs
```diff
@@ -2755,25 +2755,41 @@ public Function(IToken tok, string name, List<TypeVariable> typeParams, List<Var
...
         // that the body will be inlined.
         stream.Write("{:inline} ");
       }
+      if (DefinitionBody != null && !QKeyValue.FindBoolAttribute(Attributes, "define")) {
+        // Boogie defines any function whose .DefinitionBody field is non-null.  The parser populates the .DefinitionBody field
+        // if the :define attribute is present, but if someone creates the Boogie file directly as an AST, then
+        // the :define attribute may not be there.  We'll make sure it's printed, so one can see that this means
+        // that the function will be defined.
+        stream.Write("{:define} ");
+      }
```
RustanLeino, line 2771: Nice

Source/Core/BoogiePL.atg
```diff
@@ -520,7 +520,13 @@ Function<.out List<Declaration>/*!*/ ds.>
     if (definition != null) {
       // generate either an axiom or a function body
       if (QKeyValue.FindBoolAttribute(kv, "inline")) {
+        if (QKeyValue.FindBoolAttribute(kv, "define"))
+          SemErr("function cannot have both inline and define attributes");
         func.Body = definition;
+      } else if (QKeyValue.FindBoolAttribute(kv, "define")) {
+        func.DefinitionBody = func.CreateFunctionDefinition(definition);
+        if (func.TypeParameters.Count > 0)
+          SemErr("function with define attribute has to be monomorphic");
```
RustanLeino, line 529: I suggest also using the colon (but not curly braces) to these messages, as in:

    both :inline and :define attributes

It makes it clearer that these are names of attributes, as opposed to general words in the sentence.

Both lines 524 and 529.

Source/Core/CommandLineOptions.cs
```diff
@@ -1621,9 +1621,17 @@ public class ConcurrentHoudiniOptions
      {:bvbuiltin ""spec""}
        Rewrite the function to built-in prover function symbol 'fn'.
 
+     {:define}
+     {:define true}
+       Turn this function into a definition (using the define-fun construct)
+       when using the SMT-LIB backend.  Can only be used with non-recursive
+       functions. Cannot be combined with :inline attribute.
```
RustanLeino, line 1628: The text here is fine, but I have a question for my own understanding: Whenever these say "Can only ..." and "Cannot ...", those things are checked and enforced, right? (I saw one is in the code, but I didn't search for the other.) We should always check any such usage rules--this pays off over time, because it prevents strange bugs and backward-compatibility issues with unintended usages.

zvonimir: Good question. Yes, they are both checked and enforced. Which reminded me to add a regression for recursion check. I'll do that.

Source/Core/BoogiePL.atg
zvonimir: Good point. Done.

Source/Core/BoogiePL.atg
```diff
@@ -520,9 +520,13 @@ Function<.out List<Declaration>/*!*/ ds.>
     if (definition != null) {
       // generate either an axiom or a function body
       if (QKeyValue.FindBoolAttribute(kv, "inline")) {
+        if (QKeyValue.FindBoolAttribute(kv, "define"))
+          SemErr("function cannot have both :inline and :define attributes");
         func.Body = definition;
       } else if (QKeyValue.FindBoolAttribute(kv, "define")) {
-        ds.Add(func.CreateFunctionDefinition(definition, kv));
+        func.DefinitionBody = func.CreateFunctionDefinition(definition);
+        if (func.TypeParameters.Count > 0)
```
cbarrettfb, line 528: Error check before creating the function definition?

Source/Core/Absy.cs
```diff
@@ -2671,8 +2671,8 @@ public DatatypeMembership(Function constructor)
   public class Function : DeclWithFormals {
     public string Comment;
 
-    // the body is only set if the function is declared with {:inline}
-    public Expr Body;
+    public Expr Body; // Only set if the function is declared with {:inline}
+    public NAryExpr DefinitionBody; // Only set if the function is declared with {:define}
```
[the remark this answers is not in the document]
zvonimir: Yes, you are absolutely right, and this is a valid point. I made sure I checked off all such occurrences and I don't think I missed any.
The original plan I discussed with Shaz was not to duplicate code, but to use `Body` for both. However, `Body` is being leveraged in rarely used codes (AbstractHoudini and FixedPointVC) that we do not quite understand. Hence, I was hesitant to reuse it until someone either cleans up AbstractHoudini and FixedPointVC or completely removes them. My time was running out and I did not get to that.

Source/Core/BoogiePL.atg
zvonimir: Good point. Done.

Issue #195, opened 2020-02-14
Eliminating polymorphism in Boogie


shazqadeer:
The implementation of polymorphism in Boogie is complex because it introduces quantifiers and encodes all types into a single type.  This complexity discourages Boogie users from exploiting this potentially useful feature.  At the same time, the existing implementation is inadequate for fully encoding source-level polymorphism of a language such as Dafny.  It is time to revisit the feature, potentially eliminate the current implementation, and start afresh.  I am creating this issue so any users of could voice their concerns.

@RustanLeino : Thoughts?  If you know of other users of Boogie polymorphism, please inform them that we are considering dropping support for this feature.

mueller55:
Dear Shaz, I am not sure I fully understand your post. Are you suggesting to find a different encoding of polymorphic types or are you proposing to drop polymorphism from the Boogie type system altogether (possibly going back to an "any" type like in the initial version of Boogie)?

shazqadeer:
I was being vague in the initial post.  Thanks for the response and the suggestion to sharpen the discussion.

I would like to remove the "polymorphic map" feature entirely.  This feature is much more complex than the rest of polymorphic Boogie, which is based on parametric polymorphism found in other languages.  The polymorphic map feature was added to address a specific use case: encoding the heap of an object-oriented language.  However, this feature was not enough to address the encoding needs of Dafny and more encoding had to be done on top of it.  In summary, this feature is complex, narrow, and inadequate.

I am unsure about what to do about the other polymorphic features in Boogie.  The current situation is untenable because even the translation of a Boogie program that does not use type parameters introduces quantifiers in the verification condition.  To avoid quantifiers for a non-generic Boogie program, the user has to use the command-line flag /typeEncoding:m, a poorly-understood hack implemented by me many years ago.  There is also another related mysterious flag /monomorphize whose purpose is unclear to me.  

Overall, the Boogie community should try to answer the following questions:
- Does polymorphic Boogie aid the development of verifiers for polymorphic source languages?  Based on the experience so far, the answer does not appear to be Yes.  @RustanLeino informs me that he had to do a ton of work in the Dafny verifier to encode its polymorphic features in spite of the availability of polymorphic features in Boogie.  He observed that even if all polymorphic features in Boogie were dropped, he could address the gap in the Dafny implementation with a modest amount of work.  This is rather surprising and might indicate that source-level polymorphism interacts with other language features which makes it difficult to directly use Boogie-level polymorphic features.

[...]

PR #241, opened 2020-06-08
Add Viper tests to Test suite to showcase use of polymorphic maps

gauravpartha:
In issue #195, the suggestion has come up to simplify Boogie's polymorphic type system. In particular, one idea is to remove polymorphic maps. However, Viper (https://github.com/viperproject/silver/) makes great use of polymorphic maps when encoding Viper programs into Boogie.
This has been mentioned in issue #195 and one suggestion was to add few small tests to the Boogie test suite that make use of these polymorphic features. The goal is that if anything were to change, then these tests would also be taken into account (in the worst case, we would at least be notified that some feature has been removed). Ideally, we (the Viper team) hope that polymorphic maps won't be removed from Boogie. 

Both of the added examples make use of three polymorphic map types (all Viper encodings require the first two types, the third one is used whenever predicates are used, which happens frequently):

Encoding of the heap:
type HeapType = <A, B> [Ref, Field A B]B;

Encoding of the permission mask (tracking separation logic style permissions):
type MaskType = <A, B> [Ref, Field A B]Perm;

Encoding of known-folded permissions (discussed in http://pm.inf.ethz.ch/publications/getpdf.php?bibname=Own&id=HeuleKassiosMuellerSummers13.pdf):
type PMaskType = <A, B> [Ref, Field A B]bool;

### DIFF
--- a/Test/viper/predicate_exhale.bpl
+++ b/Test/viper/predicate_exhale.bpl
@@ -1,0 +1,767 @@
+// RUN: %boogie "%s" | %OutputCheck "%s"
+
+// ==================================================
+// Preamble of State module.
+// ==================================================
+
+function state(Heap: HeapType, Mask: MaskType): bool;
+
+// ==================================================
+// Preamble of Heap module.
+// ==================================================
+
+type Ref;
+var Heap: HeapType;
+const null: Ref;
+type Field A B;
+type NormalField;
+type HeapType = <A, B> [Ref, Field A B]B;
+const unique $allocated: Field NormalField bool;
+axiom (forall o: Ref, f: (Field NormalField Ref), Heap: HeapType ::
+  { Heap[o, f] }
+  Heap[o, $allocated] ==> Heap[Heap[o, f], $allocated]
+);
+function succHeap(Heap0: HeapType, Heap1: HeapType): bool;
+function succHeapTrans(Heap0: HeapType, Heap1: HeapType): bool;
+function IdenticalOnKnownLocations(Heap: HeapType, ExhaleHeap: HeapType, Mask: MaskType): bool;
+function IsPredicateField<A, B>(f_1: (Field A B)): bool;
+function IsWandField<A, B>(f_1: (Field A B)): bool;
+function getPredicateId<A, B>(f_1: (Field A B)): int;
+// Frame all locations with direct permissions
+axiom (forall <A, B> Heap: HeapType, ExhaleHeap: HeapType, Mask: MaskType, o: Ref, f_2: (Field A B) ::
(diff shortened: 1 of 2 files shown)


Test/viper/predicate_exhale.bpl
```diff
@@ -0,0 +1,767 @@
+// RUN: %boogie "%s" | %OutputCheck "%s"
```
zvonimir, line 1: You used a slightly different way to check Boogie output in the file above. Any particular reasons for using two different ways of checking?

gauravpartha: Yes, one assertion in predicate_exhale.bpl is expected to fail. I added another "CHECK-L" for that assertion. It can be seen here: https://github.com/boogie-org/boogie/blob/101b776424e22b81b1d3e7e304b34e520b7316ac/Test/viper/predicate_exhale.bpl#L763 

In known_folded.bpl no assertion is supposed to fail. This distinction is the reason for using two different ways of checking.

Thanks a lot for looking into the PR!

zvonimir: Sorry, but I am still confused :).
I don't understand the difference between:
```
// RUN: %boogie -proc:known_folded_1 -proc:known_folded_2 "%s" > "%t"
// RUN: %OutputCheck --file-to-check "%t" "%s"
```
and
```
// RUN: %boogie "%s" | %OutputCheck "%s"
```

Don't they accomplish the same thing? If not, they how are they different?
I am just trying to learn myself. Thanks!

gauravpartha: Oh, I misunderstood your question, sorry.

I must admit that I did not even think about the difference. I just looked at one example each from the repository where no failure is expected and one where a failure is expected. In those two cases they used different ways to run the test. I agree that one should rather do this in a consistent way here.

I think the commands do the same thing in this case, but I am not sure in general. Either way:
```
// RUN: %boogie "%s" | %OutputCheck "%s"
// CHECK-L: Boogie program verifier finished with 2 verified, 0 errors
```
does the expected thing. 

I suggest that I change the way "known_folded.bpl" runs the test to 
// RUN: %boogie "%s" | %OutputCheck "%s"
// CHECK-L: Boogie program verifier finished with 2 verified, 0 errors

That way at least the tests in the Viper folder are run in a consistent way.

Do you agree?

Thanks a lot for the comment!

zvonimir: Yes, I agree. That would be good to make uniform, and so please make this update. Then we can merge this in.

gauravpartha: I have made the update.

Issue #246, opened 2020-07-07
Incorrect CFG from structured statement


shazqadeer:
Consider the program below.

```
procedure Enter()
{
    var _old, curr: int;
    while (true) {
        call _old := CAS(0, 1);
        if (_old == 0) {
            break;
        }
        while (true) {
            call curr := Read();
            if (curr == 0) {
                break;
            }
        }
    }
}

procedure CAS(x:int, y:int) returns (z: int);
procedure Read() returns (z: int);

````
I expect to see two loops in the CFG of Enter, one nested inside the other.  However, I see three loops in the printed CFG.  Here is the output obtained using /printCFG:

digraph G {
  "anon0" [shape=box];
  "anon6_LoopHead" [shape=box];
  "anon6_LoopDone" [shape=box];
  "anon6_LoopBody" [shape=box];
  "anon7_Then" [shape=box];
  "anon7_Else" [shape=box];
  "anon3" [shape=box];
  "anon8_LoopHead" [shape=box];
  "anon8_LoopDone" [shape=box];
  "anon8_LoopBody" [shape=box];
  "anon9_Then" [shape=box];
  "anon9_Else" [shape=box];
  "anon0" -> "anon6_LoopHead";
  "anon6_LoopHead" -> "anon6_LoopDone";
  "anon6_LoopHead" -> "anon6_LoopBody";
  "anon6_LoopBody" -> "anon7_Then";
  "anon6_LoopBody" -> "anon7_Else";
  "anon7_Else" -> "anon3";
  "anon3" -> "anon8_LoopHead";
  "anon8_LoopHead" -> "anon8_LoopDone";
  "anon8_LoopHead" -> "anon8_LoopBody";
  "anon8_LoopBody" -> "anon9_Then";
  "anon8_LoopBody" -> "anon9_Else";
  "anon9_Else" -> "anon8_LoopHead";
  "anon9_Then" -> "anon0";
  "anon8_LoopDone" -> "anon6_LoopHead";
}

**Comment by bkragl** [comment]
For convenience, here is a rendering of the above CFG description.

![Enter](https://user-images.githubusercontent.com/6729391/86741599-cd667680-c037-11ea-9298-9ba378efcdf2.png)

**Comment by shazqadeer** [comment]
The situation is even worse.  Consider the following variation of the original program:
```

procedure Enter()
{
    var _old, curr: int;
    _old := 42;
    while (true) {
        call _old := CAS(0, 1);
        if (_old == 0) {
            break;
[...]

PR #247, opened 2020-07-07
Fix incorrect translation of break in a nested loop

shazqadeer:
This PR fixes issue #246.

### DIFF
--- a/Source/Core/AbsyCmd.cs
+++ b/Source/Core/AbsyCmd.cs
@@ -361,6 +361,9 @@ namespace Microsoft.Boogie {
       Contract.Requires(errorHandler != null);
       Contract.Requires(stmtList != null);
       this.stmtList = stmtList;
+      // Inject an empty big block at the end of the body of a while loop if its current end is another while loop.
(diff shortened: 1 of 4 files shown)


Source/Core/AbsyCmd.cs
```diff
@@ -391,6 +394,30 @@ int FreshAnon()
...
+        InjectEmptyBigBlockInsideWhileLoopBody(bb.ec);
+      }
+    }
+    
+    void InjectEmptyBigBlockInsideWhileLoopBody(StructuredCmd structuredCmd) {
+      if (structuredCmd is WhileCmd whileCmd) {
+        InjectEmptyBigBlockInsideWhileLoopBody(whileCmd.Body);
+        if (whileCmd.Body.BigBlocks.Count > 0) {
+          var ec = whileCmd.Body.BigBlocks.Last().ec;
+          if (ec != null && ec is WhileCmd) {
```
RustanLeino, line 408: You can write just `ec is WhileCmd`, since it will true only if if `ec` is non-null.

shazqadeer: I have simplified the conditional.

Issue #244, opened 2020-07-05
/vcsCores intermixes model output


wrwg:
The `/vcsCores` flag can dramatically speed up operation on multicore machines (e.g. my '19 MacBook pro has 8 and can easily spare 4), specifically because z3's memory usage is relative low. Unfortunately, it appears when using this option model output from different vc runs is intermixed. E.g. one output in such a situation:

```
*** MODEL
*** MODEL
...
  else -> |T@[Int]$Value!val!1|
}
Store_[$Location]$bool -> {
}
  |T@[$Location]Bool!val!1| ($Global ($StructType T@$TypeName!val!28 ($TypeValueArray |T@[Int]$TypeValue!val!0| 0) ($TypeValueArray |T@[Int]$TypeValue!val!18| 1)) 20) true -> |T@[$Location]Bool!val!3|
  else -> |T@[$Location]Bool!val!3|
}
...
*** END_MODEL
*** END_MODEL
```

This makes this flag impossible to use for tools which parse the model output. 

Is there a way to avoid this? Or does it need to be fixed?

@shazqadeer

shazqadeer:
I agree this is a problem.  There is the following option for printing the model to a file.

/printModelToFile:<file>
                print model to <file> instead of console

However, most likely the contents of the file will also get garbled if all threads are writing the models to the same file.  

One option might be to specify the model output file name via an attribute on an implementation.  If this file name is supplied, the model can be written to that file.  The client is responsible for making sure that there is a distinct file name on each implementation.

By the way, in addition to the model, wouldn't the same problem happen for counterexample traces?

wrwg:
Yes, the same would apply for execution traces.

Do you say there is an attribute which can be added fo procedures where to write the model to? Which would that be?

shazqadeer:
#258 should fix this issue.  I will wait to hear back before closing this issue.

shazqadeer:
I will close this issue.  But please reopen if the problem persists.

PR #258, opened 2020-07-15
Bring model printing in line with counterexample printing

shazqadeer:
Before this PR, counterexample printing and  model printing are happening at two different places, the former in ExecutionEngine and the latter inside the guts of VerifyImplementation.  This is undesirable for two reasons:
- Each counterexample corresponds to a particular model.  They should be printed together to make it easy to correlate the counterexample with the model.
- ExecutionEngine has synchronization in place to ensure that even when implementations are being verified in parallel, counterexample printing is serialized.  However, model printing does not have such protection. Consequently, if model printing is enabled with parallel verification, printed models are garbled (see issue #244).

This PR fixes the above problems by moving model printing to ExecutionEngine and exploiting the synchronization existing there to serialize model printing.

This PR also cleans up code  that is redundant (earlier and also as a result of this PR).

### DIFF
--- a/Source/ExecutionEngine/ExecutionEngine.cs
+++ b/Source/ExecutionEngine/ExecutionEngine.cs
@@ -450,6 +450,21 @@ namespace Microsoft.Boogie
 
     static ThreadTaskScheduler Scheduler = new ThreadTaskScheduler(16 * 1024 * 1024);
 
+    private static TextWriter modelWriter = null;
+    public static TextWriter ModelWriter {
+      get {
+        if (CommandLineOptions.Clo.PrintErrorModelFile == null)
+        {
+          return null;
+        }
+        if (ExecutionEngine.modelWriter == null)
+        {
+          ExecutionEngine.modelWriter = new StreamWriter(CommandLineOptions.Clo.PrintErrorModelFile, false);
+        }
+        return ExecutionEngine.modelWriter;
+      }
+    }
+    
     public static void ProcessFiles(List<string> fileNames, bool lookForSnapshots = true, string programId = null)
     {
       Contract.Requires(cce.NonNullElements(fileNames));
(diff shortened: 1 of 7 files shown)


Source/VCGeneration/ConditionGeneration.cs
```diff
@@ -635,8 +635,6 @@ public enum Outcome {
     protected string/*?*/ logFilePath;
     protected bool appendLogFile;
 
-    public static List<Model> errorModelList;
```
shazqadeer, line 638: This field appears to be redundant because the model  is available as a field of Counterexample itself.  Hence I removed it and simplified the code in a particular variation of VerifyImplementation (see below).

Test/test15/ModelTest.bpl.expect
```diff
@@ -1,3 +1,6 @@
+ModelTest.bpl(9,3): Error BP5001: This assertion might not hold.
```
shazqadeer, line 1: I don't understand the diff in these files.  I can't see any difference.  Perhaps some hidden character ...

bkragl: The error trace appears above the model now. I think this actually makes more sense.

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -1677,6 +1692,10 @@ private static void ProcessErrors(List<Counterexample> errors, VC.VCGen.Outcome
           {
             errorInfo.Out.WriteLine("Execution trace:");
             error.Print(4, errorInfo.Out, b => { errorInfo.AddAuxInfo(b.tok, b.Label, "Execution trace"); });
+            if (CommandLineOptions.Clo.PrintErrorModel >= 1 && error.Model != null)
+            {
+              error.Model.Write(ExecutionEngine.ModelWriter == null ? errorInfo.Out : ExecutionEngine.ModelWriter);
+            }
```
bkragl, line 1698: 1. Does the synchronization in `ExecutionEngine` ensure that there is no race on the initialization of `ExecutionEngine.ModelWriter`?
2. I was at first surprised that `ExecutionEngine.ModelWriter` is only initialized when writing to a file, until I saw that `errorInfo.Out` is used here when not writing to a file.
3. I saw that previously ModelWriter was flushed. Is this not necessary here?

shazqadeer: 1. Good catch.  It does not.  I have changed the logic now so that ModelWriter is initialized at the beginning of InferAndVerify before all the parallelization starts.

3. The Flush is happening in OutputCollector.WriteMoreOutput.

Issue #237, opened 2020-05-28
Boogie source code formatting


zvonimir:
It occasionally really bugs me how Boogie source code formatting is all over the place.

I think that Boogie traditionally went with this style:
```
if (...)
{

}
else
{

}
```

However, now there are also (large) code sections written using this style:
```
if (...) {

} else {

}
```

Which one should we be using?

Is there a good way to reformat the whole code base without losing nice git history/blame/diff functionality?

shazqadeer:
In addition, indentation is inconsistent.  In some files, it is two spaces and in others it is four.

zvonimir:
There is this tool that seems should be easy to add: https://github.com/DotNetAnalyzers/StyleCopAnalyzers

PR #259, opened 2020-07-16
Fix formatting

zvonimir:
This is trying to address #237. I turned on formatting checking, and I also reformatted all files. There is still thousands of warnings being issued. Lots of them make sense and I think they do improve code readability. Others we can probably suppress.

The question is whether we want to go down this route to fix/improve formatting and make it consistent in the long run, or not. There is quite a lot of work left to fix all/most of the warnings, which is why I created this branch in the main repo so that everyone can contribute :).


bkragl:
I did two configuration changes.

* Enable StyleCop globally in one place (the file `Directory.Build.props`).
* Put the exclusion comments for scanner/parser into `Copyright.frame`. Otherwise they would not survive a regeneration.

Some thoughts.

1. If we go for this, I think that any reformatting needs to be merged rather quickly. Otherwise it will be hard to resolve merge conflicts (we already have some now).
2. We will lose some value of the Git history (similar to a past commit that touched almost all files to fix line endings), and going piece by piece (either in terms of files or types of warning we enable) will be worse than a single commit. On the other hand, I'm not sure if fixing 27251warnings quickly and at once is realistic.
3. Did you do all the already applied fixes manually? Or is there some automation support?
4. The reason why CI is currently failing is the following :)
     ```
     The job exceeded the maximum log length, and has been terminated.
     ```

zvonimir:
Yeah, git blame/history won't work as well any more. So there are trade-offs here.

I did all the fixes automatically using Rider IDE, which I think uses the same tool under the hood. But it doesn't fix everything.

I think the way to do this (if we want to go that route) is to suppress all the warnings we still have (there must be a way to do that), and then merge whatever we have. That way we will catch all future style issues going forward and we will avoid merging them.

Once that is done, then we slowly patch up other warnings one by one until we are happy.

Thoughts?

shazqadeer:
I don't understand the issue with git blame/history.  Specifically, what will not work?

zvonimir:
Since these massive updates touch a lot of lines of code, many git blames will land on this large commit. I think one can usually get past that by repeating git blame, and all is good. But I also vaguely remember that I think that sometimes I did not manage to do that in SMACK, where we did similar major reformatting at some point several years ago.
In short, I don't think it is a major problem, but more of an annoyance. Maybe @bkragl knows more?

shazqadeer:
I looked a bit at the huge number of warnings being issued by Style Cop. There is a lot of unnecessary drama going on there.  May I ask what are the specific pet peeves of the folks interesting this cleanup.  Mine are the following:
- inconsistent indentation
- inconsistent naming of fields, variables, and methods
Consistent positioning of braces for blocks of code would be good to have but less of a concern than the two issues above.

We could simply take the union of the concerns expressed here, fix things up, and land the PR ASAP.  Later on, when somebody cares about the consistency of something else, they can initiate it.

zvonimir:
I like some of the other suggestions as well. But yeah, the ones you listed are the most important to me as well.
So I will try to figure out how to turn on checking of just these rules, and then merge it in. I'll probably do that later tonight.

zvonimir:
@shazqadeer and @bkragl : I got this into a reasonable stage. I reformatted everything and I have StyleCop up and running. There are a bunch of rules I had to disable, and so I am not totally sure what is being checked right now. I will slowly enable the important ones as I keep fixing the code.

Honestly, I am not totally sure this is the right move, but hopefully it'll help us to maintain code consistency in the long run.

shazqadeer:
I am in the middle of some other PRs. Should  I wait until your PR lands?

shazqadeer:
I just want to say that I am supportive of the change.  This makes for a good starting point for other highly-overdue code refactoring and clean up.

zvonimir:
I was just about to write the same thing. While StyleCop is not perfect and has some issues (e.g., do-while is a problem), based on my experience so far, it does help with maintaining consistency. Also, code reformatting was a step in the right direction as well I think. So I say we merge this in.
People with outstanding pull requests will have to put some effort into merging.

zvonimir:
@bkragl Done and done. Thanks!

shazqadeer:
@zvonimir :  Are you close to landing this?  I am blocked behind you.

PR #261, opened 2020-07-19
Add support for per implementation random seed

shazqadeer:
1. Before this PR, Timeout and Rlimit were set in PrepareCommon.  Therefore, it was not possible to set these limits in a fine-grained way per verification problem. Now, these are set for each verification problem inside BeginCheck.  

2. This  PR adds support for adding a random seed for each implementation.  This is useful for experimenting with different random seeds and then affixing one to the implementation once discovered so that deterministic execution is possible across machines. A key change accompanying this feature addition is that the random seed is reset to 0 before each VC is verified.  Thus, each VC is verified from the same initial random seed, thus reducing nondeterministic behavior I have observed in the past where VC may be verified or not depending on how many other VCs in the file are being verified.

3. ProverKillTime is renamed to TimeLimit and its default value is changed to 0 (as opposed to -1) leading to clean up in various parts of the code.

4. The option CivlDesugaredFile is renamed to civlDesugaredFile for consistency.

### DIFF
--- a/Source/Provers/SMTLib/ProverInterface.cs
+++ b/Source/Provers/SMTLib/ProverInterface.cs
@@ -429,7 +429,11 @@ namespace Microsoft.Boogie.SMTLib
         SendCommon("(set-info :smt-lib-version 2.6)");
         if (options.ProduceModel())
           SendCommon("(set-option :produce-models true)");
-
+        foreach (var opt in options.SmtOptions)
+        {
+          SendThisVC("(set-option :" + opt.Option + " " + opt.Value + ")");
+        }
+        
         if (!string.IsNullOrEmpty(options.Logic))
         {
           SendCommon("(set-logic " + options.Logic + ")");
@@ -580,13 +584,12 @@ namespace Microsoft.Boogie.SMTLib
 
       SendThisVC("(push 1)");
       SendThisVC("(set-info :boogie-vc-id " + SMTLibNamer.QuoteId(descriptiveName) + ")");
-
-      // send per-VC options
-      foreach (var opt in options.SmtOptions)
+      if (options.Solver == SolverKind.Z3)
       {
-        SendThisVC("(set-option :" + opt.Option + " " + opt.Value + ")");
+        SendThisVC("(set-option :" + Z3.TimeoutOption + " " + options.TimeLimit + ")");
+        SendThisVC("(set-option :" + Z3.RlimitOption + " " + options.ResourceLimit + ")");
+        SendThisVC("(set-option :" + Z3.RandomSeedOption + " " + options.RandomSeed + ")");
       }
-      // send VC
       SendThisVC(vcString);
 
       SendOptimizationRequests();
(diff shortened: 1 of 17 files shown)


Source/Provers/SMTLib/ProverInterface.cs
```diff
@@ -2654,6 +2656,17 @@ public override void SetRlimit(int limit)
...
     }
 
+    public override void SetRandomSeed(int randomSeed)
+    {
+      if (options.Solver == SolverKind.Z3)
+      {
+        var name = Z3.RandomSeedOption;
+        var value = randomSeed.ToString();
+        options.SmtOptions.RemoveAll(ov => ov.Option == name);
+        options.AddSmtOption(name, value);
```
bkragl, line 2666: `options` (which is of type `SMTLibProverOptions`) has methods `AddSmtOption` which just adds to the list of options, and `AddWeakSmtOption` which only adds an option if no entry for the same name exists yet. Maybe we should add a third method for adding/overriding options (encapsulating the call to `RemoveAll` here and in `SetTimeout` and `SetRlimit` above).

Source/VCGeneration/Check.cs
```diff
@@ -410,8 +392,12 @@ public void BeginCheck(string descriptiveName, VCExpr vc, ProverInterface.ErrorH
...
 
       thmProver.Reset(gen);
-      SetTimeout();
-      SetRlimit();
+      SetTimeout(timeout);
+      SetRlimit(rlimit);
+      if (randomSeed.HasValue)
+      {
+        SetRandomSeed(randomSeed.Value);
+      }
```
bkragl, line 400: In the else case, shouldn't the random seed be set to the default value 0? Because otherwise a custom seed of one implementation would leak to a subsequent implementation that does not declare a custom seed. Basically this means that a "set seed" command should be sent to the solver for every VC (as it is the case for timeout and rlimit).
If yes, I would move the if-then-else statement into `SetRandomSeed`.

bkragl: I just tested, and there is actually a `(set-option :smt.random_seed ...)` command issued for every VC, "resetting" to `0` if nothing is specified on the implementation. But judging by the way the `RandomSeed` property of `NamedDeclaration` is implemented, this is by accident.
`CheckIntAttribute` has the somewhat unintuitive behavior of returning false only when there was a "parsing error" (i.e., when an attribute with the right name is found, but it is not an integer expression). However, it returns true and leaves the ref parameter unchanged when no attribute with the right name is found.
Shouldn't `RandomSeed` be non-nullable and implemented similarly as `TimeLimit` and `ResourceLimit`?

shazqadeer: If we were to set random seed before verifying each implementation to 0, that would change the behavior currently implemented in Boogie?  Are you saying that would be a change for the better?

Source/Provers/SMTLib/ProverInterface.cs
shazqadeer: These method APIs appear strange to me.  There should be just one method SetSmtOption, which should either add or override as the case might be.  This is easy and fast to do if I change the representation from a list to a dictionary.

Source/VCGeneration/Check.cs
shazqadeer: I suppose it would add more determinism.  We have often seen how individual implementations may be verified alone using /proc: but not when verified together with other implementations (and vice versa).  By ensuring that the random seed is set to 0 (which also happens to be Z3's default), we will eliminate this source of nondeterminism.

Source/Provers/SMTLib/ProverInterface.cs
shazqadeer: Timeout, Rlimit, and RandomSeed are not put into the SmtOptions list any more.  They are set directly inside BeginCheck now.

Source/VCGeneration/Check.cs
shazqadeer: After some thinking, I concluded that resetting the random seed to 0 before pushing each verification condition is indeed a step in the right direction and should help us reduce the butterfly effect in Boogie usage.  The principle here is that provability of a stable verification condition should not depend on the random seed but if it does it should be explicitly admitted via an annotation.

Issue #79, opened 2017-11-28
/errorLimit parameter quirks (default and 0-parameter behaviours)


alexanderjsummers:
By default, Boogie limits the number of errors reported per procedure. Since this isn't documented except in the /help, I found it a bit surprising when first flagged.

It seems that the /errorLimit:n parameter is, by default, set to 5. If one sets this to a higher number, more errors are reported. However, there doesn't seem to be a way (in principle) to avoid limiting the number of errors seen (of course, one can just choose a very large number each time). Passing /errorLimit:0 behaves analogously to /errorLimit:1 - one sees neither all nor zero errors.

I suggest that the default behaviour should be _not_ to include a hard limit on the number of errors produced per procedure. One could discuss how a 0 parameter should be treated - personally I would find the "no limit" interpretation more useful there, too. In any case, the 0 shouldn't be treated as a 1.

RustanLeino:
I agree that it would be good to support a no-limit mode. Regardless of what the default limit is, it is useful to be able to specify any option (including no-limit) from the command line. Furthermore, whatever the default is should be reported with the `/help` switch.

  Rustan

shazqadeer:
@alexanderjsummers : Do you still care about this issue?  If yes, please suggest a strategy for implementing unlimited number of detected errors.  If not, I will simply change Boogie to print out an error message for /errorLimit:0 and close this issue.

alexanderjsummers:
Hi Shaz,

Yes - we use Boogie assuming that it will report all errors. I assume other users of Boogie as a back-end might do the same (but perhaps not realise that some errors are omitted). By strategy, do you mean where in the code the limit should be removed? Or is there a deeper issue that I'm not aware of?

What is the motivation for having a limit at all (by default)?

Thanks, Alex

shazqadeer:
What does it mean to report all errors?

alexanderjsummers:
Hi Shaz. I'm not sure if understand your question. 

I believe that Boogie currently treats each statement generating a verification requirement as an assert-assume pair; in this way, the remainder of the program is still analysed under the assumption that all previous verification conditions were successful. This is how I learned when I should expect Boogie to e.g. produce 2 rather than 1 error, and when I should not (e.g. because a later statement's VC is entirely subsumed by the previous assumptions). By the same logic, I understand when I expect an (n+1)th error from a particular line, and can incrementally walk through the program and figure out which lines should report a further error (or not).

[...]

PR #262, opened 2020-07-20
Change the meaning of /errorLimit:0 so that it returns all errors

shazqadeer:
This PR addresses issue #79 .  If the user passes /errorLimit:0 on the command line, all assertions in a verification condition are attempted to be falsified.


Source/Provers/SMTLib/ProverInterface.cs
```diff
@@ -1733,8 +1730,11 @@ public override Outcome CheckOutcomeCore(ErrorHandler handler, int taskID = -1)
 
               handler.OnModel(labels, model, result);
             }
-
-            if (labels == null || !labels.Any() || errorsLeft == 0) break;
+            
+            // errorsDiscovered > 0 at this control location.
```
bkragl, line 1734: Make this `Debug.Assert(errorsDiscovered > 0)`?

shazqadeer:
We should delete z3api folder  from the repository.

PR #264, opened 2020-07-20
remove z3api

shazqadeer:
Using Z3 via an API is no longer supported.  This PR deletes the project that provided this support in the past and the accompanying tests.


shazqadeer:
Done.

Issue #19, opened 2015-07-30
Inlining and variable capture


shuvendu-lahiri:
When Boogie inlines the spec of b inside foo, the bound variable "a" in ensures clause conflicts with the parameter "a" of foo. Is this a known bug/feature, and how to people get around it (e.g. obscure names of bound variables in quantifiers)?

For the following program foo.bpl

```
type ref;
var arr:[int]int;

procedure {:inline 1} b()
modifies arr;
ensures (forall a:int  :: {arr[a]} a < 10 ==> arr[a] == 0);
{
}
procedure foo(a:ref) 
modifies arr;
{
   call b();
}
```

When I generate the inlined version:
boogie /noVerify foo.bpl /printInstrumented > foo_1.bpl

and remove the first/last lines (Boogie verifier ...), then I get the following type checking error:

boogie /noVerify foo_1.bpl

Boogie program verifier version 2.3.0.61016, Copyright (c) 2003-2014, Microsoft.
foo_1.bpl(40,19): Error: more than one declaration of variable name: a
foo_1.bpl(40,29): Error: trigger must mention all quantified variables, but does not mention: a
2 name resolution errors detected in foo_1.bpl

RustanLeino:
That is a bug. Boogie should rename the "a" in the quantifier (both in the "ensures" clause as in the example, and also if there was a quantifier over "a" somewhere in the body of the inlined procedure), just like it currently renames any local variables of the inlined procedure.

  Rustan

PR #265, opened 2020-07-21
Fixed bug in inliner

shazqadeer:
This PR closes issue #19. 
- Make CodeCopier a private class of Inliner; external uses of CodeCopier are retargeted to Duplicator.
- Add renaming of bound variables in lambdas, quantifiers, and let expressions during inlining.
- Remove unnecessary modifies computation during inlining.

Sorry it has taken such a long time to fix this issue.


shazqadeer:
This fix currently handles only bound variables in quantifiers but not in lambdas.  In the current Boogie pipeline, inlining happens before lambda conversion into quantifiers.  We have to either include renaming of lambda-bound variables in this PR or move lambda conversion before inlining.  Opinions?

PR #267, opened 2020-07-22
Allow long quantifier identifiers

wilcoxjay:
When generating a verification condition, Boogie automatically gives ":qid"s to quantifiers based on their position in the source files. These names are especially helpful when profiling quantifier instantiation performance. However, Boogie previously limited the filename to the first 8 characters, which sometimes made it difficult to tell which exact file was being referenced. This PR removes this upper limit.

I am open to discussion on the approach, because I didn't really understand the comments in the code. It says the filenames were limited in length "so the SX file doesn't grow too big". I'm not sure what SX means, but assuming it means the `.smt2` file, this doesn't seem like a big issue to me. If people feel otherwise, we could make it a configurable option, but I kind of feel like that's overkill for this feature.

### DIFF
--- a/Source/VCExpr/Boogie2VCExpr.cs
+++ b/Source/VCExpr/Boogie2VCExpr.cs
@@ -665,12 +665,8 @@ namespace Microsoft.Boogie.VCExprAST
       {
         // generate default name (line:column position in .bpl file)
         Variable v = vars[0];
-        Contract.Assert(v != null); // Rustan's claim!
-        // Include the first 8 characters of the filename in QID (helpful in case we use /concat)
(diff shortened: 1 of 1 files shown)


shazqadeer:
.sx was the file extension for Simplify, the original SMT solver and the first backend for Boogie.

wilcoxjay:
> You could also delete the mysterious comment "Rustan's claim!".

done

shazqadeer:
Does git show you the option to merge the PR?  If so, you can merge it yourself.  If not, I can do the merge for you.

wilcoxjay:
No, I don't have commit privileges to Boogie. Please go ahead and merge it. Thanks!

PR #268, opened 2020-08-01
Allow using a Stream to parse a program that still need to be pre-processed

keyboardDrummer:
### Changes
Allow using a Stream to parse a program that still need to be pre-processed. This change is required for [this Dafny PR](https://github.com/dafny-lang/dafny/pull/761).

### Testing
Since this is a small refactoring we can rely on existing tests

### DIFF
--- a/Source/Core/BoogiePL.atg
+++ b/Source/Core/BoogiePL.atg
@@ -31,26 +31,26 @@ readonly StructuredCmd/*!*/ dummyStructuredCmd;
 ///Returns the number of parsing errors encountered.  If 0, "program" returns as
 ///the parsed program.
 ///</summary>
-public static int Parse (string/*!*/ filename, /*maybe null*/ List<string/*!*/> defines, out /*maybe null*/ Program program, bool useBaseName=false) /* throws System.IO.IOException */ {
+public static int Parse (string/*!*/ filename, List<string/*!*/> defines, out /*maybe null*/ Program program, bool useBaseName=false) /* throws System.IO.IOException */ {
   Contract.Requires(filename != null);
   Contract.Requires(cce.NonNullElements(defines,true));
 
-  if (defines == null) {
-    defines = new List<string/*!*/>();
-  }
-
   if (filename == "stdin.bpl") {
-    var s = ParserHelper.Fill(Console.In, defines);
-    return Parse(s, filename, out program, useBaseName);
-  } else {
-    FileStream stream = new FileStream(filename, FileMode.Open, FileAccess.Read, FileShare.Read);
-    var s = ParserHelper.Fill(stream, defines);
-    var ret = Parse(s, filename, out program, useBaseName);
-    stream.Close();
-    return ret;
+    return Parse(Console.In, filename, defines, out program, useBaseName);
+  } else
+  {
+    using (FileStream stream = new FileStream(filename, FileMode.Open, FileAccess.Read, FileShare.Read))
+    {
+      return Parse(new StreamReader(stream), filename, defines, out program, useBaseName);
+    }
   }
 }
 
+public static int Parse(TextReader stream, string/*!*/ filename, List<string> defines, out /*maybe null*/ Program program, bool useBaseName=false) /* throws System.IO.IOException */ 
+{
+  var preprocessedSource = ParserHelper.Fill(stream, defines);
+  return Parse(preprocessedSource, filename, out program, useBaseName);
+}
 
 public static int Parse (string s, string/*!*/ filename, out /*maybe null*/ Program program, bool useBaseName=false) /* throws System.IO.IOException */ {
   Contract.Requires(s != null);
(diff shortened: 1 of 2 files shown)


bkragl:
`Parser.cs` is a generated file. Could you please edit `BoogiePL.atg` and regenerate `Parser.cs` from it. Thanks!

keyboardDrummer:
I will. How do you activate the generator?

bkragl:
There is a `Makefile` in `Source/Core`, so you can just run `make` in that directory. It will tell you to run `dotnet tool restore` somewhere in your Boogie clone to install the compiler generator (called Coco/R).

keyboardDrummer:
Thanks. Would it make sense not to have the generated file be checked into source but instead to generate it on build? The generation time seems instant and the Coco dependency can be pulled in automatically, so I don't see any downsides.

Source/Core/BoogiePL.atg
```diff
@@ -42,15 +42,20 @@ public static int Parse (string/*!*/ filename, /*maybe null*/ List<string/*!*/>
...
+  } else
+  {
+    using (FileStream stream = new FileStream(filename, FileMode.Open, FileAccess.Read, FileShare.Read))
+    {
+      return Parse(stream, filename, defines, out program, useBaseName);
+    }
   }
 }
 
+public static int Parse(Stream stream, string/*!*/ filename, List<string> defines, out /*maybe null*/ Program program, bool useBaseName)
```
bkragl, line 54: Perhaps make the signature consistent with the above `Parse` method (default value of `useBaseName`, throws annotation).

bkragl:
> Thanks. Would it make sense not to have the generated file be checked into source but instead to generate it on pre-build? The generation time seems instant and the Coco dependency can be pulled in automatically.

Yes, I'd love to see a PR for this :)

I think this was not done before, because parser/scanner were modified rather infrequently, and getting (the correct version of) Coco was a hassle. However, since there is a NuGet package for Coco now, it might actually be quite easy to integrate the generation into the .NET Core build. The only caveat is, that there is a legacy build (standard .NET framework) which we seem to keep alive, and so the integration has to be done for this build too.

keyboardDrummer:
> The added functionality looks fine to me. However, there are a few things that look odd to me.
> 
> * Parameter `defines` can be `null` in the existing `Parse` method (and is initialized in this case), but not in the added `Parse` method. Perhaps the initialization should be removed, since the only caller within Boogie initialized the parameter before the call.
> * It would be nice to unify the call to `ParserHelper.Fill` regardless of whether we read from a file or from standard input.
> 
> Perhaps this are issues to address in another PR? @shazqadeer @zvonimir

Good suggestions, made the changes.

keyboardDrummer:
> > Thanks. Would it make sense not to have the generated file be checked into source but instead to generate it on pre-build? The generation time seems instant and the Coco dependency can be pulled in automatically.
> 
> Yes, I'd love to see a PR for this :)
> 
> I think this was not done before, because parser/scanner were modified rather infrequently, and getting (the correct version of) Coco was a hassle. However, since there is a NuGet package for Coco now, it might actually be quite easy to integrate the generation into the .NET Core build. The only caveat is, that there is a legacy build (standard .NET framework) which we seem to keep alive, and so the integration has to be done for this build too.

Couldn't get it to work for the standard .NET framework solution: https://github.com/boogie-org/boogie/pull/269

RustanLeino:
FYI, [Dafny](https://github.com/dafny-lang/dafny) also uses Coco as its parser generator. The Dafny project includes the Coco binary, https://github.com/dafny-lang/dafny/tree/master/third_party/Coco. Including binaries is not generally recommended, but it seemed the best solution at the time and it has worked fine.

bkragl:
> FYI, [Dafny](https://github.com/dafny-lang/dafny) also uses Coco as its parser generator. The Dafny project includes the Coco binary, https://github.com/dafny-lang/dafny/tree/master/third_party/Coco. Including binaries is not generally recommended, but it seemed the best solution at the time and it has worked fine.

People were protesting against including a Coco binary in the Boogie repository. I think the current solution to run Coco as a dotnet tool is much better than including a binary, both in terms of platform independence and licensing issues. Actually, Dafny could do the same to get rid of the `third_party` directory.

What @keyboardDrummer wants is to go a step further and remove the generated files `Parser.cs` and `Scanner.cs` form source control, and instead run Coco as part of the build process.

zvonimir:
I am actually a bit worried how removing these two auto-generated files will affect projects downstream, such as Corral. I haven't looked at the details, but is this going to affect the build process of Corral?
If so, then I think we should not do this. Thx!

bkragl:
> I am actually a bit worried how removing these two auto-generated files will affect projects downstream, such as Corral. I haven't looked at the details, but is this going to affect the build process of Corral?
> If so, then I think we should not do this. Thx!

Let's move this discussion to #269.

PR #269, opened 2020-08-03
Remove generated parser

keyboardDrummer:
Includes [this PR](https://github.com/boogie-org/boogie/pull/268) to prevent a merge conflict.

### Changes
Remove `Core/Parser.cs` and `Core.Parser.frame` from source and ignore them in git, in favor of generating them as a pre-build step.

### Testing
- Since this is a refactoring of common functionality, no additional testing is needed for Boogie-NetCore.sln
- Manual testing shows this fails to build for Boogie.sln, the standard .NET framework solution. I also can't get it to work for that, so I guess I should close this PR, right?


bkragl:
I would suggest to not do this change at the moment. One reason (as pointed out by @zvonimir) is that it will affect people who still want to compile Boogie on a machine without `dotnet` installed. Also, it would be nice to eventually have a solution where `dotnet restore` and `coco` do not have to be invoked explicitly as a script. I imagine referencing Coco in a similar way as FsLexYacc is referenced by Forro:
https://github.com/boogie-org/forro/blob/master/src/Forro/Forro.fsproj

Two general remarks:
* Besides `Parser.cs`, also `Scanner.cs` is generated by Coco.
* On the contrary, `Parser.frame` and `Scanner.frame` are **not** generated. These files are custom to Boogie and their directory needs to be passed to Coco with the command-line flag `-frames`.

**For now I would suggest the following:**
In order to make sure that people do not directly edit `Parser.cs`/`Scanner.cs` instead of the grammar file `BoogiePL.atg`, I would run Coco as part of CI and abort if the generated files differ from the ones in Git. Thoughts?

zvonimir:
Great idea @bkragl!

zvonimir:
I guess we are not proceeding with this one, and so it should probably be closed. @bkragl? Thanks!

keyboardDrummer:
Thanks for the comments. Given that this change breaks .NET framework compatibility I don't think it makes sense at this point. The custom `FsLexYacc` `ItemGroup` type is interesting. Do you happen to know how to define that? I can't find it in https://github.com/fsprojects/FsLexYacc

bkragl:
I don't have experience with this, but it looks like the build tasks are defined here:
https://github.com/fsprojects/FsLexYacc/blob/master/src/FsLexYacc.Build.Tasks/FsLexYacc.targets

PR #274, opened 2020-08-06
Allow default random_seed setting via /proverOpt:O:smt.random_seed=... command line option

shazqadeer:
This PR allows random seed of an implementation to be null in which case any seed supplied globally via proverOpt:0:smt.random_seed=... becomes applicable.  The output generated via /proverLog:... for Test\test2\RandomSeed.bpl is instructive.

### DIFF
--- a/Source/Core/Absy.cs
+++ b/Source/Core/Absy.cs
@@ -1572,10 +1572,9 @@ namespace Microsoft.Boogie
       return QKeyValue.FindStringAttribute(this.Attributes, name);
     }
 
-    // Look for {:name N} or {:name N} in list of attributes. Return result in 'result'
-    // (which is not touched if there is no attribute specified).
-    //
-    // Returns false is there was an error processing the flag, true otherwise.
+    // Look for {:name N} in list of attributes. Return false if attribute
+    // 'name' does not exist or if N is not an integer constant.  Otherwise,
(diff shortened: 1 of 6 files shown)


bkragl:
Sorry, but I'm not quite sure if this PR is really as intended. It deviates from my understanding of what we previously discussed. Namely that `:smt.random_seed` should be set for every VC.

Consider the previous version of `RandomSeed.bpl`:

```
// RUN: %boogie -proverLog:%t "%s"
// RUN: %OutputCheck --file-to-check "%t" "%s"
// CHECK-L: (set-option :smt.random_seed 100)
// CHECK-L: (set-option :smt.random_seed 0)
procedure {:random_seed 100} WithRandomSeed()
{
}
procedure WithoutRandomSeed()
{
}
```

It would fail now because the random seed for the second VC is not set to the default `0`. Is this really the desired behavior? I would think no, because it can make VCs unstable (depending on their position in a file).

I'm wondering if it wouldn't be better to introduce a top-level option for the random seed (with a default value of `0`), and handle it exactly the same as `rlimit` and `timeLimit`?

zvonimir:
I see. This boils down to what the semantics of Z3's `reset` is, which I think gets issued before every VC. My conjecture is that it resets all options to its default values, but I could be wrong here.

PR #276, opened 2020-08-18
Refactor type checking of yield invariant calls and introduce :yield_preserves

### LINKED ISSUE #251: [CIVL] Add :yield_preserves
Currently, yield_requires and yield_ensures are often duplicated with the same call to yield invariant.  Introduce yield_preserves to avoid thus duplication.

bkragl:
This PR unifies and encapsulates the type checking of calls to yield invariants inside `YieldInvariantCallChecker`. A bug is fixed where passing a linear output parameter of a procedure to a yield invariant was rejected.

The `:yield_preserves` attribute is introduced as a shorthand for the same yield invariant call to be used both as a precondition and as a postcondition. Examples are ported to use `:yield_preserves`.

Closes #251.


bkragl:
@shazqadeer: Yes, I think you are right. There should not be any ambiguity because global variables are anyway only allowed within `old(.)` in `:yield_ensures`. So in fact the implementation was wrong since it leaked global variable occurrences verbatim from `:yield_preserves` into `:yield_ensues`.

PR #277, opened 2020-08-24
Ensure yield sufficiency w.r.t. async calls

bkragl:
We check that there is no update to global variables between an async call and the next yield. The check is based on the computation of a simulation relation between the appropriately labeled CFG of a procedure and a specification automaton.

The example `async-yield-sufficiency.bpl` demonstrates the unsoundness that existed without this check.

Some refactoring in the existing yield sufficiency check.

### DIFF
--- a/Test/civl/async-yield-sufficiency.bpl
+++ b/Test/civl/async-yield-sufficiency.bpl
@@ -1,0 +1,42 @@
+// RUN: %boogie -useArrayTheory "%s" > "%t"
+// RUN: %diff "%s.expect" "%t"
+
+var {:layer 0,1} x:int;
+
+procedure {:yields}{:layer 1} P({:linear_in "tid"} tid1:int, {:linear "tid"} tid2:int)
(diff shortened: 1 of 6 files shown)


Test/civl/async-yield-sufficiency.bpl
```diff
@@ -0,0 +1,42 @@
...
+requires {:layer 1} tid2 == 2;
+requires {:layer 1} mem[2] == 0;
+{
+  async call Q(tid1);
+  call write();
+}
+
+procedure {:yields}{:layer 1} Q({:linear "tid"} tid1:int)
+requires {:layer 1} tid1 == 1;
+requires {:layer 1} mem[2] == 0; // This precondition is not valid at the end of procedures P
```
shazqadeer, line 17: procedures -> procedure

bkragl: Fixed.

Issue #272, opened 2020-08-05
Use and documentation for :msg


RustanLeino:
Boogie allows `assert` statements and `requires` and `ensures` declarations to have an

    {:msg "My error message here}

attribute. This is not documented by the `/attrHelp` switch, but it should be.

Also, use of this feature causes the internal `ErrorInformation` object created to have a `null` source token. Specifically, in ExecutionEngine.cs, static method `CreateErrorInformation` inspects the `.ErrorMessage` field for the `requires`/`ensures` and then calls the instance method `CreateErrorInformation` with `null` as the first parameter. The Code Contract for that method says:

    Contract.Requires(1 <= tok.line && 1 <= tok.col);

This is a bit dubious, but it seems to imply that `tok` is expected to be non-null. For an `assert`, the static method `CreateErrorInformation` also sets this `tok` to `null` on at least one branch.

This null-token bug causes a crash in Dafny, which regrettably renders the `:msg` feature useless for Dafny programs. (An alternative solution would be to anticipate a `null` token in the `ErrorInformation` in Dafny, but it feels more like a Boogie issue to me.)

shazqadeer:
I am happy to fix this in Boogie.  But what is the right fix?  I see two options:
- Use the token of the failing specification (requires/ensures/assert)
- Use Token.NoToken

The first option will not require changing the contract of CreateErrorInformation (the instance method) but the second one will require a change to:
Contract.Requires(0 <= tok.line && 0 <= tok.col);

bkragl:
I would vote for the token of the failing spec. Although it would also possible to use the token of the failing call (in case of precondition violation) and the token of the failing return (in case of a postcondition violation).

Besides the `:msg` attribute, there is a field called `ErrorData` to programmatically set error messages on calls, preconditions, postconditions, and assertions. I'm wondering if there should be support for `:msg` on calls. Adding `ErrorData` to return commands would also be possible, but I imagine that `:msg` does not work because an internal return command does not necessarily correspond to a command in the input program on which that attribute can be attached to.

PR #282, opened 2020-09-03
Fix for issue 272

shazqadeer:
Fixes issue #272.

### DIFF
--- a/Source/ExecutionEngine/ExecutionEngine.cs
+++ b/Source/ExecutionEngine/ExecutionEngine.cs
@@ -9,6 +9,7 @@ using System.Threading.Tasks;
 using VC;
 using BoogiePL = Microsoft.Boogie;
 using System.Runtime.Caching;
+using System.Diagnostics;
 
 namespace Microsoft.Boogie
 {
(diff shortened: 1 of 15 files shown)


Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -1925,27 +1931,28 @@ private static ErrorInformation CreateErrorInformation(Counterexample error, VC.
...
-          if (!CommandLineOptions.Clo.ForceBplErrors && assertError.FailingAssert.ErrorMessage != null)
+          string msg = null;
+          string bec = null;
+          if (assertError.FailingAssert.ErrorMessage == null || CommandLineOptions.Clo.ForceBplErrors)
+          {
+            msg = assertError.FailingAssert.ErrorData as string;
+            if (msg == null)
+            {
+              msg = "This assertion might not hold.";
+              bec = "BP5001";
```
bkragl, line 1942: Should we get rid of this case distinction and set `bec` in both cases? This would be consistent to pre/postcondition violations.

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -1925,27 +1931,28 @@ private static ErrorInformation CreateErrorInformation(Counterexample error, VC.
...
+            }
+          }
+          else
           {
             msg = assertError.FailingAssert.ErrorMessage;
-            tok = null;
             if (cause == "Error")
```
bkragl, line 1948: Can we remove this branch? I don't see the point.
It would then print `Error: ` in front of the message, which would be consistent with all other cases.
@alexanderjsummers @gauravpartha: I saw that Viper makes use of the `:msg` attribute. Would this break any of your parsing of Boogie's output?

bkragl (commented):
I did some further refactoring of the code to make it more readable (no semantic change).
Inlined below are two suggestions for changes to discuss.

shazqadeer:
I am ok with both changes proposed by @bkragl .

bkragl (approved):

PR #283, opened 2020-09-07
Fix contract and usages of LabelOrCmd in parser

bkragl:
Invalid programs like
```
procedure foo ()
{
  call
}
```
caused the following assertion violation and crash in Boogie:
```
crash.bpl(4,1): error: ident expected
Process terminated. Assertion failed.
   at Microsoft.Boogie.Parser.StmtList(StmtList& stmtList) in /home/bkragl/Code/boogie/Source/Core/Parser.cs:line 963
   ...
```

The contract `(c == null) != (label != null)` of `LabelOrCmd` was bogus. I guess what was meant is `(c == null) != (label == null)`, stating that either `c` is non-null or `label` is non-null (but not both). However, this does not seem to be true. In case of parsing a program with syntax errors, both `c` and `label` can be null. I fixed the two usages of `LabelOrCmd` by doing nothing in this case. I assume this is fine, because an error must have been already recorded somewhere.

### DIFF
--- a/Source/Core/BoogiePL.atg
+++ b/Source/Core/BoogiePL.atg
@@ -705,15 +705,14 @@ StmtList<out StmtList/*!*/ stmtList>
 
   {
   ( LabelOrCmd<out c, out label>
-    (. if (c != null) {
+    (. Contract.Assert(c == null || label == null);
+       if (c != null) {
          // LabelOrCmd read a Cmd
-         Contract.Assert(label == null);
          if (startToken == null) { startToken = c.tok;  cs = new List<Cmd>(); }
          Contract.Assert(cs != null);
          cs.Add(c);
-       } else {
+       } else if (label != null) {
          // LabelOrCmd read a label
-         Contract.Assert(label != null);
          if (startToken != null) {
            Contract.Assert(cs != null);
            // dump the built-up state into a BigBlock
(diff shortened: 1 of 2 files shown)


Source/Core/BoogiePL.atg
```diff
@@ -1362,11 +1361,10 @@ SpecBlock<out Block/*!*/ b>
...
   Ident<out x> ":"
   { LabelOrCmd<out c, out label>
-                       (. if (c != null) {
-                            Contract.Assert(label == null);
+                       (. Contract.Assert(c == null || label == null);
+                          if (c != null) {
                             cs.Add(c);
-                          } else {
-                            Contract.Assert(label != null);
+                          } else if (label != null) {
```
shazqadeer, line 1367: This test seems redundant because of the assertion above.

Source/Core/BoogiePL.atg
```diff
@@ -705,15 +705,14 @@ StmtList<out StmtList/*!*/ stmtList>
...
-    (. if (c != null) {
+    (. Contract.Assert(c == null || label == null);
+       if (c != null) {
          // LabelOrCmd read a Cmd
-         Contract.Assert(label == null);
          if (startToken == null) { startToken = c.tok;  cs = new List<Cmd>(); }
          Contract.Assert(cs != null);
          cs.Add(c);
-       } else {
+       } else if (label != null) {
```
shazqadeer, line 714: Same here.

Source/Core/BoogiePL.atg
bkragl: No. If `c == null` then the assertion does not tell us anything about `label`. It could be that `label == null` in which case we do not want to enter the branch.

shazqadeer (approved):

Issue #112, opened 2019-02-11
Error with recursive generic procedure


cburke496:
The following code:
```
type List T;
function tl<T>(List T): List T;

procedure Foo<T>(L: List T)
{
  call Foo(tl(L));
}
```
gives the error:
```
(5, 14): Error: more than one declaration of type variable: T
```
However, if I use a function instead of a procedure, it verifies just fine:
```
function FooF<T>(L: List T): bool
{
  FooF(tl(L))
} 
```

cburke496:
I just realized that you don't even need the `tl` function. Just recursing on `L` is enough to break it.

cburke496:
Turns out that separating the procedure into a pair of procedure and implementation that have different names for their type parameter is a workaround.

RustanLeino:
This is a bug.

PR #285, opened 2020-09-12
Fixes issue #112

shazqadeer:
This PR fixes two problems related to issue #112:
1. CheckBoundVariableOccurrences was being called during resolution of a procedure call.  A crash was happening in AddTypeBinder while preparing the state for the call to CheckBoundVariableOccurrences.  The call to CheckBoundVariableOccurrences seemed unnecessary to me since it checks whether each type parameter is used at least once in the type of some argument, a check that has already been performed at the declaration of the procedure.  So I removed the code that is setting up the call to CheckBoundVariableOccurrences.  This change is in AbsyCmd.cs.

2. During type checking of a recursive procedure call, when the substitution for the type parameters of the procedure is being calculated, there is a contract asserting that the free variables in the type of any actual are distinct from the type parameter of the procedure being called.  This assertion can be violated at a recursive polymorphic call since the type parameter in procedure and implementation are shared.  I solved this problem by eliminating the sharing and creating separate copies of the type parameter of the implementation in the parser (similar to the formals).  This change is BoogiePL.atg.

3. I also cleaned up the signature of CheckBoundVariableOccurrences whose return value is not being used.

This PR does not solve a related problem: Boogie crashes on a program in which a polymorphic recursive function calls itself as follows.  

```
type List T;
function {:inline} blah<T>(i: List T): List T {
  blah(i)
}
```

The problem here is the same as that described in 2 above.  But the fix is not the same since there is a single declaration for the function that comprises its signature and its body.  Any advice on fixing the recursive function issue would be much appreciated.

### DIFF
--- a/Source/Core/Parser.cs
+++ b/Source/Core/Parser.cs
@@ -20,6 +20,7 @@ using Bpl = Microsoft.Boogie;
 
 using System;
 using System.Diagnostics.Contracts;
+using System.Linq;
 
 namespace Microsoft.Boogie {
 
@@ -511,7 +512,7 @@ private class BvBounds : Expr {
 				Spec(pre, mods, post);
 			}
 			ImplBody(out locals, out stmtList);
-			impl = new Implementation(x, x.val, typeParams,
+			impl = new Implementation(x, x.val, new List<TypeVariable>(typeParams.Select(x => new TypeVariable(x.tok, x.Name))),
 			                         Formal.StripWhereClauses(ins), Formal.StripWhereClauses(outs), locals, stmtList, kv == null ? null : (QKeyValue)kv.Clone(), this.errors);
 			
 		} else SynErr(110);
--- a/Source/Core/Parser.cs
+++ b/Source/Core/Parser.cs
@@ -20,7 +20,6 @@ using Bpl = Microsoft.Boogie;
 
 using System;
 using System.Diagnostics.Contracts;
-using System.Linq;
 
 namespace Microsoft.Boogie {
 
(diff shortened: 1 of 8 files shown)


Source/Core/Parser.cs
```diff
@@ -511,7 +512,7 @@ public override int ComputeHashCode()
 				Spec(pre, mods, post);
 			}
 			ImplBody(out locals, out stmtList);
-			impl = new Implementation(x, x.val, typeParams,
+			impl = new Implementation(x, x.val, new List<TypeVariable>(typeParams.Select(x => new TypeVariable(x.tok, x.Name))),
```
RustanLeino, line 515: If you use `ConvertAll` instead of `Select`, you don't need `using System.Linq;`.

```suggestion
			impl = new Implementation(x, x.val, typeParams.ConvertAll(x => new TypeVariable(x.tok, x.Name)),
```

RustanLeino (changes requested):
Regarding the similar issue you found with `function`, it seems that `Function.CreateDefinitionAxiom` is already making copies of the the type variables (which happens before resolution of the `definition`). When the body is to be inlined (which I guess is happening after resolution), can the type variables in the definition be substituted then (that is, at the time the inlining happens)?

shazqadeer:
> Regarding the similar issue you found with `function`, it seems that `Function.CreateDefinitionAxiom` is already making copies of the the type variables (which happens before resolution of the `definition`). When the body is to be inlined (which I guess is happening after resolution), can the type variables in the definition be substituted then (that is, at the time the inlining happens)?

@RustanLeino : The crash in the case of  polymorphic function recursion is happening inside the type checker which happens before inlining.  Specifically, the crash happens at this line of code in MatchArgumentTypes:

Contract.Assert(Contract.ForAll(0, typeParams.Count, index => !actual.FreeVariables.Contains(typeParams[index])));

This is just FYI.  The overall point is that I didn't really understand your suggestion.

Source/Core/Parser.cs
shazqadeer: Done.

RustanLeino (approved):

Issue #280, opened 2020-09-01
[CIVL] Name clashes due to inlining of atomic actions


bkragl:
Code originating from an atomic action declaration is inlined in at least two places:

1. Refinement checking: assuming the gate and asserting the transition relation of a refined action.
2. Translating calls to actions: asserting the gate.

This can cause name clashes between bound variables coming from the action and local variables coming from the context. This is not a problem when running CIVL directly (because the variables are denoted by different objects in memory), but causes type checking errors in the generated desugared file.

The following program illustrates the problem.

```
procedure {:atomic}{:layer 1} action()
{
  assert (forall i:int :: true);
  assume (forall i:int :: true);
}

procedure {:yields}{:layer 0}{:refines "action"} P ()
{
  var i:int;
  // two clashes from refinement checking:
  // * when assuming the gate
  // * when asserting the transition relation
}

procedure {:yields}{:layer 1} Q ()
{
  var i:int;
  call P(); // clash when asserting the gate
}
```

PR #290, opened 2020-09-24
Allow bound variables to shadow similarly-named variables in outer scope

shazqadeer:
This PR changes the resolver so that bound variables in quantifiers are allowed to shadow locals in outer scope.  Note that shadowing of globals was already allowed in Boogie.  I have verified that the correct thing is happening in the VC generation by examining the SMT output manually for the program shown below. I couldn't come up with sensible regression tests of the sort common in the Boogie test folder where the golden output is indicated with verification/errors on procedures.  I would be glad to incorporate suggestions from the reviewers about possible regression tests. 

```
var addr: int;

procedure {:inline 1} blah(i: int) {
    assume addr == 64;
}

procedure foo(addr: int) returns (i: int)
requires addr == 46;
ensures addr == 84;
{
    call blah(addr);
    assume addr == i;
    assume (forall addr: int:: (exists addr: int :: addr == 0));
    assert (forall addr: int:: (exists addr: int :: addr == 0));
}

```

This PR should  fix issue #280 .

### DIFF
--- a/Source/Core/Absy.cs
+++ b/Source/Core/Absy.cs
@@ -2405,7 +2405,7 @@ namespace Microsoft.Boogie
     public override void Register(ResolutionContext rc)
     {
       //Contract.Requires(rc != null);
-      rc.AddVariable(this, true);
+      rc.AddVariable(this);
     }
 
     public override void Resolve(ResolutionContext rc)
@@ -2497,7 +2497,7 @@ namespace Microsoft.Boogie
     public override void Register(ResolutionContext rc)
     {
       //Contract.Requires(rc != null);
-      rc.AddVariable(this, true);
+      rc.AddVariable(this);
     }
 
     public override Absy StdDispatch(StandardVisitor visitor)
@@ -2535,7 +2535,7 @@ namespace Microsoft.Boogie
     public override void Register(ResolutionContext rc)
     {
       //Contract.Requires(rc != null);
-      rc.AddVariable(this, false);
+      rc.AddVariable(this);
     }
 
     /// <summary>
@@ -2591,7 +2591,7 @@ namespace Microsoft.Boogie
     public override void Register(ResolutionContext rc)
     {
       //Contract.Requires(rc != null);
-      rc.AddVariable(this, false);
+      rc.AddVariable(this);
     }
 
     public override Absy StdDispatch(StandardVisitor visitor)
@@ -2645,7 +2645,7 @@ namespace Microsoft.Boogie
     public override void Register(ResolutionContext rc)
     {
       //Contract.Requires(rc != null);
-      rc.AddVariable(this, false);
+      rc.AddVariable(this);
     }
 
     public override Absy StdDispatch(StandardVisitor visitor)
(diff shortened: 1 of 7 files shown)


RustanLeino:
How about the following tests:

```
function Inc2(x: int): int {
  (var x' := x + 4; x' - 2)
}

procedure {:inline} Inc3(x: int) returns (y: int) {
  y := (var y' := x + 3; y');
}

procedure P() {
  var x: int;
  var b: bool;

  x := 10;
  call x := Inc3(Inc2(x + 3));
  assert x == 13;
  
  b := (var y := x; (var y' := y + 1; (forall x': int :: x' == y' ==> x' == 14)));
  assert b;
  b := (var y := x; (forall x': int :: x' == y ==> (var x'' := x' + 20; x'' == 23)));
  assert b;
  b := (var y := x; (exists x': int :: x' < y && x == 28));
  assert !b;
}


function ShadowInc2(x: int): int {
  (var x := x + 4; x - 2)
}

procedure {:inline} ShadowInc3(x: int) returns (y: int)
{
  y := (var y := x + 3; y);
}

procedure ShadowP() {
  var x: int;
  var b: bool;

  x := 10;
  call x := ShadowInc3(ShadowInc2(x + 3));
  assert x == 13;
  
  b := (var y := x; (var y := y + 1; (forall x: int :: x == y ==> x == 14)));
  assert b;
  b := (var y := x; (forall x: int :: x == y ==> (var x := x + 20; x == 23)));
  assert b;
  b := (var y := x; (exists x: int :: x < y && x == 28));
  assert !b;
}
```

Routines `ShadowInc2`, `ShadowInc3`, and `ShadowP` are expected to behave like their non-shadowing counterparts `Inc2`, `Inc3`, and `P`.

I suggest you also add the example you gave above, and/or modify/extend mine to also include a global variable.

bkragl (approved):
Looks good to me.

I noticed a few typos in @RustanLeino's test.
* `Inc3`/`ShadowInc3` must be `{:inline 1}`.
* In `P`/`ShadowP`
  * `assert x == 13;` => `assert x == 18;`
  * `x == 14` => `x == 19` (first assignment to `b`)
  * `x == 23` => `x == 38` (second assignment to `b`)
  * I'm not sure about the significance of `x == 28` in the third assignment to `b`

How about the following test?
```
var x:int;

procedure P()
requires x == 0;
{
  assert (forall y:int :: x == 0); // x should be global variable
  assert (forall x:int :: x == 0); // x should be bound variable
}
```

RustanLeino:
Thanks. (blush) (I made some "improvements" in my example just before I posted it.)

The `x == 28` is also a typo in `P`. It should have said `x' == 28`. The number `28` is anything that's `18` or larger, so that the `exists` evaluates to `false`.

shazqadeer:
Thanks a lot, @bkragl & @RustanLeino. I have added the tests you suggested.

shazqadeer:
@bkragl : I will close #280 once this PR lands.  Please create a new issue if you want to pursue refactoring of atomic action gate  and transition relation into Boogie functions as we discussed.  It may be worthwhile to do so just to increase readability of the desugared code.

I recommend you catch a hold  of @RustanLeino some time and show him all the amazing things you have done in CIVL by standing atop Boogie's giant shoulders.

Issue #243, opened 2020-07-04
Inconsistency in solution: z3 formula vs Boogie VCExpr


stanlysamuel:
Hi all,

I use Boogie's Verification Condition engine as part of my research. I found a formula that z3 can solve but not Boogie. This seems very unintuitive to me, since Boogie invokes z3 under the hood.

Consider the formula (say VC) consisting of a free variable 's' and a bound variable 's@0' of Integer sort:

```
( Forall s@0::
    (And
        (Implies (== s 2)
            (Implies (== s@0 4) (!= s@0 5))
        )
        (Implies (!= s 2)
            (Implies (== s@0 s) (!= s@0 5))
        )
    )
)
```
It is easy to see that VC is equivalent to the formula below:

`(!= s 5)`

However Boogie is unable to prove this equivalence (due to the inability of being able to prove `(Implies VC (!= s 5))`) whereas z3 is able to prove it.

Here are the formulae in z3 and Boogie formats for reference.

**Z3:**

```
; Validity Check

(declare-const s Int)

(assert (not 
            (= 
              (not (= s 5))
                    
              (forall ((s0 Int)) 
                  (and  (=> (= s 2) 
                            (=> (= s0 4) (not (= s0 5)))
                        )
                         (=> (not (= s 2)) 
                              (=> (= s0 s) (not (= s0 5))) 
                         )                            
                    ) 
               )
          ) 
        ) 
      )
(check-sat)
```
**Boogie VCExpr:**

```
       // Added in BeginCheck() function in VC.cs
        VCExprVar s = exprGen.Variable("s", Microsoft.Boogie.Type.Int);
        VCExprVar s0 = exprGen.Variable("s@0", Microsoft.Boogie.Type.Int);
        VCExpr five = exprGen.Integer(BigNum.FromInt(5));
        VCExpr sne5 = exprGen.Neq(s,five);
        VCExpr s0ne5 = exprGen.Neq(s0,five);
        
        VCExpr four = exprGen.Integer(BigNum.FromInt(4));
        VCExpr two = exprGen.Integer(BigNum.FromInt(2));

        VCExpr s0s = exprGen.Eq(s0,s);
        VCExpr s04 = exprGen.Eq(s0,four);
        VCExpr s2 = exprGen.Eq(s,two);
        VCExpr ns2 = exprGen.Neq(s,two);

        VCExpr VC_ = exprGen.And(
        exprGen.Implies(s2,exprGen.Implies(s04,s0ne5)),
[...]

PR #291, opened 2020-09-25
Build and publish NuGet packages for all projects

keyboardDrummer:
**Context**

Boogie currently publishes a command line tool to NuGet. However, some Boogie consumers, such as  [Dafny](https://github.com/dafny-lang/dafny), want to consume Boogie as a library that provides a compiler API. Currently these consumers can only do this by consuming Boogie from source, which makes setting up a developer environment more difficult and making building the solution fragile and slow.

This PR configures all Boogie projects so that they build a NuGet package, and changes the .travis.yml file so that it publishes those NuGet packages. 

**Notes**
- Each .NET project has its own NuGet package. This is also how the Roslyn compiler API is vended, see [here](https://www.nuget.org/packages/Microsoft.CodeAnalysis) for more information.
- I haven't added package descriptions to the packages so they're published with an empty description. Would someone like to take a stab at adding these descriptions or suggesting them?
- I haven't configured a `PackageId` for the projects so this defaults to the assembly name. All the assembly names, such as BoogieExecutionEngine, start with Boogie, but more common seems to be to separate parts of the package id with dots instead of just CamelCasing. We could decide to format the package ids like this: `Boogie.ExecutionEngine`.
- For SMTLib the assembly name did not start with Boogie so a PackageId has been set to `BoogieSMTLib`
- If this has not already been done, I recommend registering Boogie as a package ID prefix under the Boogie NuGet account, as is described [here](https://docs.microsoft.com/en-us/nuget/nuget-org/id-prefix-reservation).


bkragl:
Good point. I was wondering if the existing NuGet package can be consumed as library. @keyboardDrummer, can you confirm this is impossible? If so, I agree that we should support it. However, I think I am against packaging every single project separately. This creates a lot of clutter and I think the project structure is not necessarily designed to provide separate libraries for clients outside Boogie. Just my two cents.
How about a single library package? Perhaps generated form the `ExecutionEngine` project?

keyboardDrummer:
> I was wondering if the existing NuGet package can be consumed as library. @keyboardDrummer, can you confirm this is impossible?

I can confirm it's not possible. I tried it and I got an error message about the NuGet package being a tool:

```
Invalid project-package combination for Boogie 2.7.31. DotnetToolReference project style can only contain references of the DotnetTool type 
Package 'Boogie 2.7.31' has a package type 'DotnetTool' that is not supported by project '/Users/rwillems/Documents/SourceCode/dafny/Source/DafnyDriver/DafnyDriver.csproj'.
```

> However, I think I am against packaging every single project separately. This creates a lot of clutter and I think the project structure is not necessarily designed to provide separate libraries for clients outside Boogie. Just my two cents.
How about a single library package? Perhaps generated form the ExecutionEngine project?

I wanted to do this initially for exactly the same reason you describe, to avoid clutter, but I changed my mind after reading more documentation. 

.NET by design does not support configuring a project to create a NuGet package that includes its dependencies. You can work around this by creating a .nuspec file and specifying the paths of the .dll files to include in the package, although I feel that is a somewhat fragile solution. From what I saw, Microsoft published NuGet packages don't specify a .nuspec file and stick to publishing each .NET project as a separate NuGet package.

From the consumption side, it doesn't matter whether all the .dll's are in one NuGet package or spread over multiple packages. If you reference only the root NuGet package the behavior is the same in both cases. For example, adding the single reference

```
  <ItemGroup>
    <PackageReference Include="Microsoft.CodeAnalysis" Version="3.8.0-3.final" />
  </ItemGroup>
```

will download a whole tree of NuGet packages but the consumer is agnostic to that. They simply see multiple .dll references show up in their project, as they would when referring to a NuGet package that includes multiple .dll files but has no dependencies.

> I think the project structure is not necessarily designed to provide separate libraries for clients outside Boogie.

Thinking about descriptions for each of the Boogie projects may show whether some of them can be merged. If none of these projects except `ExecutionEngine` are of any use on their own, why are they separate projects?

bkragl:
I did not say that none of the projects are of any use on their own. However, my understanding is that they are more of an internal structuring mechanism (e.g., some grew out of prototyping a research idea), rather than a careful partitioning in reusable components. See for example [here](https://github.com/boogie-org/boogie/issues/243#issuecomment-661531087), where @shazqadeer explains that `VCExpr`'s are a private detail of Boogie, not intended to be directly used by clients.

Of course, improving the project structure, finding descriptions, etc. is all great. But I think we need to answer two questions:

1) How clean/stable/independent do we want to have the individual projects in order to provide them as separate packages? I'd be reluctant to give a wrong impression. For example, there is a longstanding issue #2 regarding the use of Boogie as a library (which includes the dynamic loading of a prover DLL). Do we want to have this resolved first? I don't really know how much work this would be.

2) Do we generally want to give up the flexibility to do internal restructuring? Once a package is out, there is some commitment to keep it.

It would be great to get some more perspectives. @RustanLeino @shazqadeer @zvonimir @michael-emmi

michael-emmi:
I tend to think that separate packages are better.

 As @keyboardDrummer notes, the dependency mechanism is agnostic: one import of a root package brings the entire tree.

As far as polluting NuGet search results: seems like a non-issue, since I don’t think NuGet search for Boogie would be a common operation. I imagine most consumers simply type `dotnet install…`, never heading to `nuget.org`.

I buy @bkragl ’s objection that there are nuances like Issue #2 in the way packages are currently organized, but one would hope that such nuances would be few and resolvable without sacrificing modularity here. For now, some explanation of this caveat on the documentation should do just fine.

Furthermore, I can already imagine other uses for modular packages, e.g., using only the Boogie AST or AST and parser as libraries.

So I would definitely encourage the current modular structure if possible.

michael-emmi:
> I haven't configured a PackageId for the projects so this defaults to the assembly name. All the assembly names, such as BoogieExecutionEngine, start with Boogie, but more common seems to be to separate parts of the package id with dots instead of just CamelCasing. We could decide to format the package ids like this: Boogie.ExecutionEngine.

I would vote to go with convention here, i.e., `Boogie.ExecutionEngine` and the like.

michael-emmi:
> If this has not already been done, I recommend registering Boogie as a package ID prefix under the Boogie NuGet account, as is described here.

I am attempting to register the package ID prefix now.

bkragl:
> > I haven't configured a PackageId for the projects so this defaults to the assembly name. All the assembly names, such as BoogieExecutionEngine, start with Boogie, but more common seems to be to separate parts of the package id with dots instead of just CamelCasing. We could decide to format the package ids like this: Boogie.ExecutionEngine.
> 
> I would vote to go with convention here, i.e., `Boogie.ExecutionEngine` and the like.

Then I would suggest to change all `AssemblyName`'s accordingly. An change `Provers.SMTLib` to `Boogie.Provers.SMTLib`. I think then we do not need to set `PackageId` anywhere explicitly, right?

keyboardDrummer:
> I did not say that none of the projects are of any use on their own.

Sorry, I didn't mean that either. My phrasing was a bit coarse.

> Do we generally want to give up the flexibility to do internal restructuring? Once a package is out, there is some commitment to keep it.

If packages are merged or removed in the future then I think that would be fine. Consumers who consume ExecutionEngine would not be impacted. Packages that are removed can be marked as deprecated in NuGet to indicate they won't receive further updates.

> How clean/stable/independent do we want to have the individual projects in order to provide them as separate packages? I'd be reluctant to give a wrong impression. For example, there is a longstanding issue #2 regarding the use of Boogie as a library (which includes the dynamic loading of a prover DLL). Do we want to have this resolved first? I don't really know how much work this would be.

Dafny would be helped with NuGet packages of the Boogie projects in their current state. The issue you reference can be resolved in a later NuGet package update.

> Then I would suggest to change all AssemblyName's accordingly. An change Provers.SMTLib to Boogie.Provers.SMTLib. I think then we do not need to set PackageId anywhere explicitly, right?

Sounds great. I'll change that.

michael-emmi (approved):
Looks good to me, though I’ve never tried generating library packages, so I will assume that it works as advertised.

zvonimir:
I don't have a strong opinion on this. My main concern is how and whether this update would affect downstream tools such as Corral (https://github.com/boogie-org/corral/). So before merging this in, could someone make sure that it won't cause a major headache for Corral?

I guess at some point Corral should be switched to use Boogie package (instead of being built against Boogie source), but I don't know how to achieve that (and also don't have time to do it).

bkragl (changes requested):
@keyboardDrummer, could you unify all `PackageLicenseFile` and `GeneratePackageOnBuild` configurations into the file `Directory.Build.props`? Perhaps now this also makes sense for `Authors` and `RepositoryUrl`.

And for packing the license file you can write:
```
  <ItemGroup Condition="$(MSBuildProjectName.Contains('-NetCore'))">
    <None Include="$(MSBuildThisFileDirectory)..\LICENSE.txt" Pack="true" PackagePath="LICENSE.txt" />
  </ItemGroup>
```

I'm thinking that we might want to drop `GeneratePackageOnBuild` and explicitly call `dotnet pack` in CI, but we can still decide on that later.

In `.travis.yml` the logic for picking up the `nupkg` files for publishing must be updated to also look in subdirectories.

Finally, before this gets merged, I'd like to put up a PR to remove the `ParserHelper` project and potentially rename other projects.

RustanLeino:
I have no prior expertise in how these things should be done. What's been said seems reasonable and good to me. Dafny would benefit from being able to obtain Boogie via these nuget packages. As pointed out above, it would be good for someone to try out Corral with this PR before it's merged (though I'm guessing that as long as Corral doesn't change how it gets to Boogie, any update needed for Corral should be easy).

keyboardDrummer:
> @keyboardDrummer, could you unify all `PackageLicenseFile` and `GeneratePackageOnBuild` configurations into the file `Directory.Build.props`? Perhaps now this also makes sense for `Authors` and `RepositoryUrl`.
> 
> And for packing the license file you can write:
> 
> ```
>   <ItemGroup Condition="$(MSBuildProjectName.Contains('-NetCore'))">
>     <None Include="$(MSBuildThisFileDirectory)..\LICENSE.txt" Pack="true" PackagePath="LICENSE.txt" />
>   </ItemGroup>
> ```
> 
> I'm thinking that we might want to drop `GeneratePackageOnBuild` and explicitly call `dotnet pack` in CI, but we can still decide on that later.
> 
> In `.travis.yml` the logic for picking up the `nupkg` files for publishing must be updated to also look in subdirectories.
> 
> Finally, before this gets merged, I'd like to put up a PR to remove the `ParserHelper` project and potentially rename other projects.

Sounds good. I've made the changes you suggested.

keyboardDrummer:
> Finally, before this gets merged, I'd like to put up a PR to remove the ParserHelper project and potentially rename other projects.

Let me know if you need help with that.

bkragl:
@keyboardDrummer Could you rebase this PR on the current master branch, where the `ParserHelper` project was removed?

I think the expression for publishing the packages still has a problem. In my shell (Bash) the `**` wildcard does not pick up projects in subdirectories (in particular `Provers/SMTLib`). One solution would be to set `shopt -s globstar` somewhere. Not sure if there is a better way.

And please revert all whitespaces changes (dropped empty lines, ...) from the diff.

Thanks!

shazqadeer:
@keyboardDrummer :  It looks like all discussion has been done and only a few changes are required.  Do you think you could do the necessary changes in the next week and land this PR?

keyboardDrummer:
> I think the expression for publishing the packages still has a problem.

Should be good now. It also finds SMTLib

> And please revert all whitespace changes (dropped empty lines, ...) from the diff.

Done

shazqadeer:
@bkragl : Are you satisfied?

bkragl:
> > I think the expression for publishing the packages still has a problem.
> 
> Should be good now. It also finds SMTLib

On which system did you try it? I still see the following locally.

```
> ls Source/Core/bin/Debug/Boogie.Core.1.0.0.nupkg 
Source/Core/bin/Debug/Boogie.Core.1.0.0.nupkg
> ls "Source/**/bin/Debug/Boogie*.nupkg"
ls: cannot access 'Source/**/bin/Debug/Boogie*.nupkg': No such file or directory
```

But maybe this is just my shell. We can try and see if it works in Travis. Another option would be to write ``dotnet nuget push `find Source -name "*.nupkg"` ...``. If I'm not mistaken, the Debug and Release build are completely separate, so when the Release build runs this, it should not pick up packages generated by the Debug build. Right?

> Note the root library is still called Boogie.ExecutionEngine. I think that's fine. I think it can be changed later. There won't be too many clients and if there's a rename they will simply have to depend on the new name of the root package.

Yes, I completely agree.

bkragl:
> > If this has not already been done, I recommend registering Boogie as a package ID prefix under the Boogie NuGet account, as is described here.
> 
> I am attempting to register the package ID prefix now.

@michael-emmi Did the registration go through?

keyboardDrummer:
> On which system did you try it? I still see the following locally.

It might be best to test with `dotnet nuget push`.

I see for example (Added `Provers.` so it immediately finds that one):

```
dotnet nuget push "Source/**/bin/DEBUG/Boogie.Provers.*.nupkg" -s https://api.nuget.org/v3/index.json
warn : No API Key was provided and no API Key could be found for 'https://www.nuget.org/api/v2/package'. To save an API Key for a source use the 'setApiKey' command.
Pushing Boogie.Provers.SMTLib.2.7.31-buildpackages0001.nupkg to 'https://www.nuget.org/api/v2/package'...
  PUT https://www.nuget.org/api/v2/package/
  Unauthorized https://www.nuget.org/api/v2/package/ 925ms
```

bkragl (approved):
Looks great now. Thanks @keyboardDrummer for putting this together! I think this will really simplify dependency management for Boogie clients.

bkragl:
The deployment failed.

```
Pushing Boogie.2.7.38.nupkg to 'https://www.nuget.org/api/v2/package'...
  PUT https://www.nuget.org/api/v2/package/
  Created https://www.nuget.org/api/v2/package/ 855ms
Your package was pushed.
Pushing Boogie.VCExpr.2.7.38.nupkg to 'https://www.nuget.org/api/v2/package'...
  PUT https://www.nuget.org/api/v2/package/
  Forbidden https://www.nuget.org/api/v2/package/ 137ms
error: Response status code does not indicate success: 403 (The specified API key is invalid, has expired, or does not have permission to access the specified package.).
```

I don't know exactly how nuget.org works, but it looks like the new packages first need to be created over there. @michael-emmi, do you know what to do? I see a "boogie-org" account on nuget.org. Is this in any way connected to the "boogie-org" organization on GitHub?

keyboardDrummer:
> I don't know exactly how nuget.org works, but it looks like the new packages first need to be created over there.

It depends on the API key you're using. The key configuration determines whether to accept new packages, and has a glob pattern to determine which packages it accepts. You should log into the account Boogie uses for NuGet, and verify the key accepts the new packages.

michael-emmi:
@bkragl @keyboardDrummer the existing API key that Travis was using only allowed publishing new versions of existing packages. I have just created a new API key that allows publishing of new packages too, and replaced existing key in Travis. So publishing new packages ought to work now.

Shall we switch back to the old API key once these new packages are published for the first time?

bkragl:
The new packages are out now.

Just re-running the build did not work, because the previously existing Boogie package was actually published before and submitting it again caused an error. So I created a dummy commit.

> Shall we switch back to the old API key once these new packages are published for the first time?

As you wish. If we do, I'll try to remember in case we decide to change the package structure in the future.

keyboardDrummer:
> Looks great now. Thanks @keyboardDrummer for putting this together! I think this will really simplify dependency management for Boogie clients.

Thank you as well :) and @michael-emmi

PR #292, opened 2020-09-26
Debug printing of strings and values

shazqadeer:
This PR introduces the augmented trace feature which allows printing of arbitrary strings and variables via attributes on assignment, assert, and assume statements.  The feature is enabled if the special attribute "debug" is used  as in the procedure below and the command-line option /enhancedErrorMessages:1 is used.  As you may have guessed, the implementation of this feature requires examination of the model returned by the solver.  

```
procedure foo(i: int) returns (o: int)
requires i == 42;
ensures o == i;
{
    o := i;
    if (*) {
        assume {:debug "line 25"} {:debug "o = ", o} true;
        o := o + 1;
    }
    if (*) {
        assume {:debug "line 29"} {:debug "o = ", o} true;
        o := {:debug "just to show off"} o + 1;
    }
    assert {:debug "exit of foo"} {:debug "o = ", o} true;
}

```
For the above program, the augmented trace part of the output is as follows:
```
Augmented execution trace:
line 25
o = 42
line 29
o = 43
just to show off
exit of foo
o = 44
```
 
I am looking for feedback, especially if you have concerns about how this feature is exposed to the user. I have not documented the "debug" attribute yet; I will do it once the design is finalized based on your feedback.

### DIFF
--- a/Source/Core/CommandLineOptions.cs
+++ b/Source/Core/CommandLineOptions.cs
@@ -1970,6 +1970,16 @@ namespace Microsoft.Boogie
        Prints <string> rather than the standard message for assertion failure.
        Also applicable to requires and ensures declarations. 
 
+---- On statements --------------------------------------------------
+
+     {:print e0, e1, e2, ...}
+       Indicates that expressions e0, e1, e2, ... must be printed.  Must be used 
+       in conjunction with /enhancedErrorMessages:n command-line option.  
+
+     {:captureState s}
+       Indicates that state must be captured under the name s.  Must be used
+       together with /mv:<string> command-line option.
+
   ---- The end ---------------------------------------------------------------
 ");
     }
--- a/Source/Core/Inline.cs
+++ b/Source/Core/Inline.cs
@@ -851,7 +851,13 @@ namespace Microsoft.Boogie
         {
           return cmd;
         }
-        return BoundVarAndReplacingOldSubstituter.Apply(substMap, oldSubstMap, prefix, cmd);
+        var newCmd = BoundVarAndReplacingOldSubstituter.Apply(substMap, oldSubstMap, prefix, cmd);
+        if (cmd is ICarriesAttributes attrCmd && attrCmd.Attributes != null)
+        {
+          var attrCopy = (QKeyValue) attrCmd.Attributes.Clone();
+          ((ICarriesAttributes) newCmd).Attributes = Substituter.ApplyReplacingOldExprs(Subst, OldSubst, attrCopy);
+        }
+        return newCmd;
       }
 
       public Expr CopyExpr(Expr expr)
--- a/Source/ExecutionEngine/ExecutionEngine.cs
+++ b/Source/ExecutionEngine/ExecutionEngine.cs
@@ -1798,19 +1798,18 @@ namespace Microsoft.Boogie
             WriteErrorInformationToXmlSink(errorInfo, error.Trace);
           }
 
-          if (CommandLineOptions.Clo.EnhancedErrorMessages == 1)
-          {
-            foreach (string info in error.relatedInformation)
-            {
-              Contract.Assert(info != null);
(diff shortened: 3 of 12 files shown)


wrwg:
> This PR introduces the augmented trace feature which allows printing of arbitrary strings and variables via attributes on assignment, assert, and assume statements. The feature is enabled if the special attribute "debug" is used as in the procedure below and the command-line option /enhancedErrorMessages:1 is used. As you may have guessed, the implementation of this feature requires examination of the model returned by the solver.
> 
> ```
> procedure foo(i: int) returns (o: int)
> requires i == 42;
> ensures o == i;
> {
>     o := i;
>     if (*) {
>         assume {:debug "line 25"} {:debug "o = ", o} true;
>         o := o + 1;
>     }
>     if (*) {
>         assume {:debug "line 29"} {:debug "o = ", o} true;
>         o := {:debug "just to show off"} o + 1;
>     }
>     assert {:debug "exit of foo"} {:debug "o = ", o} true;
> }
> ```
> 
> For the above program, the augmented trace part of the output is as follows:
> 
> ```
> Augmented execution trace:
> line 25
> o = 42
> line 29
> o = 43
> just to show off
> exit of foo
> o = 44
> ```
> 
> I am looking for feedback, especially if you have concerns about how this feature is exposed to the user. I have not documented the "debug" attribute yet; I will do it once the design is finalized based on your feedback.

This should work well for our use cases in the Move Prover. Thanks for making this happen!

RustanLeino:
Cool feature. Can it be combined with the `{:CaptureState}` attribute (see `Test/test15/CaptureState.bpl`)?

Also, please describe the feature in the `/attrHelp` message (which I'm seeing that `{:CaptureState}` should have been, too).

shazqadeer:
@RustanLeino : I wasn't aware of the CaptureState feature when I was working on this PR.  But I was reminded of it because of a comment on this other issue #293 I had created. Could you please chime in there also?

shazqadeer:
@RustanLeino: Regarding this PR, I will go ahead and document both :debug and :captureState and land.   As I explained in my comment on #293, although there are common aspects between these two features, there are significant differences between how the functionality is exposed.   It would be great to unify but I don't quite understand how just yet.

zvonimir:
Thanks for implementing this!

What is the semantics of `{:debug "o = ", o}`?
I would guess it takes the value of `o` and appends it to the end of the provided string. Is that right?

Also, is this going to work with inlining?

shazqadeer:
The semantics of {:debug a, b, c, ...} is that each item is printed out literally unless it is the name of a variable (represented as an IdentifierExpr inside Boogie) in which case the value of the variable is looked up at that control point in the error trace and printed out.  A new line is added after printing out all of the items inside one debug attribute.  Therefore, the printing of different debug attributes is  separated by a newlines.

RustanLeino:
> I will go ahead and document both :debug and :captureState and land. As I explained in my comment on #293, although there are common aspects between these two features, there are significant differences between how the functionality is exposed. It would be great to unify but I don't quite understand how just yet.

Sounds good. I like the "println" feel of the `:debug` attribute (in fact should `:debug` perhaps be renamed to something like `:debugprint` or `:modelprint` or just `:print`?). This seems useful if you know what to look for when the Boogie program is being generated. Since `:CaptureState` ultimately causes all variables to be printed (and to be printed as a succession of program states, along the trace that leads to the error), which is more useful for an tool that performs some post-processing of this debug information.

zvonimir (approved):
Looks good to me.

shazqadeer:
@RustanLeino: I have renamed :debug to :print.

shazqadeer:
@zvonimir: 

> Also, is this going to work with inlining?

Yes, the feature works with inlining also.  In the test file that I updated, I have examples with and without inlining.

shazqadeer:
I have added documentation for :captureState and :print.

PR #294, opened 2020-09-29
Small refactor in visitors

shazqadeer:
This PR replaces calls to Visit with calls to VisitExpr in the visitor of AssignCmd.  This refactoring ensures that all processing of expressions descends through VisitExpr which can be overridden to do something specific.


shazqadeer:
inductive-sequentialization/is.sh is failing in CI but not on my machine.  Test flakiness?

bkragl (approved):
Looks good to me.

PR #295, opened 2020-09-29
Use inclusive language

RustanLeino:


Source/Houdini/Houdini.cs
```diff
@@ -1046,7 +1046,7 @@ public HoudiniState(WorkQueue workQueue, Dictionary<Variable, bool> currentAssig
         this._assignment = currentAssignment;
         this._implementation = null;
         this._outcome = new HoudiniOutcome();
-        this.blackList = new HashSet<string>();
+        this.denylist = new HashSet<string>();
```
zvonimir, line 1049: This should probably be `denyList`. Otherwise, looks good!

zvonimir (changes requested):

zvonimir (approved):
Looks good to me.

PR #296, opened 2020-09-30
Merging and renaming projects

bkragl:
This PR eliminates the `ParserHelper` project by moving its content back to `Core` (the code for `Token` is now in a separate file).

Since PR #291 aims to publish all Boogie projects as separate packages, I'd like to discuss a renaming of the `ExecutionEngine` project to something more descriptive. `ExecutionEngine` would be the main package that clients that use (all of) Boogie as library would reference. So maybe `BoogieLib`or `Library`? Just `Boogie` (such that the package name is `Boogie.Boogie`) is a bit odd. But the class itself which clients call into could be just `Boogie` instead of `ExecutionEngine`.
Also, `AbsInt` could be expanded to `AbstractInterpretation`.

I'm happy to hear other suggestions. And if you think such a renaming causes more trouble for existing clients that what it helps, we can also stick with the existing names.
@RustanLeino @shazqadeer @zvonimir @michael-emmi @keyboardDrummer


shazqadeer:
I like the merging of ParserHelper into Core.  I always found that separation odd.

I am fine with renaming ExecutionEngine to BoogieLib and AbsInt to AbstractInterpretation.  But I don't have a strong opinion.

PR #297, opened 2020-10-01
Monomorphize subset of Boogie programs

shazqadeer:
This PR contains an initial attempt to monomorphize a subset of polymorphic Boogie. This subset excludes polymorphic  procedures, implementations, and maps but includes polymorphic functions and datatypes.  The subset allows only universal quantified type parameters at the outermost level in axioms.  The motivation behind this work is to build a reusable library of useful function definitions and axioms (see examples in the new tests).

The monomorphization flow is as follows:
- If the program is monomorphic, do nothing.
- Otherwise, check if the program meets the restrictions under which the current approach is  guaranteed to produce a monomorphic program.  If not, do nothing.
- Otherwise, monormophize as explained below.

Monomorphization approach:
- Walk over all procedures, implementations, and monomorphic axioms recursively discovering invocations of polymorphic functions and uses of polymorphic datatypes and creating monomorphic instantiations.  If a function call is discovered, it is also walked over recursively.
- Each axiom with universal type parameters is allowed to be annotated with the name of a type constructor whose arity should match the number of type parameters of the axiom.  This axiom will be instantiated with the arguments of every type constructed with that type constructor discovered during step 1.
- Axiom instantiation can discover other type terms which feed into axiom instantiation until a fixpoint is reached. 

To accomplish the aforementioned main objective of this PR, the implementation of datatypes is also refactored.  The main changes are as  follows:
- Allow datatypes to be polymorphic.
- Add DatatypeTypeCtorDecl which extends TypeCtorDecl and contains all the constructors inside it.  Constructors, selectors, and membership functions are no long put into TopLevelDeclarations; rather they are nested inside their instance of DatatypeTypeCtorDecl.

### DIFF
--- a/Source/Concurrency/CivlTypeChecker.cs
+++ b/Source/Concurrency/CivlTypeChecker.cs
@@ -871,19 +871,19 @@ namespace Microsoft.Boogie
 
     private void TypeCheckPendingAsyncMachinery()
     {
-      foreach (var typeCtorDecl in program.TopLevelDeclarations.OfType<TypeCtorDecl>())
+      foreach (var datatypeTypeCtorDecl in program.TopLevelDeclarations.OfType<DatatypeTypeCtorDecl>())
       {
-        if (typeCtorDecl.HasAttribute(CivlAttributes.PENDING_ASYNC))
+        if (datatypeTypeCtorDecl.HasAttribute(CivlAttributes.PENDING_ASYNC))
         {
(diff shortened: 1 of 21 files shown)


bkragl (approved):
The new way of processing datatypes looks good to me. I did not check the actual monomorphization in detail, but it is properly put behind a command-line option and marked as experimental.

PR #299, opened 2020-10-11
add capability to include a library file

shazqadeer:
This PR introduces the /lib option to included a library of definitions automatically with the user-supplied Boogie files.  The library definitions are embedded a resource in the Core project and parsed if /lib is supplied on the command line.

### DIFF
--- a/Source/Core/BoogiePL.atg
+++ b/Source/Core/BoogiePL.atg
@@ -29,13 +29,13 @@ readonly StructuredCmd/*!*/ dummyStructuredCmd;
 
 public static int ParseLibraryDefinitions(out Program program)
 {
-	string libraryDefinitionsFileName = "LibraryDefinitions.bpl";
-	Assembly asm = Assembly.GetExecutingAssembly();
-	var resourceName = asm.GetManifestResourceNames().First(s => s.EndsWith(libraryDefinitionsFileName, StringComparison.CurrentCultureIgnoreCase));
-	using (Stream resourceStream = asm.GetManifestResourceStream(resourceName))
-	{
-		return Parse(new StreamReader(resourceStream), libraryDefinitionsFileName, new List<string>(), out program);
-	}
+  string libraryDefinitionsFileName = "LibraryDefinitions.bpl";
+  Assembly asm = Assembly.GetExecutingAssembly();
+  var resourceName = "Core-NetCore.LibraryDefinitions.bpl";
+  using (Stream resourceStream = asm.GetManifestResourceStream(resourceName))
+  {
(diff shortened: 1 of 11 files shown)


bkragl (approved):
Looks good. Some minor suggestions below.

Source/Core/BoogiePL.atg
```diff
@@ -27,11 +27,22 @@ readonly List<Expr>/*!*/ dummyExprSeq;
 readonly TransferCmd/*!*/ dummyTransferCmd;
 readonly StructuredCmd/*!*/ dummyStructuredCmd;
 
+public static int ParseLibraryDefinitions(out Program program)
+{
+	string libraryDefinitionsFileName = "LibraryDefinitions.bpl";
+	Assembly asm = Assembly.GetExecutingAssembly();
+	var resourceName = asm.GetManifestResourceNames().First(s => s.EndsWith(libraryDefinitionsFileName, StringComparison.CurrentCultureIgnoreCase));
```
bkragl, line 34: I would remove this search and just directly use the resource name in `GetManifestResourceStream` below. On my machine the name is `Core-NetCore.LibraryDefinitions.bpl`. Is it the same for you? I don't quite get why it is not `Boogie.Core.LibraryDefinitions.bpl`, because I thought the prefix is the assembly name. But somehow the prefix is taken from the name of the `csproj` file.

If you do this change, you can also drop `using System.Linq;` from `Parser.frame`.

And the whole method `ParseLibraryDefinitions` uses tabs instead of spaces for indentation.

Source/Core/CommandLineOptions.cs
```diff
@@ -2012,6 +2022,7 @@ public override void Usage()
   Multiple .bpl files supplied on the command line are concatenated into one
   Boogie program.
 
+  /lib           : Include definitions in LibraryDefinitions.bpl
```
bkragl, line 2025: I would not mention the file name `LibraryDefinitions.bpl` here, because it is something internal.

Test/monomorphize/monomorphize1.bpl
```diff
@@ -1,34 +1,13 @@
 
-// RUN: %boogie /monomorphize /useArrayTheory "%s" > "%t"
+// RUN: %boogie /lib /monomorphize /useArrayTheory "%s" > "%t"
```
bkragl, line 2: I think the Boogie test suite generally uses the `-` version of command-line options (instead of `/`).

bkragl (approved):
Somehow the inline comments did not make it in my previous review.

shazqadeer:
I have addressed all the feedback.

PR #300, opened 2020-10-14
[CIVL] Refactor monomorphization

shazqadeer:
This PR refactors the monomorphization code to make it available as a library to CIVL.  All the CIVL regression tests are now ported to use both the /lib and /monomorphize feature.  Consequently, many redundant declarations in the examples are eliminated and the code of linear type checker is simplified.


shazqadeer:
@bkragl : I have incorporated the suggestion you made this morning about splitting out the visitor part of the Monomorphizer from the public API.

Test/civl/inductive-sequentialization/paxos/Paxos.bpl
```diff
@@ -48,31 +48,13 @@ function {:pending_async "A_Propose"}{:constructor} Propose_PA(round: Round, ps:
...
-function {:builtin "MapConst"} MapConstVoteResponse(int): [VoteResponse]int;
-function {:builtin "MapAdd"} MapAddVoteResponse([VoteResponse]int, [VoteResponse]int): [VoteResponse]int;
-function {:builtin "MapSub"} MapSubVoteResponse([VoteResponse]int, [VoteResponse]int): [VoteResponse]int;
-
 ////////////////////////////////////////////////////////////////////////////////
 //// Functions
 
-function {:inline} NoPAs(): [PA]int { MapConstPA(0) }
+function {:inline} NoPAs(): [PA]int { MapConst(0) }
 function {:inline} SingletonPA(pa:PA): [PA]int { NoPAs()[pa := 1] }
```
bkragl, line 55: In addition to singleton sets (currently called `MapUnit`), it would be nice to introducing polymorphic functions for empty sets as well as empty and singleton multisets. Then it would be possible to eliminate quite some more boilerplate definitions. But this can also be done in another PR.

bkragl (approved):
This feature is great, cleaning up both CIVL itself as well as CIVL programs. The API for monomorphization also looks good.

Test/civl/inductive-sequentialization/paxos/Paxos.bpl
shazqadeer: We already have the polymorphic function for "empty set" (MapConst(false)) and "empty multiset" MapConst(0). I hesitate to call these things sets and multisets because they are allowed to be infinite.  I was hoping to reserve the words set and multiset for the finite versions that will be introduced later.  

For singleton, would it be fine to introduce the following generic function:

function {:inline} MapAllExcept<T,U>(t: T, u: U, default: U): [T]U {
  MapConst(default)[t := u]
}

Issue #293, opened 2020-09-26
ModelViewer and BVD projects


shazqadeer:
@RustanLeino, @gauravpartha, @michalmoskal, @shuvendu-lahiri:

There are two projects, ModelViewer and BVD, that are currently not built with Boogie-Netcore.sln.  The code in these projects is relevant to visualization of models (returned by the SMT solver and processed into an instance of Microsoft.Boogie.Model) from the lens of the client language that compiles to Boogie for verification.  I would like to delete this code from the Boogie repository because (1) it is not being maintained, and (2) it appears to belong clients of Boogie rather than Boogie itself.  Any objections?

gauravpartha:
Fine from the Viper side, thanks for asking. We use the "/mv" option to obtain models from Boogie, but we don't seem to make use of the ModelViewer/BVD projects.

mmoskal:
I surely have no strong opinions here. Thanks for checking though!

shazqadeer:
Adding @zvonimir and @bkragl for their opinions also.

shazqadeer:
@gauravpartha: I am glad that you responded and mentioned that you use /mv feature in Boogie.  I was not aware of any users.  I have put up PR #292  for an augmented trace feature that outputs data values.  I am wondering if there is overlap in the functionality of my PR and the /mv feature.  I have never used the /mv feature myself.

shazqadeer:
To educate myself, I looked at the implementation of the ModelViewInfo feature in Boogie.  There is definitely overlap between this feature and PR #292 in the part related to examination of data values.  The main difference is that ModelViewInfo feature aims to provide the user with the entire state visible in the Boogie procedure at the state capture point (indicated by :captureState), whereas PR #292 aims to print only specific variables indicated by the user at the capture point (indicated by :debug).  PR #292 also has an explicit goal of capturing custom error traces that have source information from the program that was compiled to Boogie.  I suppose this goal could be met indirectly by naming the captured states in ModelViewInfo feature appropriately.

The implementation of ModelViewInfo feature is not robust.  It depends on the assumption that the solver is parsimonious in bindings of the model for an uninterpreted function, i.e., it does not provide bindings that are unnecessary to justify the "path" taken in the Boogie program.   It is possible to fix the robustness issue.  But I am curious to know if the ModelViewInfo feature is being used seriously.  Would the functionality provided by PR #292 be adequate instead?

gauravpartha:
[...]

PR #304, opened 2020-10-19
Remove BVD, ModelViewer, and Provers/TPTP

shazqadeer:
This PR fixes issue #293.  It also  removes unused arguments to /printModel:n. Now n can be either  0 or 1.


bkragl (approved):
How about making `-printModel` a Boolean flag?

shazqadeer:
> How about making -printModel a Boolean flag?

I prefer to leave it an integer in case anybody wishes to extend it in the future.  I suspect extensions were planned earlier but never executed.

PR #311, opened 2020-11-11
CIVL cleanup: simplified Absy construction and consistent use of AssertCmd

bkragl:
This PR cleans up the construction of `Absy` objects throughout the CIVL codebaes using helper functions. Now we don't have to pass `Token.NoToken` everywhere, and the construction of certain objects like `AssertCmd`, `CallCmd`, `Block`, etc. is further simplified.

Also, every CIVL check is now formulated as `AssertCmd` instead of `Ensures` (of course except the checks that actually come from postconditions in the input program).


shazqadeer (approved):
Great!

PR #312, opened 2020-12-01
CIVL: better refinement error messages

bkragl:


Source/Concurrency/RefinementInstrumentation.cs
```diff
@@ -173,14 +173,14 @@ public override List<Cmd> CreateAssertCmds()
       var skipOrTransitionRelationAssertCmd = CmdHelper.AssertCmd(
         tok,
         Expr.Or(Expr.Ident(pc), Expr.Or(OldEqualityExprForGlobals(), transitionRelation)),
-        "Transition invariant violated in initial state");
+        "Some yield-to-yield fragment modifies state in a way that does not match the refined atomic action");
```
shazqadeer, line 176: Did you consider "a" or "this" instead of "some" in the new error message?

shazqadeer: Another suggestion:

"Some yield-to-yield fragment modifies state without matching the refined atomic action"

Source/Concurrency/RefinementInstrumentation.cs
```diff
@@ -173,14 +173,14 @@ public override List<Cmd> CreateAssertCmds()
...
       CivlUtil.ResolveAndTypecheck(skipOrTransitionRelationAssertCmd);
 
       // assert pc ==> g_old == g && o_old == o;
       AssertCmd skipAssertCmd = CmdHelper.AssertCmd(
         tok,
         Expr.Imp(Expr.Ident(pc), Expr.And(OldEqualityExprForGlobals(), OldEqualityExprForOutputs())),
-        "Transition invariant violated in final state");
+        "Some yield-to-yield fragment modifies state subsequent to a yield-to-yield fragment that already modified state");
```
shazqadeer, line 183: Again "Some" -> "a"  or "this"?

Source/Concurrency/RefinementInstrumentation.cs
```diff
@@ -224,7 +224,7 @@ public override List<Cmd> CreateUpdatesToRefinementVars(bool isMarkedCall)
       {
         // assert !pc;
         // pc, ok := true, true;
-        cmds.Add(CmdHelper.AssertCmd(tok, Expr.Not(Expr.Ident(pc)), "Visible state changed before marked call"));
+        cmds.Add(CmdHelper.AssertCmd(tok, Expr.Not(Expr.Ident(pc)), "State modified before marked call"));
```
shazqadeer, line 227: Why not "visible state"?

Source/Concurrency/RefinementInstrumentation.cs
bkragl: I actually had "this", but then I realized that it does not really make sense because the token points to the procedure declaration. I actually would not know which token to choose to identify a yield-to-yield fragment, since they are dynamically scoped.

I like your suggestion, and we can go with "a" if you prefer.

Source/Concurrency/RefinementInstrumentation.cs
bkragl: It sounded mysterious to me. You have to understand which part of the state is relevant for refinement checking, and if you don't, I think that "visible" won't help much.

Source/Concurrency/RefinementInstrumentation.cs
shazqadeer: ok. "a" is better.

Source/Concurrency/RefinementInstrumentation.cs
shazqadeer: Yes, it is mysterious along with myriad of other CIVL concepts. My expectation is that the user of CIVL has read some documentation or is willing to read documentation to understand the meaning of error messages.  

Perhaps, "Either globals or outputs modified before marked call".

shazqadeer (approved):

Source/Concurrency/RefinementInstrumentation.cs
bkragl: Done.

Source/Concurrency/RefinementInstrumentation.cs
bkragl: Done.

Source/Concurrency/RefinementInstrumentation.cs
bkragl: As discussed, we will call it "layer-n state", where n is the next layer above the disappearing layer of the yielding procedure.

shazqadeer (approved):

PR #314, opened 2020-12-02
[CIVL] renamed non-blocking -> cooperation, terminates -> cooperates

shazqadeer:


Source/Core/CommandLineOptions.cs
```diff
@@ -1934,8 +1934,8 @@ Yield invariant.
      {:refines ""action""}
        Refined atomic action of a yielding procedure.
 
-     {:terminates}
-       Terminating loop.
+     {:cooperates}
+       Cooperating loop.
```
bkragl, line 1938: Explanation was incomplete:
`Cooperating loop or mover procedure.`

bkragl (approved):

PR #316, opened 2020-12-12
Improve command-line interface

bkragl:
Some modifications that make Boogie's command-line interface more "standard".

**Main change:**
On the one hand, `-version` is added to print the Boogie version number and terminate (until now it was only possible to get the version number by actually running Boogie on a valid `.bpl` file). And on the other hand, `-nologo` is removed and version number is not printed in the regular output.

In general, Boogie will only print a message and terminate immediately if `-version`, `-help`, `-attrHelp`, or `-proverHelp` is supplied.


zvonimir:
Nice cleanup! Question - what is the difference between `-version` and `/version`?

On my linux machine (Boogie installed using dotnet) only `/version` works. But then I see some regressions that it seems are successfuly using `-version`. I am running Boogie 2.7.30 and so maybe this got updated later on?

bkragl:
> Nice cleanup! Question - what is the difference between `-version` and `/version`?
> 
> On my linux machine (Boogie installed using dotnet) only `/version` works. But then I see some regressions that it seems are successfuly using `-version`. I am running Boogie 2.7.30 and so maybe this got updated later on?

If I understand correctly, you are not asking about the `version` flag added in this PR, but all Boogie flags in general, right?

Boogie allows both `/` and `-` interchangeably to indicate a command-line flag (I guess because the former is common on Windows and the latter on Linux). It would be odd if you get a different behavior.

The regressions consistently use the `-` variant.

shazqadeer (approved):

zvonimir:
Yes, I was asking in general. What you wrote is what I would expect, but on my Linux machine for some strange reason I get this when I run it with `-version`:
```
zvonimir@betina:~$ boogie -version
Boogie: Error: unknown switch: -version
Use /help for available options
```
Boogie version 2.7.30 is installed with dotnet as a package. Running it with `/version` works as one would expect.
No big deal, but it is for sure an interesting discrepancy.

bkragl:
Ah, I know what happens. This is what you must see:

```
> boogie /version
Boogie program verifier version 2.7.30.0, Copyright (c) 2003-2014, Microsoft.
*** Error: '/version': Filename extension '' is not supported. Input files must be BoogiePL programs (.bpl).
```

In the output, the first line is the "logo" header (which was previously printed on every Boogie run, could be suppressed with `-nologo`, and is now gone completely). Then `/version` is interpreted as a path and the second line in the output is an error message complaining that it is not a `.bpl` file.

So the discrepancy between `/` and `-` options is that if `something` is not a know option name, then Boogie tries to interpret `/something` as a file path, whereas it rejects `-something`.

zvonimir:
Sorry, but I don't quite get it. This is what I see:
```
zvonimir@betina:~$ boogie -version
Boogie: Error: unknown switch: -version
Use /help for available options
```
Shouldn't `-version` be a known option name?
I tried also `boogie -version test.bpl` and I get the same error.

bkragl:
> Shouldn't `-version` be a known option name?

It *was* not previously. It *is* as of this PR.

zvonimir:
Got it! All good now. Thanks!

PR #319, opened 2020-12-20
Update GitVersionTask to latest version to resolve issue requiring MS…

keyboardDrummer:
Update GitVersionTask to latest version to resolve issue requiring MSBUILDSINGLELOADCONTEXT workaround

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,9 +8,9 @@ git:
   depth: false
 env:
   global:
-    - BOOGIE_GITVERSION=1
     - SOLUTION=Source/Boogie.sln
     - Z3URL=https://github.com/Z3Prover/z3/releases/download/z3-4.8.8/z3-4.8.8-x64-ubuntu-16.04.zip
+    - BOOGIE_GITVERSION=1
   jobs:
     - CONFIGURATION=Debug
     - CONFIGURATION=Release
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ git:
   depth: false
 env:
   global:
+    - BOOGIE_GITVERSION=1
     - SOLUTION=Source/Boogie.sln
     - Z3URL=https://github.com/Z3Prover/z3/releases/download/z3-4.8.8/z3-4.8.8-x64-ubuntu-16.04.zip
   jobs:
@@ -28,9 +29,9 @@ script:
   - diff --strip-trailing-cr Source/Core/Parser.cs Source/Core/Parser.cs.old
   - diff --strip-trailing-cr Source/Core/Scanner.cs Source/Core/Scanner.cs.old
   # Build Boogie
-  - dotnet build -c ${CONFIGURATION} ${SOLUTION}
+  - dotnet pack -c ${CONFIGURATION} ${SOLUTION}
   # Run unit tests
-  - dotnet test --no-build -c ${CONFIGURATION} ${SOLUTION}
+  - dotnet test -c ${CONFIGURATION} ${SOLUTION}
   # Run lit test suite
   - lit -v --timeout=120 -D configuration=${CONFIGURATION} Test
   # Do not deploy if GitVersionTask could not be used to extract the correct version number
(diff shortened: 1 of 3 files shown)


bkragl (approved):
Great, this will hopefully resolve the main source of issues coming in lately :)

In the CI log I saw warnings of the following kind:
```
/home/travis/.nuget/packages/gitversiontask/5.5.1/build/GitVersionTask.targets(10,9): warning : 2020-12-20 16:18:25		    WARN [12/20/20 16:18:25:14] Failed to inherit Increment branch configuration, ended up with: master, origin/master [/home/travis/build/boogie-org/boogie/Source/BoogieDriver/BoogieDriver.csproj]
/home/travis/.nuget/packages/gitversiontask/5.5.1/build/GitVersionTask.targets(10,9): warning : Falling back to master branch config [/home/travis/build/boogie-org/boogie/Source/BoogieDriver/BoogieDriver.csproj]
```

And packages with a strange version number (incremented minor version):
```
Successfully created package '/home/travis/build/boogie-org/boogie/Source/BoogieDriver/bin/Debug/Boogie.2.9.0-PullRequest0319.16.nupkg'.
```

But I guess the correct version will be computed after merging.

bkragl:
OK, there is actually a problem! I just tried to compile on my Ubuntu 20.10 machine, and it does not work. A lot of exceptions, apparently because of a library called LibGit2Sharp that is not supported on recent Ubuntu versions:
https://github.com/GitTools/GitVersion/issues/1744

It's really annoying that this GitVersion thing cannot just work.

Two options:
1) We don't merge this and need to continue to advertise the `MSBUILDSINGLELOADCONTEXT` workaround for all platforms.
2) We merge this and need to advertise the `BOOGIE_NO_GITVERSION` workaround to Ubuntu users (which means that they won't be able to generate packages with correct version numbers).

keyboardDrummer:
The `MSBUILDSINGLELOADCONTEXT` problem should be solved from GitVersionTask version 5.3.0. I'm guessing that version suffers from the same Ubuntu problem as the 5.5.1 in this PR, but it might be worth to try if it doesn't : )

> We merge this and need to advertise the BOOGIE_NO_GITVERSION workaround to Ubuntu users (which means that they won't be able to generate packages with correct version numbers).

Is it important for anything but the build server to generate versioned packages? Can we not only enable GitVersion on the build server?

bkragl:
> The `MSBUILDSINGLELOADCONTEXT` problem should be solved from GitVersionTask version 5.3.0. I'm guessing that version suffers from the same Ubuntu problem as the 5.5.1 in this PR, but it might be worth to try if it doesn't : )

Yes, it does. Turns out that even the current version 5.1.2 does. I kept the `BOOGIE_NO_GITVERSION` setting when recently switching to Ubuntu 20.10 and actually never compiled with GitVersion on that machine.

> Is it important for anything but the build server to generate versioned packages? Can we not only enable GitVersion on the build server?

Yeah, maybe this is the best move. Of course it would be nice to get the correct version number in binaries, but it is much more annoying that people constantly run into issues building Boogie. Once the day comes when GitVersion works reliably, we can re-enable it.

If others agree, let's do the following:
Change `BOOGIE_NO_GITVERSION` to `BOOGIE_GITVERSION` (flipping the conditional) in `Directory.Build.props` and set it in the Travis script.

Also, I think it makes sense not to generate the package files on every building.
Can you drop `<GeneratePackageOnBuild>true</GeneratePackageOnBuild>` and instead call `dotnet pack Source/Boogie.sln` in the Travis script? And drop the line about a generated package from the README.

@shazqadeer @michael-emmi @zvonimir

keyboardDrummer:
Thanks, I don't often do this type of "instruction driven programming" but I have to say it's very efficient :D

keyboardDrummer:
Shall we merge this?

shazqadeer (approved):

shazqadeer:
@keyboardDrummer : I appreciate your contribution in this PR towards addressing a longstanding annoyance for Boogie users.

PR #324, opened 2020-12-24
Treat incomplete-arithmetic and model-unavailable

### LINKED ISSUE #321: Unexpected prover response for incomplete Z3 arithmetic
Z3 v.4.8.9 can return `unknown` and state its reason as

    (:reason-unknown "smt tactic failed to show goal to be sat/unsat (incomplete (theory arithmetic))")

When it does, Boogie reports a "Prover error: Unexpected prover response":

```
Prover error: Unexpected prover response (getting info about 'unknown' response): (:reason-unknown "smt tactic failed to show goal to be sat/unsat (incomplete (theory arithmetic))")

d.bpl(1,11): Verification inconclusive (Test)

```

This isn't very friendly to a Boogie client (like Dafny) that would rather just get an ordinary Boogie "cannot prove assert" error. For example, if Z3 reports `unknown` and states its reason as having to do with incomplete quantifiers, then Boogie just report "cannot prove assert" with further hullabaloo.

To provoke Z3 to reveal its incomplete arithmetic theory, I used the following Boogie program:

```
procedure Test(a: int, b: int)
  requires 1 <= a;
  requires 1 <= b;
{
  assert a <= a * b;
}
```

and I invoked Boogied as follows:

```
boogie "/proverOpt:O:smt.arith.solver=2" d.bpl
```

RustanLeino:
This is an attempt to fix #321. However, when Z3 reports an arithmetic incompleteness, it apparently doesn’t always have a model available. The Boogie implementation does does not handle this gracefully.

Without my edit to `Source/Provers/SMTLib/SMTLibProcess.cs`, the further unexpected Z3 output causes Boogie to hang. With my edit to `Source/Provers/SMTLib/SMTLibProcess.cs`, Boogie reports 0 errors (and also 0 verified) on the repro example in #321:

```
$ boogie -proverOpt:O:smt.arith.solver=2 IncompleteArithmetic-git-issue-321.bpl

Boogie program verifier finished with 0 verified, 0 errors
```

This borders on an unsoundness, since it can easily be mistaken for an "everything is correct" outcome.

I've traced through the code trying to find a place to increase the error count even when there's no model. Alarmingly (or perhaps just because I don't understand the intent of the code), I found a number of places in the code that silently just return `null` if it cannot calculate a control-flow path (through the `CalculatePath` method, the `OnModel` that's called upon return of `CalculatePath`, and also `TraceCounterexample` if I remove its `if (labels.Count == 0) return;` and change its `Contract.Assert(traceNodes.Contains(entryBlock));` to add the disjunct `labels.Count == 0 ||`).

What is the intended behavior if Z3 does not return a model? Perhaps there's nothing Boogie can do in that case, so it should just report a `ProverError` and terminate (which is close to the behavior before I tried to fix #321)?

Help appreciated.

### DIFF
--- a/Source/Provers/SMTLib/ProverInterface.cs
+++ b/Source/Provers/SMTLib/ProverInterface.cs
@@ -1722,7 +1722,7 @@ namespace Microsoft.Boogie.SMTLib
                 labels = CalculatePath(handler.StartingProcId());
                 if (labels.Length == 0)
                 {
-                  // With a path to an error, we don't know what to report
+                  // Without a path to an error, we don't know what to report
                   globalResult = Outcome.Undetermined;
                   break;
                 }
--- a/Source/Provers/SMTLib/ProverInterface.cs
+++ b/Source/Provers/SMTLib/ProverInterface.cs
@@ -2356,6 +2356,8 @@ namespace Microsoft.Boogie.SMTLib
             {
               case "incomplete":
               case "(incomplete quantifiers)":
+              case "(incomplete (theory arithmetic))":
+              case "smt tactic failed to show goal to be sat/unsat (incomplete (theory arithmetic))":
                 break;
               case "memout":
                 currentErrorHandler.OnResourceExceeded("memory");
--- a/Source/Provers/SMTLib/ProverInterface.cs
+++ b/Source/Provers/SMTLib/ProverInterface.cs
@@ -1720,6 +1720,12 @@ namespace Microsoft.Boogie.SMTLib
               else
               {
                 labels = CalculatePath(handler.StartingProcId());
+                if (labels.Length == 0)
+                {
+                  // With a path to an error, we don't know what to report
+                  globalResult = Outcome.Undetermined;
+                  break;
+                }
               }
 
               handler.OnModel(labels, model, result);
--- a/Test/test2/IncompleteArithmetic-git-issue-321.bpl.expect
+++ b/Test/test2/IncompleteArithmetic-git-issue-321.bpl.expect
@@ -1,0 +1,5 @@
+/Users/leino/Source/boogie/Test/test2/IncompleteArithmetic-git-issue-321.bpl(11,3): Error BP5001: This assertion might not hold.
+Execution trace:
+    /Users/leino/Source/boogie/Test/test2/IncompleteArithmetic-git-issue-321.bpl(11,3): anon0
+
+Boogie program verifier finished with 0 verified, 1 error
--- a/Test/test2/IncompleteArithmetic-git-issue-321.bpl.expect
+++ b/Test/test2/IncompleteArithmetic-git-issue-321.bpl.expect
@@ -1,5 +1,2 @@
-IncompleteArithmetic-git-issue-321.bpl(11,3): This assertion might not hold.
-Execution trace:
-    IncompleteArithmetic-git-issue-321.bpl(11,3): anon0
 
-Boogie program verifier finished with 0 verified, 1 error
+Boogie program verifier finished with 1 verified, 0 errors
--- a/Test/test2/IncompleteArithmetic-git-issue-321.bpl.expect
+++ b/Test/test2/IncompleteArithmetic-git-issue-321.bpl.expect
@@ -1,5 +1,5 @@
-/Users/leino/Source/boogie/Test/test2/IncompleteArithmetic-git-issue-321.bpl(11,3): Error BP5001: This assertion might not hold.
+IncompleteArithmetic-git-issue-321.bpl(11,3): This assertion might not hold.
 Execution trace:
-    /Users/leino/Source/boogie/Test/test2/IncompleteArithmetic-git-issue-321.bpl(11,3): anon0
+    IncompleteArithmetic-git-issue-321.bpl(11,3): anon0
 
 Boogie program verifier finished with 0 verified, 1 error
--- a/Source/Core/VCExp.cs
+++ b/Source/Core/VCExp.cs
@@ -76,7 +76,7 @@ LOG_FILE=<string>         Log input for the theorem prover. The string @PROC@ in
 APPEND_LOG_FILE=<bool>    Append, rather than overwrite the log file.
 MEMORY_LIMIT=<int>        Memory limit of the prover in megabytes.
 VERBOSITY=<int>           The higher, the more verbose.
-TIME_LIMIT=<int>          Time limit per verification condition in miliseconds.
+TIME_LIMIT=<int>          Time limit per verification condition in milliseconds.
 
 The generic options may or may not be used by the prover plugin.
 ";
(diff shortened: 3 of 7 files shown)


Test/test2/IncompleteArithmetic-git-issue-321.bpl.expect
```diff
@@ -0,0 +1,5 @@
+/Users/leino/Source/boogie/Test/test2/IncompleteArithmetic-git-issue-321.bpl(11,3): Error BP5001: This assertion might not hold.
+Execution trace:
+    /Users/leino/Source/boogie/Test/test2/IncompleteArithmetic-git-issue-321.bpl(11,3): anon0
+
+Boogie program verifier finished with 0 verified, 1 error
```
shazqadeer, line 5: The PR description says:

```
With my edit to Source/Provers/SMTLib/SMTLibProcess.cs, Boogie reports 0 errors (and also 0 verified) on the repro example in #321:

$ boogie -proverOpt:O:smt.arith.solver=2 IncompleteArithmetic-git-issue-321.bpl

Boogie program verifier finished with 0 verified, 0 errors
```

which contradicts what I am seeing in the .expect file.  What am I not understanding?

bkragl: I think the `.expect` file shows (something close to) what we want to get as an output. But the changes in the PR do not yet produce this desired result.

RustanLeino: Correct. Sorry for the confusion. The `.expect` file I gave is what I (think I) expect the result to be, but my PR does not achieve that result.

shazqadeer: @RustanLeino : Recall that Boogie's counterexample generation is now based on querying the model (rather than labels as was the case earlier).  Since in your example a model is not produced, it is unreasonable to expect a counterexample in the .expect file.

shazqadeer: Actually, I see that .expect file has changed again to "1 verified, 0 errors", which also does not make any sense to me since I am expecting that the version of Z3 supported by Boogie (presumably 4.8.9) is unable to verify this example.

shazqadeer:
It might also be helpful to realize that the error count in message reported on the console by Boogie:

is the value in stats.ErrorCount.  The update of ErrorCount filed is in ExecutionEngine.cs and is tied to the actual number of counterexamples calculated.

The management of errorsDiscovered variable inside CheckOutcomeCore is purely some bookkeeping internal to CheckOutcomeCore.

Test/test2/IncompleteArithmetic-git-issue-321.bpl.expect
RustanLeino: The error happens with Z3 v.4.8.9. I was about to check in the corresponding .expect file when I noticed that the Boogie build uses Z3 v.4.8.8. That version of Z3 _is_ is able to the assertion, and hence I changed the .expect file to reflect the "1 verified, 0 errors".

I have also tested the behavior with Z3 v.4.8.9, which returned "unknown" and has no model to show for. This PR includes code that attempts to report the error. I may not have done that in the best way. It seems that `CalculatePath` silently swallows any unexpected communication with Z3, which can result in it returning an empty path.

RustanLeino:
If Z3 has no model, then Boogie has no way to even report which assertion failed, right? Without knowing the failing assertion, the error message that Boogie or Dafny would report is rather lame (e.g., "there an error somewhere in your method").

After just adding the cases for "incomplete (theory arithmetic)" in the code, then I had cases where Boogie would end up saying "0 verified, 0 errors". It's misleading to not account for procedures in the input. If it says "0 errors", then a user expects everything to be good (and the "N verified" part of the message is probably ignored--especially for Dafny users, for whom the number of Boogie procedures generated behind the scenes is not known). So, in the event that Z3 produces no model, then Boogie still needs to do something. Since I wasn't able to produce a `Counterexample` without a model, I tried for the `Undetermined` outcome instead. Maybe you can suggest a better way?

Note, in the near future, Z3 will have a mode to produce a model no matter what, see https://github.com/Z3Prover/z3/issues/4924. However, I thinking we'd like Boogie to support a variety of versions of Z3. (For example, I'm bringing Dafny up to Z3 v.4.8.5, but Dafny doesn't do well with Z3 v.4.8.8 or v.4.8.9, so it would be nice to be able to use the latest Boogie and an older Z3.)

shazqadeer:
@RustanLeino :  Would I be able to commit edits into your branch which is the basis of this PR?  And is that ok with you?  I thought it might be easier if I could collaborate directly on the code with you.

shazqadeer:
@RustanLeino :  Your fix in CheckOutcomeCore looks good to me.  The code there is not written cleanly.  I would like to refactor it but I can do that in a separate PR.  I will approve your PR.

shazqadeer (approved):

RustanLeino:
> Your fix in CheckOutcomeCore looks good to me. The code there is not written cleanly. I would like to refactor it but I can do that in a separate PR. I will approve your PR.

Great. Thanks. (I saw your other request a little late, but would have been fine with you committing directly to my branch.)

RustanLeino:
There's some build error, some duplicate name thing during the `dotnet pack`. This seems unrelated to this PR.

Source/Provers/SMTLib/ProverInterface.cs
```diff
@@ -1720,6 +1720,12 @@ public override Outcome CheckOutcomeCore(ErrorHandler handler, int taskID = -1)
               else
               {
                 labels = CalculatePath(handler.StartingProcId());
+                if (labels.Length == 0)
+                {
+                  // With a path to an error, we don't know what to report
```
bkragl, line 1725: With → Without

bkragl (commented):
Changes look good to me.

Just two things to clarify:

For the test `IncompleteArithmetic-git-issue-321.bpl` I get the output in the `.expect` file (`1 verified, 0 errors`) with Z3 4.8.8. With Z3 4.8.9 I get `0 verified, 0 errors, 1 inconclusive` which seems like the right thing to report, but I also see the line `Prover error: line 31 column 30: model is not available`. Is this something we eventually want to eliminate, but it needs further refactoring before this can be done?

For the test `IncompleteArithmetic-RealTypes.bpl` I get `2 verified, 5 errors` with both above Z3 versions. The comment in the file says that it has a different `:reason-unknown` response from Z3 than the above test. But why is there no difference in Boogie output with the two Z3 versions?

bkragl:
> There's some build error, some duplicate name thing during the `dotnet pack`. This seems unrelated to this PR.

Yeah, we are having PR builds fail nondeterministically recently. I'll try something to see if this can be fixed. Anyway, I just restarted the build job and it went through as expected, no problem with this PR.

RustanLeino:
Thanks, @bkragl . I improved the comments in various files to clarify the situation.

@shazqadeer , if you do rewrite the code related to CheckOutcomeCore, I think it would be nice if the "Prover error" that @bkragl refers to above could be eliminated. I added some comments about this into `Test/test2/IncompleteArithmetic-git-issue-321.bpl`.

shazqadeer:
@bkragl : Are we ready to merge this PR?

bkragl (approved):

PR #325, opened 2020-12-27
Remove Boogie error codes

bkragl:
**Context**

Boogie outputs the following error codes alongside error messages:

`BP5001` failed assertion
`BP5002` failed precondition
`BP5003` failed postcondition
`BP5004` failed loop invariant initialization
`BP5005` failed loop invariant maintenance
`BP5010` anything causing a `VCGenException`

So an error message might look like this:
```
bla1.bpl(2097,5): Error BP5001: This assertion might not hold.
Execution trace:
    bla1.bpl(751,3): start#1
    ...
```

**The problem**

These error codes look odd when tools encode their own checks but use Boogie's error reporting with custom error messages.

Dafny example:
```
Inc.dfy(318,17): Error BP5001: cannot prove termination; try supplying a decreases clause
```

CIVL example:
```
left-mover.bpl(19,32): Error BP5001: Commutativity check between init and inc failed
```

**This PR**

This PR drops error codes. They do not seem particularly helpful to Boogie users. It's just the 6 above; nothing for parsing, type checking, etc. And there is no documentation.

**Other options**

Two other options could be:

1. Introduce a configuration flag that disables error codes. Clients that don't like error codes can disable them.
2. Add the possibility to attach error codes to `Requires`, `Ensures`, and `AssertCmd` (anything else?). Clients can either put a custom error code or disable (by setting to `null`).

**Let's vote**

@RustanLeino @shazqadeer @zvonimir @akashlal @alexanderjsummers @gauravpartha 
Do you want to keep error codes? In which form?
Does this affect your tools?

### DIFF
--- a/Source/ExecutionEngine/ExecutionEngine.cs
+++ b/Source/ExecutionEngine/ExecutionEngine.cs
@@ -254,7 +254,6 @@ namespace Microsoft.Boogie
     public readonly IToken Tok;
     public string Msg;
     public string Category { get; set; }
-    public string BoogieErrorCode { get; set; }
     public readonly List<AuxErrorInfo> Aux = new List<AuxErrorInfo>();
     public string OriginalRequestId { get; set; }
     public string RequestId { get; set; }
@@ -267,13 +266,7 @@ namespace Microsoft.Boogie
     {
       get
       {
-        var prefix = Category;
-        if (BoogieErrorCode != null)
-        {
-          prefix = prefix == null ? BoogieErrorCode : prefix + " " + BoogieErrorCode;
-        }
-
-        return prefix != null ? string.Format("{0}: {1}", prefix, Msg) : Msg;
+        return Category != null ? string.Format("{0}: {1}", Category, Msg) : Msg;
       }
     }
 
@@ -1239,7 +1232,6 @@ namespace Microsoft.Boogie
           {
             var errorInfo = errorInformationFactory.CreateErrorInformation(impl.tok,
               String.Format("{0} (encountered in implementation {1}).", e.Message, impl.Name), requestId, "Error");
-            errorInfo.BoogieErrorCode = "BP5010";
             errorInfo.ImplementationName = impl.Name;
             printer.WriteErrorInformation(errorInfo, output);
             if (er != null)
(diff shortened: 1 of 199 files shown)


akashlal:
Corral does not use these error codes.

shazqadeer:
From my point of view, it is fine to drop these error codes.

gauravpartha:
Viper does not use these error codes, thanks for asking.

RustanLeino:
Dafny makes use of Boogie's distinction between the categories above (e.g., it uses Boogie's "loop invariant initialization" and "loop invariant maintenance" distinction), but it never looks at or otherwise uses error codes. So, from Dafny's perspective, this is a good PR to accept. I also don't think the error codes add anything for Boogie itself.

shazqadeer (approved):

DavePearce:
I realise this is a moot issue now, but personally I found the error codes useful.  In particular, for consistency, Whiley doesn't report Boogie errors directly.  Rather, for each class, we report the standard Whiley error for it.  This means error reporting is consistent across backend verifiers.

@RustanLeino It sounds like you do something similar.  Presumably you are not matching on the _contents_ of the error message.  That seems the only way I can make it work, but its brittle if the error messages change.

RustanLeino:
The Dafny tool links directly with Boogie (that is, it does not call Boogie as an executable). In this way, Boogie reports errors to Dafny by, essentially, handing Dafny the AST node (assert, requires, ensures, ...) that fails (rather than Dafny needing to parse some text output from Boogie). Dafny associates with each such AST node an error message to be displayed, so when Boogie hands Dafny the AST nodes associated with the failures, Dafny prints the error messages it had prepared.

DavePearce:
Hey @RustanLeino,

Right, of course --- you link directly to the DLL.  Sadly, I don't have that luxury having implemented Whiley in Java.  Still, I do a reverse lookup on the AST (roughly as you are describing) based on the text errors reported by Boogie.

RustanLeino:
Hi @DavePearce ,

To make your reverse lookup easier and more precise, you can use the `{:msg "your text goes here"}` annotation on assertions (see `boogie -attrHelp`). For example, you could generate a unique ID for each assertion and then put that into the `:msg` argument.

PR #329, opened 2021-01-07
[CIVL] fixed monomorphization bug

shazqadeer:
Monomorphization had a bug where the modifies list of a procedure was not being transformed appropriately.  This PR fixes the bug.  This PR also does small cleanup in the trieber stack sample.

### DIFF
--- a/Source/Core/Monomorphization.cs
+++ b/Source/Core/Monomorphization.cs
@@ -639,6 +639,11 @@ namespace Microsoft.Boogie
       return exprMonomorphizationVisitor.VisitExpr(node);
     }
 
+    public override Expr VisitIdentifierExpr(IdentifierExpr node)
+    {
+      return exprMonomorphizationVisitor.VisitExpr(node);
+    }
+
     public override Declaration VisitTypeCtorDecl(TypeCtorDecl node)
     {
       if (node is DatatypeTypeCtorDecl datatypeTypeCtorDecl)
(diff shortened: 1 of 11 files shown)


bkragl (approved):
Nice! I also ported seqlock and Paxos to use polymorphism. Please take a look at Paxos.

It is nice that instead of

```
type {:datatype} OptionValue;
function {:constructor} SomeValue(value: Value): OptionValue;
function {:constructor} NoneValue(): OptionValue;

type {:datatype} OptionVoteInfo;
function {:constructor} NoneVoteInfo(): OptionVoteInfo;
function {:constructor} SomeVoteInfo(value: Value, ns: NodeSet): OptionVoteInfo;

var {:layer 1,3} voteInfo: [Round]OptionVoteInfo;
var {:layer 0,3} decision: [Round]OptionValue;
```

we get

```
type {:datatype} Option _;
function {:constructor} None<T>(): Option T;
function {:constructor} Some<T>(t: T): Option T;

type {:datatype} VoteInfo;
function {:constructor} VoteInfo(value: Value, ns: NodeSet): VoteInfo;

var {:layer 1,3} voteInfo: [Round]Option VoteInfo;
var {:layer 0,3} decision: [Round]Option Value;
```

But consequently, instead of `value#SomeVoteInfo(voteInfo[r'])` we have to do nested deconstructions `value#VoteInfo(t#Some(voteInfo[r']))`. Maybe something other then `t` for the parameter to `Some` would be better?

shazqadeer:
I don't have a better suggestion for t.

PR #330, opened 2021-01-12
Fixed error output when :msg attribute is supplied

### LINKED ISSUE #328: Unexpected behaviour of :msg attribute
It seems that since the merged pull request https://github.com/boogie-org/boogie/pull/282 (which fixes issue https://github.com/boogie-org/boogie/issues/272) the behaviour of the :msg attribute for assertions has changed in a way that I did not expect (I noticed this while trying to update a previous Boogie version to the newest one for Viper).

Before the PR, failure of the assertion `assert {:msg "test message"} false` rendered the output `test message` (using `/errorTrace:0`). Since the PR, the output is:  `.\test.bpl(2,5): Error: test message`. 

So, the output includes a part that was never specified as a value for the msg attribute (filename + Boogie position + "Error:"). This is a problem for our use case, since we directly output Boogie's output.

Was this change in behaviour intended or is it possible to adjust the behaviour to just output the value specified?

shazqadeer:
Closes #328.

### DIFF
--- a/Source/ExecutionEngine/ExecutionEngine.cs
+++ b/Source/ExecutionEngine/ExecutionEngine.cs
@@ -1565,12 +1565,14 @@ namespace Microsoft.Boogie
                   }
                   else
                   {
-                    msg = assertError.FailingAssert.ErrorData as string;
-                    if (!CommandLineOptions.Clo.ForceBplErrors && assertError.FailingAssert.ErrorMessage != null)
+                    if (assertError.FailingAssert.ErrorMessage == null || CommandLineOptions.Clo.ForceBplErrors)
+                    {
+                      msg = assertError.FailingAssert.ErrorData as string;
+                    }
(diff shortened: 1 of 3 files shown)


bkragl (approved):
Code looks good. Can we add a simple test that shows the usage of `:msg`? Perhaps it makes sense to also set `-errorTrace:0` in this test to indicate that this flag makes sense in combination with `:msg` (because otherwise an error trace with Boogie source locations would be printed).

One caveat here is that users of `:msg` will get the same output, no matter if the failure cause is an actual proof error or out of time/memory/resources.

shazqadeer:
I have added a test case.  CI does not appear to be running.

bkragl:
Could you extend the test to also have `:msg` on a precondition and a postcondition?

shazqadeer:
Done.

PR #331, opened 2021-01-12
Put functions and procedures into different namespaces

shazqadeer:
This PR puts functions and procedures into separate namespaces allowing a function to have the same name as a procedure.


shazqadeer:
@bkragl : The CI checks do not appear to be running on this PR.  Any idea what could be wrong?

bkragl:
> @bkragl : The CI checks do not appear to be running on this PR. Any idea what could be wrong?

If you go [here](https://travis-ci.com/github/boogie-org/boogie) it says:
> Builds have been temporarily disabled for public repositories due to a negative credit balance. Please go to the Plan page to replenish your credit balance or alter your Consume paid credits for OSS setting.

Looks like Travis CI has a [new pricing model](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing) (search for the section `Building on a public repositories only`), and we ran out of our 10K free credit allotment. We should get in contact with the Travis CI support to apply for more credits. @michael-emmi @zvonimir Any idea what a good/sufficient amount would be?

shazqadeer:
I applied for more credits.  I noticed that each PR runs two builds which together take about 13 minutes (according to data published by Travis). I suggested in my request that we need 1500 minutes of build time every year.

zvonimir:
We (meaning @shaobo-he) recently switched SMACK CI to github actions instead of travis. It seems that it wasn't too hard of a switch. You can look at our config file to maybe serve as an example. Github actions are free and the switch really improved our CI turnaround time. I was paying for Travis for 2 months, but it is just not worth it.

bkragl (approved):
Very nice, the CIVL examples with pending asyncs look much better now!

I did some more refactoring to replace all non-generic collections in `ResolutionContext` with generic ones, which eliminated a number of type casts.

bkragl:
Perhaps we should wait with merging this until we have CI running again.

shazqadeer:
Travis support got back to my request with a questionnaire.  I responded.   Now we are waiting to hear back.

@zvonimir :  I am open to github actions  even if the travis support comes through.

PR #333, opened 2021-01-23
Add example that asynchronously sums the numbers from 1 to n

bkragl:
This is a simple example to illustrate inductive sequentialization.

It creates `n` asynchronous `ADD(i)` operations (`i` between `1` and `n`), each adding `i` to the global variable `x`. Instead of considering all interleavings, inductive sequentialization allows us to only consider `ADD(1); ADD(2); ...; ADD(n)`. Thus we can use the identity `1 + 2 + ... + n = (n * (n+1)) / 2` (which Gauss allegedly figured out as a child) in the invariant.

Because nonlinear invariants are so much fun, also a sequential version of the example is added.

### DIFF
--- a/Test/civl/inductive-sequentialization/Gauss.bpl
+++ b/Test/civl/inductive-sequentialization/Gauss.bpl
@@ -1,0 +1,58 @@
+// RUN: %boogie "%s" > "%t"
+// RUN: %diff "%s.expect" "%t"
+
+var {:layer 0,2} x:int;
+
+type {:pending_async}{:datatype} PA;
+function {:constructor} ADD(i: int) : PA;
+
+////////////////////////////////////////////////////////////////////////////////
+
+function trigger(i: int) : bool { true }
+
+procedure {:atomic}{:layer 1}
+{:IS "MAIN'","INV"}{:elim "ADD"}
+SUM (n: int)
+returns ({:pending_async "ADD"} PAs:[PA]int)
+modifies x;
+{
+  assert n >= 0;
+  assert trigger(0); // base hint
+  PAs := (lambda pa: PA :: if is#ADD(pa) && 1 <= i#ADD(pa) && i#ADD(pa) <= n then 1 else 0);
+}
+
+procedure {:atomic}{:layer 2}
+MAIN' (n: int)
+modifies x;
+{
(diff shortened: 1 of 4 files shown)


shazqadeer (approved):

Test/civl/inductive-sequentialization/Gauss.bpl
```diff
@@ -0,0 +1,58 @@
...
+  assert trigger(0); // base hint
+  PAs := (lambda pa: PA :: if is#ADD(pa) && 1 <= i#ADD(pa) && i#ADD(pa) <= n then 1 else 0);
+}
+
+procedure {:atomic}{:layer 2}
+MAIN' (n: int)
+modifies x;
+{
+  assert n >= 0;
+  x := x + ((n * (n+1)) div 2);
```
shazqadeer, line 29: you might not need one level of parentheses.

Test/civl/inductive-sequentialization/Gauss.bpl
```diff
@@ -0,0 +1,58 @@
...
+
+procedure {:atomic}{:layer 1}
+{:IS "MAIN'","INV"}{:elim "ADD"}
+SUM (n: int)
+returns ({:pending_async "ADD"} PAs:[PA]int)
+modifies x;
+{
+  assert n >= 0;
+  assert trigger(0); // base hint
+  PAs := (lambda pa: PA :: if is#ADD(pa) && 1 <= i#ADD(pa) && i#ADD(pa) <= n then 1 else 0);
```
shazqadeer, line 21: Because PA has only one constructor, is#Add(pa) is always true.

Test/civl/inductive-sequentialization/Gauss.bpl
bkragl: Right, I updated it. It would even be possible to drop the parentheses around `n * (n+1)`, but this is getting too close to anarchy for me :smile:

Test/civl/inductive-sequentialization/Gauss.bpl
bkragl: Yes, but I decided to put it nevertheless, in case somebody is playing with the example and adds another constructor.

PR #334, opened 2021-01-31
Remove :captureState dependence on model.compact

RustanLeino:
This PR reimplements the `:captureState` behavior (for default VC generation) in a way that does not depend on disabling model compression in Z3. Since the option to turn off model compression depends on the Z3 version, this PR lets the functionality be used with a variety of Z3 versions.

## Background and purpose

The `{:captureState "any string"}` attribute can be placed on `assume` statements (typically, on `assume true;` statements). If such a statement is along the trace to an error, Boogie will harvest the SMT solver's counterexample model to reconstruct values for all variables in that "captured" state. This information can be printed out using the `-mv` switch, which will list the sequence of captured states, each one described with the user-specified `"any string"`.

This feature is being used by the [VS Code plug-in for Dafny](https://github.com/dafny-lang/language-server-csharp), since the Boogie programs that Dafny emits contain such `:captureState` attributes. This feature was originally designed and used by the now-deprecated Boogie Verification Debugger.

## Problem

The encoding that made `:captureState` work uses a function `$mv_state`, which it arranges to be set to `true` along the error trace. The idea was that these points of being `true` would show up in Z3's model. However, since the points of interest are all `true` and there is never a reason for the SMT solver to pick a model with a value `false`, the model-compression feature of Z3 would just say that `$mv_state` is `true` everywhere, which is of no use to the `:captureState` mechanism.

To get the `$mv_state` encoding to work, model compression must be disabled in the SMT solver. In Z3, model compression was introduced in v4.8.1, and along with it an option `model_compress` that would turn model compression off. This was still supported in Z3 v.4.8.5 (and maybe slightly later, I don't know). This global Z3 option `model_compress` was being phased out, its replacement being named `model.compact`. (I'm saying "replacement", but I'm not actually sure if `model.compact` is supposed to do the same thing as `model_compress` did.)

From my experiments, Z3 v4.8.5 responds as expected to `model_compress`. Apparently, this version also accepts the existence of a `model.compact` option, but ignores whatever it's set to. In Z3 v4.8.8, any use of `model_compress` gives an error, but `model.compact` seems to work satisfactorily. The upshot of this is that if you are using `:captureState`, then:

* With Z3 v4.8.5, you (or Boogie) should set `model_compress=false`. It's okay if you also set `model.compact=false`, since that seems not to do anything.
* With Z3 v4.8.8, you _must not_ set `model_compress_false`, since that will give an error. Instead, you should set `model.compact=false`.

Boogie's test suite currently seems to expect Z3 v4.8.8, so Boogie is currently setting `model.compact=false` (and _not_ doing anything with the deprecated `model_compress`). This makes tests like `test15/CaptureState.bpl` work.

Dafny uses (the current Boogie and) Z3 v4.8.5. This means that Boogie supplies `model.compact=false`, which doesn't do anything with Z3 v4.8.5, which means that Dafny's use of `:captureState` doesn't work. (At present, it is not feasible to upgrade Dafny to use a newer version of Z3.)

## Possible solution

A quick solution would be for Dafny (and, in particular, the VS Code plug-in for Dafny to supply the Z3 option `model_compress=false`. This would work as long as the Z3 version used is v4.8.5. However, this would mean one cannot experimentally try Dafny with a newer version of Z3, since the use of `model_compress` would then cause an error.

## The solution

This PR provides a different solution, which is to use Boogie's (somewhat new) `ControlFlow` encoding to figure out which `:captureState` attributes lie along the error trace. This is, after all, how Boogie computes error traces. Since `ControlFlow` uses an integer function that points out successor blocks in the error trace, it is not sensitive to model compression in the way that `$mv_state` was.

While Boogie's test suite is currently geared for Z3 v4.8.8, I've manually successfully tested this PR on the new `Test/test15/MoreCapturedStates.bpl` with both Z3 v4.8.5 and v4.8.8.

## Limitations

I did not remove the `$mv_state` encoding, so those functions are still being generated into the SMT formula. The reason I didn't is that FixedpointVC.cs and StratifiedVC.cs still make use of it. I couldn't tell how these use `$mv_state`, and didn't notice tests that prove that behavior. Therefore, I left that code unchanged. (If this functionality is not needed or can be replaced, then `$mv_state` and `$mv_state_const` can be removed from the encoding, and also `ModelViewInfo.CapturePoints` in ModelViewInfo.cs.)

### DIFF
--- a/Source/Core/CommandLineOptions.cs
+++ b/Source/Core/CommandLineOptions.cs
@@ -1879,8 +1879,6 @@ namespace Microsoft.Boogie
        shows the values of variables at each {:captureState ...} point in
        the counterexample trace. The argument 's' is to be a string, and it
        will be printed as part of /mv's output.
-       Note, the use of :captureState and /mv have not been implemented for
-       use with procedure inlining.
 
   ---- CIVL ------------------------------------------------------------------
 
--- a/Source/Core/CommandLineOptions.cs
+++ b/Source/Core/CommandLineOptions.cs
@@ -1873,8 +1873,14 @@ namespace Microsoft.Boogie
        used in conjunction with /enhancedErrorMessages:n command-line option.
 
      {:captureState s}
-       Indicates that state must be captured under the name s.  Must be used
-       together with /mv:<string> command-line option.
+       When this attribute is applied to assume commands, it causes the 
+       /mv:<string> command-line option to group each counterexample model
+       into a sequence of states. In particular, this sequence of states
+       shows the values of variables at each {:captureState ...} point in
+       the counterexample trace. The argument 's' is to be a string, and it
+       will be printed as part of /mv's output.
+       Note, the use of :captureState and /mv have not been implemented for
+       use with procedure inlining.
 
   ---- CIVL ------------------------------------------------------------------
 
--- a/Source/ExecutionEngine/ExecutionEngine.cs
+++ b/Source/ExecutionEngine/ExecutionEngine.cs
@@ -1787,7 +1787,7 @@ namespace Microsoft.Boogie
 
           if (CommandLineOptions.Clo.ModelViewFile != null)
           {
-            error.PrintModel(errorInfo.Model);
+            error.PrintModel(errorInfo.Model, error);
           }
 
           printer.WriteErrorInformation(errorInfo, tw);
--- a/Source/VCGeneration/ConditionGeneration.cs
+++ b/Source/VCGeneration/ConditionGeneration.cs
@@ -911,7 +911,7 @@ namespace VC
         ChecksumHelper.ComputeChecksums(c, currentImplementation, variableCollector.UsedVariables, currentChecksum);
         variableCollector.Visit(c);
         currentChecksum = c.Checksum;
-        TurnIntoPassiveCmd(c, incarnationMap, oldFrameSubst, passiveCmds, mvInfo);
+        TurnIntoPassiveCmd(c, b, incarnationMap, oldFrameSubst, passiveCmds, mvInfo);
       }
 
       b.Checksum = currentChecksum;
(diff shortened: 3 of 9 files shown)


shazqadeer:
@akashlal : I noticed that  in StratifiedVC.cs, there is a function GetModelWithStates that is looking at $mv_state.  Is this really being used by Corral?

akashlal:
@shazqadeer This looks like experimental code. Corral is not using it.

akashlal:
I see that the same code is copied into FixedpointVC. I don't really understand the purpose of this code. It might be populating a counterexample trace with values of variables, but I don't know if its just for debugging, or it serves a different purpose.

shazqadeer:
@RustanLeino:

In Z3.cs, I see the following code:

```
// {:captureState} does not work with compressed models
options.AddWeakSmtOption("model.compact", "false"); // default: false
```

Should both of those lines be dropped now as a consequence of your PR?  If you agree, please make the change.

shazqadeer:
@RustanLeino:

There is a rat's nest here. It looks to me that there are two parallel implementations of "state capturing" in Boogie.  Take a look at the class CapturedState in Model.cs.  Both implementations are using the `$mv_state` function.  

Ideally, we should unify the two implementations.  But I would understand if you want to land Dafny 3.0 before embarking on this journey.

RustanLeino:
I looked some more at the code in StratifiedVC.cs. It's not clear to me, either, what code is still being used.

I tried removing the `model.compact=false` from Boogie. The test suite still passes, modulo differences in values in the models. So, this setting can be removed.

However, by experimentation, I found that the test suite still relies on `model.v2=true`. So, even if `model.compact=false` is removed in the future, we should keep `model.v2=true`.

From what's been said, I propose that we merge this PR as is, and then embark on a different PR that cleans up the state capturing in StratifiedVC.cs and FixedpointVC.cs.

shazqadeer (approved):

shazqadeer:
Approving the PR.

PR #345, opened 2021-02-17
New method for programmer-aided quantifier instantiation

shazqadeer:
This PR adds a new method for instantiating quantifiers based on user-supplied attributes on bound variables and other syntactic elements that produce values.  This scheme is complementary and orthogonal to the existing scheme based on pattern-matching and triggers.

New attributes:
- `:inst_at "Label"` may be added to a bound variable of a quantifier or a lambda. This hint suggests that any terms added to set denoted by "Label" should be used to instantiate this bound variable. 
- `:inst "Label", expr` may be added to a PredicateCmd, AssignCmd, or the body of a quantifier or lambda.  This hint suggests that `expr` be added to the set `"Label"`.   If the attribute is put on a quantifier or lambda, this addition occurs when the quantifier is either skolemized or instantiated or when the lambda is instantiated.

The implementation is done as a preprocessing on the passified implementation of a procedure just before the conversion to a VCExpr.  The implementation handles both skolemization and instantiation. The original quantifier is left in the formula; the generated instances are simply added to it.

The PR also does a couple of adjustments to other parts of Boogie:
- Moves the resolution of attributes on variables to the code that resolves variables (used to be in the code which resolved where clauses).
- Changes the order of bound variables in the axiom generated for a lambda; now the captured variables occur before the bound variables of the lambda.  

TODO:
- More tests
- Handle instantiation of quantifiers in non-lambda axioms
- Documentation of attributes

I am looking for a review of the code  in QuantifierInstantiationEngine.cs and suggestions about better names for the two attributes.


shazqadeer:
One other point for the consideration of the reviewers:  The implementation of the new quantifier instantiation method is always on.  As a result, even if no occurrences of :inst or :inst_at are present in the program, skolemization would be done.  This is the reason for the failure of the two regression (see CI results).  Skolemization can create more errors in case subsumption is used.  It can also result in changes in the returned model.  Should the implementation be hidden under a command-line option or perhaps I can trigger it by the existence of one of the relevant attributes in the program?

Test/civl/async/async4.bpl
```diff
@@ -0,0 +1,33 @@
...
+{ assert 0 <= n; x := x + n; }
+
+procedure {:IS_invariant}{:layer 1} INV(n: int) returns ({:pending_async "A_Inc"} PAs: [PA]int)
+modifies x;
+{
+  var {:inst_at "A"} i: int;
+  assert 0 <= n;
+  assume {:inst "A", i} {:inst "A", i+1} 0 <= i && i <= n;
+  x := x + i;
+  PAs := MapConst(0)[A_Inc() := n - i];
```
shazqadeer, line 20: @bkragl :  If I write this assignment as 

PAs[A_Inc()] := n - i

it is logically equivalent but the verification does not work. This happens because the initial value of the output variable PAs becomes an existentially quantified variable of the transition relation.  I could have supplied a hint for this variable also but that would have been no better than the alternative I chose.

bkragl: Yes, I see. But this is a special case where `PAs[A_Inc()] := n - i` only works because the domain of `PAs` is a singleton (because `PA` has only a single constructor with no parameters). I think the way you write it is much better than supplying a hint.

bkragl: Did you notice the syntax highlighting we have now for Boogie code? :smiley:

shazqadeer: I hadn't noticed it.  But I see it now.  Cool!

shazqadeer:
I am going to land this PR.  I have barely scratched the surface with this initial PR on this new approach to instantiation.  There  will be other opportunities to examine the code.  

I am not going to document the attributes for now since I feel that they are still experimental and not adequately tested.

PR #352, opened 2021-03-20
Quantifier instantiation on VC expressions

shazqadeer:
This PR makes the following changes:

- It revises the earlier label-based quantifier instantiation to work on VC expressions rather Boogie AST expressions.  
   - The names of attributes have changed.  
   - Many more tests have been added.  
   - BroadcastConsensus.bpl has been ported to BroadcastConsensus2.bpl.

- It cleans up the way quantifier weight is being put into an instance of  VCQuantifierInfo.

- It changes fields of BoundVarTraversingVCExprVisitor from Dictionary to Set.


shazqadeer:
FYI: @RustanLeino , @bkragl

bkragl:
What made you change your mind to implement QI on VC expr rather than on Boogie AST? Just curious.

shazqadeer:
After the implementation over Boogie AST was done, I discovered that functions are inlined during VC generation.  Consequently, my implementation would not have handled quantifiers inside inlined functions.  This was the forcing function.  But doing the implementation on VCExpr's also seems the right thing to do because of generality and fewer cases, e.g., I don't have to think about assumes/asserts etc.

PR #354, opened 2021-03-21
Move Vec to LibraryDefinitions

### LINKED ISSUE #353: Bug in monomorphization
The BPL file [bpl.zip](https://github.com/boogie-org/boogie/files/6176394/bpl.zip) produces the following errors when run by Boogie: 

```
Prover error: line 22 column 19: invalid datatype declaration, unknown sort 'T@$TypeName'

Prover error: line 22 column 19: invalid datatype declaration, unknown sort 'T@$TypeName'

Prover error: Prover died

...
```

Command line (after unzipping):

```
boogie -doModSetAnalysis -printVerifiedProceduresCount:0 -printModel:1 -enhancedErrorMessages:1 -proverOpt:PROVER_PATH=/home/wrwg/bin/z3 -proverOpt:O:smt.QI.EAGER_THRESHOLD=100 -proverOpt:O:smt.QI.LAZY_THRESHOLD=100 -vcsCores:12 output.bpl
```

The culprit seems to be the application of the generic function `function $ForceTrigger<T>(x: T): bool { true }`. I have tried to nail this down to a smaller use case but was not able to do so.

@shazqadeer

shazqadeer:
This PR does the following:

- Moves native Vec (finite vectors) into LibraryDefinitions.bpl
- Automatically turn on /useArrayTheory and /monomorphize when /lib is used
- Improve error reporting when polymorphism combined with datatypes

Closes #353 .

### DIFF
--- a/Source/Core/LibraryDefinitions.bpl
+++ b/Source/Core/LibraryDefinitions.bpl
@@ -39,6 +39,7 @@ function {:constructor} Vec<T>(contents: [int]T, len: int): Vec T;
 
 const Identity: [int]int;
 axiom (forall x: int :: Identity[x] == x);
+
 function {:inline} AtLeast(x: int): [int]bool
 {
   MapLe(MapConst(x), Identity)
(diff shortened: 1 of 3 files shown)


DavePearce:
Hey,

So, how do you use these library definitions then?  Can you import them somehow?

shazqadeer:
Use the command-line option `-lib` to pull them in.  See this [sample](https://github.com/boogie-org/boogie/blob/master/Test/monomorphize/vector.bpl) for examples.  Also, here is the link to [LibraryDefinitions.bpl](https://github.com/boogie-org/boogie/blob/master/Source/Core/LibraryDefinitions.bpl) which is included into the compiled into Boogie.Core.dll.

DavePearce:
Hey,

Ok, interesting --- thanks.  Looking at the file, I see that Boogie has some form of generics!  What's the best reference for the Boogie language, as I wasn't aware this was possible!

RustanLeino:
Boogie generics are described in Boogie language reference manual [This is Boogie 2](https://www.microsoft.com/en-us/research/publication/this-is-boogie-2-2/) and in the TACAS 2010 paper [A Polymorphic Intermediate Verification Language: Design and Logical Encoding](http://leino.science/papers/krml199.pdf). Note that some Boogie options are supported only in the monomorphic mode.

PR #375, opened 2021-04-18
Invoking yices-smt2 as Yices2 binary

### LINKED ISSUE #374: Yices2 binary is called yices-smt2
When Yices2 is installed, it comes with the following binaries: `yices`, `yices-sat`, `yices-smt`, `yices-smt2`.
The one Boogie needs is called yices-smt2. Currently Boogie is invoking `yices2`, and the solver options is `Yices2`.
I suggest we either rename the solver option as `yices-smt2`, or alternatively it remains `Yices2` but we map it to `yices-smt2` being invoked.

Thoughts?

zvonimir:
This is the right binary to be invoking. The current implementation
is invoking yices2, which does not exist when Yices2 is installed.

Fixes #374

### DIFF
--- a/Source/Provers/SMTLib/SMTLibProverOptions.cs
+++ b/Source/Provers/SMTLib/SMTLibProverOptions.cs
@@ -128,25 +128,33 @@ namespace Microsoft.Boogie.SMTLib
     {
       base.PostParse();
 
+      string SolverBinaryName = null;
       switch (Solver)
       {
         case SolverKind.Z3:
           Z3.SetDefaultOptions(this);
           SolverArguments.Add("-smt2 -in");
+          SolverBinaryName = Solver.ToString().ToLower();
           break;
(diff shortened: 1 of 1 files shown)


bkragl (approved):

Source/Provers/SMTLib/SMTLibProverOptions.cs
```diff
@@ -128,25 +128,30 @@ public override void PostParse()
     {
       base.PostParse();
 
+      string SolverBinaryName;
```
bkragl, line 131: Needs to be initialized to `null`.

zvonimir: Done.

zvonimir:
Ok, I fixed a few minor things to get it to work, and squashed the commits. Should be ready to merge. Thx!

Issue #384, opened 2021-04-21
CVC4 <-> Boogie communication broken


wrwg:
With the newest version 2.8.31, the communication between Boogie and cvc4 seems to be fully broken (previously it was broken in some cases, see #370).

The attached bpl produces output as such:

```
Prover error: Unexpected value: $Error

Prover error: Unexpected value: $Error

Prover error: Unexpected value: $Error

Prover error: Unexpected value: $Error

Advisory: $TestArithmetic_div_by_zero_u64_incorrect$verify SKIPPED because of internal error: unexpected prover output: Exception of type 'Microsoft.Boogie.SMTLib.SMTLibProcessTheoremProver+BadExprFromProver' was thrown.

...
```

Command line:
```
boogie -doModSetAnalysis -printVerifiedProceduresCount:0 -printModel:1 -enhancedErrorMessages:1 -monomorphize -proverOpt:SOLVER=cvc4 -proverOpt:PROVER_PATH=/home/wrwg/bin/cvc4 -proverOpt:O:smt.QI.EAGER_THRESHOLD=100 -proverOpt:O:smt.QI.LAZY_THRESHOLD=100 -vcsCores:12 output.bpl
```

The cvc4 used is this one:

```
https://cvc4.cs.stanford.edu/downloads/builds/minireleases/cvc4-aac53f51-x64-osx
```

(The one which is installed by default with the Move prover.)

[bug.zip](https://github.com/boogie-org/boogie/files/6353172/bug.zip)

@shazqadeer

PR #387, opened 2021-05-02
Fixes #370 and #384

### LINKED ISSUE #370: Unexpected prover response getting model when using cvc4
Symptom: Error message "Prover error: Unexpected prover response getting model" when using -proverOpt:SOLVER=cvc4

Issue:I have narrowed down the issue, and I believe the issue is in https://github.com/boogie-org/boogie/blob/master/Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs#L1737 (boogie expects cvc4 get-model output to start as (model ...) but output is just (() (define-fun ...)) (in particular, resp.Name is just "")

Additional info: The latest version of CVC4 (v1.8), `get-model` output does start with `(model....)`. This only seems to happen with the prerelease version 1.9 of cvc4. That said, I looked into SMTLIB standard and it seems that this is CVC4 becoming more SMTLIB compliant, as the standard doesn't have 'model' in the output. So the fix should be on Boogie side to support SMTLIB compliant output.

Boogie file to reproduce issue: [output_cvc4.bpl.zip](https://github.com/boogie-org/boogie/files/6305476/output_cvc4.bpl.zip)

Full command to reproduce:

```
boogie -trace -doModSetAnalysis -printVerifiedProceduresCount:0 -printModel:1 -enhancedErrorMessages:1 -monomorphize -proverOpt:SOLVER=cvc4 -proc '$TestArithmetic_div_by_zero_u64_incorrect$verify' output_cvc4.bpl
```

Versions:
CVC4 version 1.9-prerelease [git master aac53f51]
Boogie program verifier version 2.8.29.0

shazqadeer:


shazqadeer:
FYI: @wrwg, @kbansal

Issue #43, opened 2016-09-01
don't perform verification when there are 0 proof obligations


0xabu:
I have a large project that contains lots of very simple functions. I am consequently spending lots of my time waiting for things such as the following:

```
Verifying CheckWellformed$$_module.__default.NR__L1PTES ...
  [0.647 s, 0 proof obligations]  verified

Verifying CheckWellformed$$_module.__default.NR__L2PTES ...
  [0.717 s, 0 proof obligations]  verified

Verifying CheckWellformed$$_module.__default.reveal__smc__remove__premium ...
  [0.923 s, 0 proof obligations]  verified

Verifying CheckWellformed$$_module.__default.reveal__smc__mapSecure__premium ...
  [0.880 s, 0 proof obligations]  verified
```

It would be nice if Boogie could skip invoking the verifier if there are really no proof obligations.

shazqadeer:
@0xabu : Sorry it has taken almost four years to respond to this issue.  

I first educated myself on what the number of proof obligations means.  Turns out it means the total number of assert statements in the desugared code of the  implementation being verified.  By desugared code, I mean the code obtained by converting all specifications including preconditions, postconditions, loop invariants, etc. into assert and assume statements.  0 proof obligations means that there were no assert statements in the desugared code, so nothing to verify.

I am surprised by the time taken for each implementation in the output snippet shown by you.  Running times between 0.5 and 1 s are really large even if the theorem prover is invoked.  In addition to the actual invocation of SMT solver, there is time being spent in converting each implementation into a logical formula (this is where the total number of assertions is estimated, for example).  So, it is possible that each of your implementations is really large but with nothing to verify.  If so, this would be an "edge" case whose optimization might be wiser to leave to the verification tool that targets Boogie.

In any case, to take more action, we would need more investigation from your end.  In case you have moved on and are no longer interested in further investigation, let us know so I can close this issue.

0xabu:
Hi Shaz,

the verification tool here was the Dafny backend to [Vale](https://github.com/project-everest/vale), and the sources to it were [Komodo](https://github.com/Microsoft/Komodo). Since Vale is primarily a mechanical translation of text into Dafny, and then Dafny does another mechanical translation to Boogie, I think both tools lacks sufficient knowledge of what they might (or in this case might not) be asking Boogie to verify.

That said, yes, I've moved on.

RustanLeino:
[...]

PR #389, opened 2021-05-03
fixes issue #43

shazqadeer:
This PR "fixes" issue #43 using the simple approach of returning `true` as the verification condition when the assertion count is zero.  This fix will still pay the cost of VC generation when there are no assertions but will avoid the cost of converting the VC to a string and sending it to the SMT solver.


shazqadeer:
@zvonimir :  The test roundingmodes/FloatOpsWithoutRoundingMode.bpl  is failing in my PR.  I get the following output when I run the test in master.

```
Prover error: line 23 column 82: invalid number of arguments to floating point operator

Prover error: Prover died

FloatOpsWithoutRoundingMode.bpl(6,11): Verification inconclusive (foo)

```

But when I add an `assert true` at the end of `procedure foo`,  I get the following output which is different:
```
Prover error: line 22 column 94: sort mismatch, expected equal FloatingPoint sorts as arguments

Prover error: Prover died

FloatOpsWithoutRoundingMode.bpl(6,11): Verification inconclusive (foo)

```

What is the intention behind this test?  And is it ok to put `assert true` in `foo` and  change the expected output in the test so that it continues to pass?

Source/VCGeneration/VCGen.cs
```diff
@@ -637,6 +637,10 @@ public VCExpr CodeExprToVerificationCondition(CodeExpr codeExpr, List<VCExprLetB
       }
 
       CumulativeAssertionCount += assertionCount;
+      if (assertionCount == 0)
+      {
+        return VCExpressionGenerator.True;
```
RustanLeino, line 642: So, after `True` is returned, what code prevents this trivial VC from being sent to the SMT solver?

shazqadeer: The VC will be sent to the SMT solver.   My PR description may be misleading.  I only wanted to claim that the original VC is not sent to the SMT solver.  But a trivial VC will be sent.

shazqadeer:
@zvonimir :  I added an `assert true` at the beginning of foo.  This helps maintain the test.

zvonimir:
@shazqadeer I looked at what is happening in ` roundingmodes/FloatOpsWithoutRoundingMode.bpl`. This regression is testing that the prover fails when floating-points operations with wrong numbers of arguments are passed to it. which can happen if a user is not careful when using `builtin` attribute. But I think this is something we do not need a regression for, and so I suggest you remove this regression.

Source/VCGeneration/VCGen.cs
shazqadeer: @RustanLeino : Does this fix look reasonable to you?

RustanLeino (approved):

Issue #391, opened 2021-05-06
CVC4: Boogie crashes when using cvc4 with smt vectors


wrwg:
Boogie crashes with:

```
Unhandled exception. System.AggregateException: One or more errors occurred. (Index was outside the bounds of the array.) (Index was outside the bounds of the array.)
 ---> System.IndexOutOfRangeException: Index was outside the bounds of the array.
   at Microsoft.Boogie.SMTLib.SMTLibProcess.ParseId() in /home/runner/work/boogie/boogie/Source/Provers/SMTLib/SMTLibProcess.cs:line 289
   at Microsoft.Boogie.SMTLib.SMTLibProcess.ParseSExprs(Boolean top)+MoveNext() in /home/runner/work/boogie/boogie/Source/Provers/SMTLib/SMTLibProcess.cs:line 389
   at System.Collections.Generic.LargeArrayBuilder`1.AddRange(IEnumerable`1 items)
   at System.Collections.Generic.EnumerableHelpers.ToArray[T](IEnumerable`1 source)
   at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
   at Microsoft.Boogie.SMTLib.SMTLibProcess.ParseSExprs(Boolean top)+MoveNext() in /home/runner/work/boogie/boogie/Source/Provers/SMTLib/SMTLibProcess.cs:line 370
```

This is apparently model output produced by cvc4 while using smt arrays. This case works fine if using Boogie arrays instead.

Repro: 
```
boogie -doModSetAnalysis -printVerifiedProceduresCount:0 -printModel:1 -enhancedErrorMessages:1 -monomorphize -proverOpt:SOLVER=cvc4 -proverOpt:PROVER_PATH=/home/wrwg/bin/cvc4 -useArrayTheory /proverOpt:O:smt.array.extensional=false -vcsCores:12 output.bpl
```

[bug.zip](https://github.com/boogie-org/boogie/files/6432274/bug.zip)

@barrettcw

wrwg:
Opened https://github.com/CVC4/CVC4/issues/6495 and closing this one.

wrwg:
Close wrong issue, this one is still active.

PR #392, opened 2021-05-06
fixes issue #391

shazqadeer:


shazqadeer:
This PR should fix issue #391.  After the fix, I see CVC4 terminating with a parse error:

Prover error: Parse Error: <stdin>:211.151: Illegal argument detected
CVC4::ArrayStoreAll::ArrayStoreAll(const CVC4::TypeNode &, const CVC4::Node &)

  `value' is a bad argument; expected value.isConst() to hold
ArrayStoreAll requires a constant expression

It looks like Boogie is generating illegal input for CVC4.  

@cbarrettfb : Could you please provide input on what Boogie might be doing wrong?   A cursory examination of the SMT output emitted by Boogie does not reveal any problem.  If you want, I can include the SMT file here.

shazqadeer:
@wrwg: Move Prover's use of the Z3 array theory goes beyond what is supported by CVC4.  CVC4 would understand only the Select and Update operators but not the extended operators supported by Z3 (search for :builtin in the various theory/prelude files).  Also, see this issue I filed on CVC4: https://github.com/cvc5/cvc5/issues/6500.  

Although this PR fixes the crash in the model parsing code, it looks like while using CVC4 we may need to stick to Boogie arrays.

PR #398, opened 2021-06-19
Add a new sample 

shazqadeer:
This PR introduces a new sample to illustrate use of instantiation pools with a program that uses Slice/Concat operations on vectors.  The sample also also uses :split_here to increase robustness of verification and reduce chances of timeout.


shazqadeer:
@RustanLeino : FYI, I am checking in a sample that uses :split_here in a useful way.  If I don't use the :split_here annotations in the file, the running time is larger.  It would be good to run this example while testing whatever enhancements Jatin is making to :split_here.

PR #401, opened 2021-06-27
Monomorphization of procedures and implementations

shazqadeer:
This PR monomorphizes procedures and implementations by doing type instantiations similar to generic functions.  One major difference is that generic non-inlined procedures need to be verified also.  This verification is achieved by instantiating the type parameters of the procedure with fresh uninterpreted types.

This PR also moves the reusable function and procedure definitions from vector.bpl into the Boogie library for vectors. 

While doing this work, a bug was found and fixed in the generation of semantic linearity checks.  In the presence of multiple linear domains, it was hitherto possible to create a Boogie implementation with multiple basic blocks with the same name.


Source/Concurrency/LinearTypeChecker.cs
```diff
@@ -921,12 +921,12 @@ private class LinearityCheck
...
       public string name;
 
-      public LinearityCheck(Expr assume, Expr assert, string message, string name)
+      public LinearityCheck(LinearDomain domain, Expr assume, Expr assert, string message, string name)
       {
         this.assume = assume;
         this.assert = assert;
         this.message = message;
-        this.name = name;
+        this.name = $"{domain.domainName}_{name}";
```
bkragl, line 929: Maybe store `domain` (or just `domain.domainName`) in `LinearityCheck` and compute the label where the block is created.

bkragl (commented):
The bugfix in the linear type checker looks good. Just one small comment below.

Source/Concurrency/LinearTypeChecker.cs
shazqadeer: Done.

PR #403, opened 2021-06-28
Update to .NET 5.0

### LINKED ISSUE #402: Boogie on .NET 5.0 ?
I couldn't find a duplicate of this, though I figure it has been asked before.  Currently, it seems like you need `dotnet-runtime-3.1` to run Boogie on Linux.  Otherwise, you end up with an error message like this:

```
It was not possible to find any compatible framework version
The framework 'Microsoft.NETCore.App', version '3.1.0' was not found.
  - The following frameworks were found:
      5.0.7 at [/usr/share/dotnet/shared/Microsoft.NETCore.App]

You can resolve the problem by installing the specified framework and/or SDK.
```

It seems to me that modern Linux distro's support .NET 5.0 and it seems to generally be the default choice.  So, I'm always either manually installing `dotnet-runtime-3.1` or asking someone to do it for me.  Anyway, I don't know what the trade-offs are here.

bkragl:
Closes  #402

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,7 +24,7 @@ jobs:
       - name: Setup dotnet
         uses: actions/setup-dotnet@v1
         with:
-          dotnet-version: 3.1
+          dotnet-version: '5.0.x'
       - name: Checkout Boogie
         uses: actions/checkout@v2
         with:
--- a/Source/AbsInt/IntervalDomain.cs
+++ b/Source/AbsInt/IntervalDomain.cs
@@ -320,11 +320,7 @@ namespace Microsoft.Boogie.AbstractInterpretation
 
     static Expr NumberToExpr(BigInteger n, Type ty)
     {
-      if (n == null)
-      {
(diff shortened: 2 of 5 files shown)


shazqadeer (approved):

shazqadeer:
Does the Boogie upgrade to 5.0 mean that Dafny will have to be upgraded to 5.0?

bkragl:
> Does the Boogie upgrade to 5.0 mean that Dafny will have to be upgraded to 5.0?

Dafny is already on .NET 5.0.

shazqadeer:
@akashlal :  Any concerns?

akashlal:
Sounds good to me. No concerns.

zvonimir:
Is Corral going to switch too? Or maybe it already did? @akashlal? Thx!

akashlal:
@zvonimir Corral hasn't switched yet, but we should do it after boogie moves.

shazqadeer:
@bkragl :  Merge?

PR #404, opened 2021-07-01
Fix the issue when using cvc5 as backend.

Ying1123:
This pr change the` Logic = "ALL_SUPPORTED";` to `Logic = "ALL";` which is consistent with SMT-LIB standard.
Starting from CVC4 1.8 https://github.com/cvc5/cvc5/pull/6664, CVC4/CVC5 no longer support the syntax ALL_SUPPORTED.


Source/Provers/SMTLib/SMTLibProverOptions.cs
```diff
@@ -139,7 +139,7 @@ public override void PostParse()
         case SolverKind.CVC4:
           SolverArguments.Add(
             "--lang=smt --no-strict-parsing --no-condense-function-values --incremental --produce-models");
-          if (Logic == null) Logic = "ALL_SUPPORTED";
+          if (Logic == null) Logic = "ALL";
```
shazqadeer, line 142: @Ying1123 :  This looks great.  Could you please add a comment about the version of CVC5 which is needed for this change. I am also wondering if we should either 
(1) rename SolverKind.CVC4 to SolverKind.CVC5, or
(2) add a different option called SolverKind.CVC5 keeping the original option unchanged.

Do you have any advice?

@bkragl, @RustanLeino , @zvonimir  :  FYI

Ying1123: I have added.

shazqadeer: @Ying1123:   Is there any reason to support both CVC4 and CVC5 in Boogie?  If note, I recommend renaming SolverKind.CVC4 to SolverKind.CVC5 and the command-line option string to "cvc5". If you search for CVC4 and "cvc4" in the Boogie code, you should be able to find all occurrences.

Ying1123: I think there is no reason to have both. And I can change it into cvc5.

Source/Provers/SMTLib/SMTLibProverOptions.cs
```diff
@@ -166,8 +166,8 @@ public override string Help
           @"
 SMT-specific options:
 ~~~~~~~~~~~~~~~~~~~~~
-SOLVER=<string>           Use the given SMT solver (z3, cvc4, yices2; default: z3)
-LOGIC=<string>            Pass (set-logic <string>) to the prover (default: empty, 'ALL_SUPPORTED' for CVC4)
+SOLVER=<string>           Use the given SMT solver (z3, cvc5, yices2; default: z3)
+LOGIC=<string>            Pass (set-logic <string>) to the prover (default: empty, 'ALL_SUPPORTED' for CVC5)
```
shazqadeer, line 170: ALL_SUPPORTED -> ALL

shazqadeer (approved):

PR #407, opened 2021-07-05
Fix augmented trace

zvonimir:
Augmented trace is set to null when Corral is used.


shazqadeer:
Looks good.  Great catch on the file name.

PR #408, opened 2021-07-06
Fix ModelParser error for not setting the else path

arminvakil:
The ModelParser does not populate the else path for the boogie functions that only have an else path such as the following:
`$Is -> {`
`  else -> true`
`}`
This is because `fn` is not created at the time the else path is parsing, and as a result, else path will not be set.
Moving the else path check after the `if (fn == null)` block will fix the issue and else path will be set.

### DIFF
--- a/Source/Model/ModelParser.cs
+++ b/Source/Model/ModelParser.cs
@@ -298,7 +298,10 @@ namespace Microsoft.Boogie
 
               if (tuple0 == "else")
               {
-                if (fn != null && !(resultName is string && ((string)resultName) == "#unspecified") && fn.Else == null)
+                if (fn.Else != null) {
+                  BadModel("multipe else cases");
+                }
+                if (!(resultName is string && ((string) resultName) == "#unspecified"))
                 {
                   fn.Else = GetElt(resultName);
                 }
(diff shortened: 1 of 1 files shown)


Source/Model/ModelParser.cs
```diff
@@ -305,6 +296,15 @@ internal override void Run()
                 }
               }
 
+              if (tuple0 == "else")
+              {
+                if (fn != null && !(resultName is string && ((string)resultName) == "#unspecified") && fn.Else == null)
```
shazqadeer, line 301: If it is guaranteed that fn != null at this line, then fn != null should be removed from the conditional.

shazqadeer: Also, if it turns out that tuple0 == "else" holds, then shouldn't it be guaranteed that fn.Else == null?  I suppose that there would be exactly one (or perhaps at most one) "else" in the model for a function.

arminvakil: You are right. fn != null should be removed.
There should only be one `else` case in the boogie model. A check can be added to throw an exception if there are multiple `else` statements in the model. I will add this, shortly.

shazqadeer (approved):
Looks good!

PR #409, opened 2021-07-09
[Civl] Fixes in attribute propagation during transition relation computation

shazqadeer:
- propagate attributes on variables throughout transition relation computation
- add more pool hints to the paxos proof

### DIFF
--- a/Source/Concurrency/TransitionRelationComputation.cs
+++ b/Source/Concurrency/TransitionRelationComputation.cs
@@ -78,18 +78,6 @@ namespace Microsoft.Boogie
       }
     }
 
-    private static QKeyValue CloneAttributes(QKeyValue kv, Dictionary<Variable, Variable> substMap)
-    {
-      if (kv == null)
-      {
-        return null;
-      }
-      return new QKeyValue(kv.tok, kv.Key,
-        kv.Params.Select(p => p is Expr ? SubstitutionHelper.Apply(substMap, (Expr) p) : p).ToList(),
-        CloneAttributes(kv.Next, substMap));
-    }
-    
-    
     private static Expr ComputeTransitionRelation(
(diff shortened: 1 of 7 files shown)


Source/Concurrency/TransitionRelationComputation.cs
```diff
@@ -484,7 +473,10 @@ private void AddBoundVariablesForRemainingVars()
         var remainingVars = NotEliminatedVars.Except(IntermediateFrameWithWitnesses);
         existsVarMap = remainingVars.ToDictionary(v => v, v => (Variable) VarHelper.BoundVariable(v.Name, v.TypedIdent.Type));
         var substMap = remainingVars.ToDictionary(v => copyToOriginalVar[v], v => existsVarMap[v]);
-        existsVarMap.Iter(kv => kv.Value.Attributes = CloneAttributes(copyToOriginalVar[kv.Key].Attributes, substMap));
+        existsVarMap.Iter(kv =>
+        {
+          kv.Value.Attributes = copyToOriginalVar[kv.Key].Attributes;
+        });
```
bkragl, line 479: I'm a bit surprised that we can drop the substitution that `CloneAttributes` did. Why is that?
(Looks like `substMap` is unused now.)

shazqadeer: In an earlier version of pool-based QI, the pool annotation on a bound variable had an expression. In the current design, there is no expression and instead there is just the pool name.  The expressions are all in the body of a quantifier or lambda or on a predicate command.  So I cleaned up the code to avoid giving the impression that the substitution matters.  If we introduce something fancier later on that requires the substitution, we can bring it back in.

I will delete the initialization of substMap; indeed, substMap is not used any more.

PR #414, opened 2021-07-15
Default arity value for functions with a single else path

Dargones:
This issue is closely related to #408 and concerns cases when a function only has an else path, which are numerous when translating Dafny to Boogie and using the `/proverOpt:O:model.completion=true` command line argument.

When a function only has an else path, its arity is currently set to 1 by default, even if the actual arity is different. This is a problem for two reasons: 

- arity is supposed to be immutable, so any attempt to use such a function later on will lead to a failure. In particular, [DafnyServer](https://github.com/dafny-lang/dafny/blob/master/Source/DafnyServer/boogie/DafnyProvider.cs) has a call to `MkFunc("Seq#Index", 2)` which fails if the original model only had the else branch for this function. I propose a fix (first commit) that allows MkFunc to change the arity of the function if no function applications were previously recorded.
- the names of `MapType1Store`, `MapType2Select`, etc. depend on their arity. When these function only have an else path, their names clash and this leads to "multiple else cases" exceptions. Incidentally, their names can also clash with Dafny's `[2]` function. The long-term fix would be to stop renaming these function but this can potentially lead to significant consequences downstream so I propose to remove the "multiple else cases" exception instead (second commit). This will at least stop the parser failing when processing a model generated from Dafny code with `/proverOpt:O:model.completion=true` and should not introduce any issues that were not already present.

### DIFF
--- a/Source/Model/Model.cs
+++ b/Source/Model/Model.cs
@@ -659,16 +659,24 @@ namespace Microsoft.Boogie
     public Func MkFunc(string name, int arity)
     {
       Func res;
+      Element elseBranch = null;
       if (functionsByName.TryGetValue(name, out res))
       {
-        if (res.Arity != arity)
-          throw new ArgumentException(string.Format(
-            "function '{0}' previously created with arity {1}, now trying to recreate with arity {2}", name, res.Arity,
-            arity));
-        return res;
+        if (res.Arity != arity) {
+          if (res.apps.Count > 0) {
+            throw new ArgumentException(string.Format(
+              "function '{0}' previously created with arity {1}, now trying to recreate with arity {2}", name, res.Arity,
+              arity));
+          }
+          elseBranch = res.Else;
+          functions.Remove(res);
+          functionsByName.Remove(res.Name);
+        } else 
+          return res;
       }
 
       res = new Func(this, name, arity);
+      res.Else = elseBranch;
       functionsByName.Add(name, res);
       functions.Add(res);
       return res;
--- a/Source/Model/Model.cs
+++ b/Source/Model/Model.cs
@@ -259,26 +259,27 @@ namespace Microsoft.Boogie
     {
       public readonly Model Model;
       public readonly string Name;
-      private int? TrueArity; // can be null if arity is unknown
+      private int? arity; // can be null if arity is unknown
       internal readonly List<FuncTuple> apps = new List<FuncTuple>();
 
-      public virtual int Arity {
+      /// <summary>
+      /// Arity can be set to null if it is not known at the time of creation
+      /// Once set to an integer, however, an attempt to change arity leads
+      /// to an exception
+      /// </summary>
+      public virtual int? Arity {
         get {
-          return TrueArity ?? 1;  // return 1 if arity is unknown
+          return arity; 
         }
         set {
-          if ((TrueArity != null) && (TrueArity != value)) {
+          if ((arity != null) && (value != null) && (arity != value)) {
             throw new ArgumentException(string.Format(
               "function '{0}' previously created with arity {1}, " +
-              "is now being assigned arity {2}", Name, TrueArity,
+              "is now being assigned arity {2}", Name, arity,
               value));
           }
-          TrueArity = value;
+          arity = value;
         }
-      }
-
-      public bool ArityIsKnown() {
-        return TrueArity != null;
       }
 
       public IEnumerable<FuncTuple> Apps
@@ -297,12 +298,13 @@ namespace Microsoft.Boogie
       {
         Model = p;
         Name = n;
-        TrueArity = a;
+        arity = a;
       }
 
-      public override string ToString()
-      {
-        return string.Format("{0}/{1}", Name, Arity);
+      public override string ToString() {
+        if (Arity != null) 
+          return string.Format("{0}/{1}", Name, Arity);
+        return string.Format("{0}/{1}", Name, "?");
       }
 
       internal void Substitute(Dictionary<Element, Element> mapping)
@@ -330,10 +332,10 @@ namespace Microsoft.Boogie
         }
       }
 
-      public void SetConstant(Element res)
-      {
-        if (Arity != 0 || apps.Count > 0)
+      public void SetConstant(Element res) {
+        if (apps.Count > 0)
           throw new ArgumentException();
+        Arity = 0; // will throw exception if arity is not null or already 0
         var t = new FuncTuple(this, res, null);
         apps.Add(t);
         res.references.Add(t);
@@ -398,10 +400,8 @@ namespace Microsoft.Boogie
       /// <summary>
       /// For a nullary function, return its value.
       /// </summary>
-      public Element GetConstant()
-      {
-        if (Arity != 0)
-          throw new ArgumentException();
+      public Element GetConstant() {
+        Arity = 0; // will throw exception if arity is not null or already 0
         if (apps.Count == 0)
         {
           // value missing in the returned model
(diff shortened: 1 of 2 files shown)


zvonimir:
@shazqadeer Hi Shaz, could you please review this one? Thx!

shazqadeer:
@Dargones :  Thank you for the PR.  I understand your first bullet.  But I do not understand your second bullet.  Where are MapType1Store and MapType2Select coming from?  What is Dafny's [2] function?

I am happy to take your first commit.  But before accepting the second commit (the one that eliminates the multiple else cases exception), I would like to understand more.

Dargones:
Thank you, @shazqadeer ! My second point is related to lines 272-297 in [ModelParser](https://github.com/boogie-org/boogie/blob/master/Source/Model/ModelParser.cs). These lines are responsible for renaming a select set of functions. So, for instance, the name of any function that matches the `^MapType[0-9]*Select$` regex pattern is replaced with `string.Format("[{0}]", arity)`. Because the default arity for functions with only the else branch is `1`, all of `MapType1Select`, `MapType2Select`, etc. are renamed to `[1]` when they don't have any applications. Consequently, this leads to name clashes and to the "Multiple else branches" exception. 

The best long-term solution to get around this would be, I think, to stop renaming any functions, but I am afraid that doing so could break code that depends on the parser. This would also require setting the values inside the `selectFunctions` and `storeFunctions` dictionaries (see lines 281 and 290) in some new way. I am not sure what some of these functions are used for, so I am cautious about refactoring this code in such a significant manner.

The short-term fix would, therefore, be to remove the `Multiple else branches` exception. This will not stop the name clashes but it will stop the parser from failing immediately after encountering one. 

Dafny's `[2]` function, which can clash with any of the `MapType[0-9]*Select` functions with the arity of `2`, maps a heap and a reference to an object in that heap to an object itself.

shazqadeer:
@michalmoskal , @RustanLeino :   Do either of you understand the reason for the renaming (see the last comment by @Dargones) in the ModelParser?  Input much appreciated.

shazqadeer:
> My second point is related to lines 272-297 in ModelParser. These lines are responsible for renaming a select set of functions. So, for instance, the name of any function that matches the ^MapType[0-9]*Select$ regex pattern is replaced with string.Format("[{0}]", arity). Because the default arity for functions with only the else branch is 1, all of MapType1Select, MapType2Select, etc. are renamed to [1] when they don't have any applications. Consequently, this leads to name clashes and to the "Multiple else branches" exception.

This diagnosis is puzzling to me.  The arity used in the renaming is not coming from the default arity of a function.  Rather, it is coming from line 274 which is reached only when processing a particular application.  So, at this point the true arity appears to be known.

shazqadeer:
@Dargones :  To make progress on this PR, I would like to request that you split it into two PRs, one for each commit.

For the first commit, I am ready to get it landed but I would like some changes.  I would like the field Arity of class Func to be int? rather than int to clearly record that the arity is not available.  This will allow the unknown arity to be recorded as null and we can still have a tight assertion on update.

For the second commit, please take a look at my questions above.  I am happy to accommodate your request if it is somehow critically urgent to land this fix (please explain if it is so).  Otherwise, let us discuss a bit more.

I want to note here that I am truly appreciative of the time you are taking to make Boogie better for all of us.

Dargones:
@shazqadeer : thank you! I will split the PR in two as you suggest. The problem with line 274 is that, when there are no applications, it processes the else branch as if it were an application, which sets `arity` to one (because the tuple being processed is ["else", "->", "{SOMETHING}"], which has length 3). This, in turn, causes the like renaming of different functions, leads to a name clash, and, finally, to a "multiple else cases" exception.

Here are the concrete steps to reproduce this bug:
1) Create a `test.dfy` containing the following:
`method a() { assert false;}`
2) Use the following to generate a model for the failing assertion: 
`dafny /proverOpt:O:model.completion=true /printModel:1 /printModelToFile:model.txt test.dfy`
3) Attempt to parse the model. The easiest way to do this is to insert the following line at the top of BoogieDriver's Main, compile boogie, and run it passing the path to`model.txt` as argument:
`Model.ParseModels(new StreamReader(args[0]));`

The reason for my pull request is that I am working on a counterexample feature for Dafny and need the parser to succeed when processing models generated with the `/proverOpt:O:model.completion=true` flag.

zvonimir:
@Dargones Could you provide a regression in Boogie (without Dafny)?
Shaz is not using Dafny I don't think.

Dargones:
@zvonimir @shazqadeer Apologies! here is the same example in Boogie:

1) Download [this file](https://gist.github.com/Dargones/b57d253ad74b8b23b3d06373557cc878). It is generated by Dafny but it is a valid Boogie program. I am afraid I can't immediately reduce this to a simpler example as I don't fully understand what the MapType.. functions are used for or why they are being renamed.
2) Run the following to generate a model for the failing assertion.:
`boogie /proverOpt:O:model.completion=true /printModel:1 /printModelToFile:model.txt test.bpl`
3) Attempt to parse the model. The easiest way to do this is to insert the following line at the top of BoogieDriver's Main, compile boogie, and run it passing the path tomodel.txt as argument:
Model.ParseModels(new StreamReader(args[0]));

Dargones:
I reverted the commit that removed the exception for this PR and made the changes you asked for regarding the first commit. Unfortunately, I don't think there is a way to make `Arity` a nullable int directly. This is because it should remain public and readonly to the client but must also be internally modifiable for the purposes of changing it from null to a fixed value. What I did is make `Arity` into a public property of type `int`. The `get` method returns `1` if arity is not known to preserve compatibility and the `set` method makes the property effectively readonly once its value is set to anything but null. The `ArityIsKnown` method can be used to check if the arity has been set to an integer or is still null.

RustanLeino:
Can the code be changed so that it sets the arity of the select/store functions correctly? The arity should be evident from the name. For example, `MapType2Select` (aka `[2]` and aka the newly proposed `[,]`) has arity 2.

shazqadeer:
> Can the code be changed so that it sets the arity of the select/store functions correctly? The arity should be evident from the name. For example, MapType2Select (aka [2] and aka the newly proposed [,]) has arity 2.

@RustanLeino : Based on my reading of the code that implements type encoding, the functions MapType[0-9]*Select and MapType[0-9]*Store the part [0-9]* is not the arity but some number derived from the polymorphic context.  For example, when I was debugging, I discovered that both MapType0Select and MapType1Select have arity 3.  In short, the whole renaming scheme is thoroughly busted.

I will put up a PR to fix the problem now that I have input from you that this is not a Dafny-specific hack.  But a question still remains:  if a certain name was desired why not just use that name when the SMT query is constructed in the first place?  If we were to do this, then no renaming would be required.

Source/Model/Model.cs
```diff
@@ -259,9 +259,28 @@ public class Func
...
+            throw new ArgumentException(string.Format(
+              "function '{0}' previously created with arity {1}, " +
+              "is now being assigned arity {2}", Name, TrueArity,
+              value));
+          }
+          TrueArity = value;
+        }
+      }
+
+      public bool ArityIsKnown() {
```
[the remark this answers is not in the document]
Dargones: Thank you! I have implemented the changes you suggested (see my latest commit). My only concern is that exposing the null value of the arity is likely to break client code. So, for instance, this change already required a small fix to ModelParser (although this is not client code and hence not that much of a problem).

Source/Model/ModelParser.cs
```diff
@@ -308,7 +310,8 @@ internal override void Run()
                 continue;
               }
 
-              var args = new Model.Element[fn.Arity];
+              fn.Arity = tuple.Count - 2;
+              var args = new Model.Element[fn.Arity ?? 1]; // arity cannot be null here
```
shazqadeer, line 314: If Arity cannot be null, don't you want new Model.Element[fn.Arity]?

shazqadeer (approved):

Source/Model/ModelParser.cs
shazqadeer: You can cast (int) fn.Arity.

Dargones: Right, thank you! Not sure why I did it that way. Fixed that in the new commit.

PR #415, opened 2021-07-16
Make prover death a hard failure

shazqadeer:
Many errors are possible when Boogie interacts with a prover.  Currently, all of these errors are lumped as "correct" output, this output is dumped on stdout, and Boogie exits with code 0.  This behavior is not helpful for tools that use Boogie as a subroutine in their toolchain.  These tools would rather use the exit code to detect that something bad has happened. This PR takes one step towards improving this behavior by resulting in exit code 1 in case the prover dies while interacting with Boogie.

### DIFF
--- a/Source/BoogieDriver/BoogieDriver.cs
+++ b/Source/BoogieDriver/BoogieDriver.cs
@@ -62,7 +62,7 @@ namespace Microsoft.Boogie
 
       Helpers.ExtraTraceInformation("Becoming sentient");
 
-      ExecutionEngine.ProcessFiles(fileList);
+      var success = ExecutionEngine.ProcessFiles(fileList);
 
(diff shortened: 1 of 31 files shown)


zvonimir (approved):
Looks good to me.

zvonimir:
The implementation looks good to me, but you have to update two regressions that are now failing.
This might break dependencies of users that somehow rely on Boogie always returning 0. I would expect those to be super rare, if there are even any.
Good improvement.

shazqadeer:
@zvonimir :  Thank you for looking at this PR.  This other PR #416 needs to land first.  Once that one lands, the remaining test failure in this PR will be automatically fixed without doing any further changes here.  Could you please review #416 ?  Thanks.

PR #416, opened 2021-07-17
Add checking for inline and define attributes on functions

shazqadeer:
This PR tightens the type checking of function declarations with respect to attributes and how they call each other.

1. :inline attribute must have no parameter or a Boolean parameter
2. :inline attribute on a function requires function to have a body
3. function call graph projected onto functions with :inline or :define attribute must be acyclic

### DIFF
--- a/Source/Core/Absy.cs
+++ b/Source/Core/Absy.cs
@@ -3189,8 +3189,6 @@ namespace Microsoft.Boogie
       otherDefinitionAxioms.Add(axiom);
     }
 
-    public bool doingExpansion;
-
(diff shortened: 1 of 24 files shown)


zvonimir (changes requested):
It would be good to have a regression that fails with `Inlined function must have a body`.
Is there one such regression already there?
I did not see it, but maybe I just missed it.

shazqadeer:
My latest commit adds the tests you suggested.

zvonimir (approved):
Looks good to me.

PR #417, opened 2021-07-19
Remove multiple else cases exceptions

Dargones:
This is a continuation of #414 (thank you for all the help with it!) made into a separate PR as requested. The short description is that I propose to remove the "Multiple Else Cases" exception because it makes the model parser fail on a relatively large set of models. The full explanation is below:

The current version of the [ModelParser](https://github.com/boogie-org/boogie/blob/master/Source/Model/ModelParser.cs) renames any function matching the `^MapType[0-9]*Select$` and `^MapType[0-9]*Store$` regex patterns. The new name assigned to such a function depends on its arity. When a function has no application, its arity is computed from the else branch as if the latter were an application. The result is that the arity is set `1` (or `null` if #414 is adopted). Combined with the renaming, this leads to a name clash when two or more such functions only have an else branch. A name clash leads to a “Multiple else cases” exception down the road (because an else branch is assigned more than once). 

 It is not clear to me if one can stop the renaming without breaking the code that might depend on it. Neither is it clear how to compute the true arity when there is only the else branch. As such, I can’t think of an immediate way to avoid the name clash. The fix proposed here is, therefore, to remove the recently introduced “Multiple else cases” exception. Once removed, it will stop the parser from failing to process models generated with the `/proverOpt:O:model.completion=true` command line argument from certain Boogie programs (most notably almost all programs translated from Dafny to Boogie). This command line argument is critical to implementing a feature I am working on for Dafny.

Guide for reproducing the error in question:

1) Download [this file](https://gist.github.com/Dargones/b57d253ad74b8b23b3d06373557cc878). It is generated by Dafny but it is a syntactically valid Boogie program (which fails verification). I am afraid I can't immediately reduce this to a simpler example as I don't fully understand what the MapType.. functions are used for or why they are being renamed. The source Dafny code is this: 
`method a() {assert false;}`
2) Run the following to generate a model for the failing assertion.:
`boogie /proverOpt:O:model.completion=true /printModel:1 /printModelToFile:model.txt test.bpl`
3) Attempt to parse the model. The easiest way to do this is to insert the following line at the top of BoogieDriver's Main, compile boogie, and run it passing the path to `model.txt` as argument:
`Model.ParseModels(new StreamReader(args[0]));`

Thank you!

### DIFF
--- a/Source/Model/ModelParser.cs
+++ b/Source/Model/ModelParser.cs
@@ -298,10 +298,7 @@ namespace Microsoft.Boogie
 
               if (tuple0 == "else")
               {
-                if (fn.Else != null) {
-                  BadModel("multiple else cases");
-                }
-                if (!(resultName is string && ((string) resultName) == "#unspecified"))
+                if ((fn.Else == null) && (!(resultName is string && ((string) resultName) == "#unspecified")))
                 {
                   fn.Else = GetElt(resultName);
                 }


shazqadeer:
@Dargones :  The functions that match the regular expressions ^MapType[0-9]*Select$ and ^MapType[0-9]*Store$ are being generated in type encoding performed by Boogie.   To see where, search for "Select" and "Store" in the files Source/VCExpr/TypeErasureArguments.cs and Source/VCExpr/TypeErasurePremisses.cs.  My suspicion is that the renaming of the function is some Dafny-specific hack that found its way into the model parsing in Boogie.  I would prefer it if the model parsing in Boogie does not rename the function and Dafny does the renaming appropriately at its end.  You could disable the renaming in Boogie and see what breaks in Dafny.  At my end, I will disable the renaming and see if any tests break in Boogie.

shazqadeer:
@RustanLeino :  Please chime in.

shazqadeer:
I disabled renaming and ran Boogie tests.  No change.

RustanLeino:
The renaming is not Dafny-specific, but it was meant to give names that look closer to Boogie syntax. The names have the form `[1]` and `[1:=]` for 1-dimensional maps and, for example, `[3]` and `[3:=]` for 3-dimensional maps. (It occurs to me now that names like `[]`, `[:=]`, `[,,]`, and `[,,:=]` would have been even closer to Boogie syntax.)

However, there shouldn't be any duplicates of these.

zvonimir:
This is super helpful @RustanLeino.
I like the suggestion for the new syntax. Maybe we should switch to that.
Does Boogie have regressions that check models? If so, then it would be great to add a regression that checks this model generation path.

RustanLeino:
Boogie's model output used to be used by the Boogie Verification Debugger (BVD), but I don't think BVD has been maintained. (There had been a Dafny plug-in for BVD, but we abandoned it when Dafny stopped supporting the full Visual Studio IDE.)

Dafny's VS Code IDE uses models in some way, but I don't know the details. I don't know what other tools use Boogie's model output.

RustanLeino:
Deciding on what the names of these functions should be and adding Boogie tests would be great!

shazqadeer:
Coalescing multiple functions into one seems strange to me.  The SMT solver would generate separate models for each function.  If they are coalesced, what would the resulting model of the coalesced function be?

zvonimir:
@shazqadeer here is an example I just got from @Dargones. Note that for the purpose of this project, he needs to invoke the prover with the following argument `/proverOpt:O:model.completion=true`, which generates complete models. It might be that this bug was not exposed earlier because nobody has been extensively using complete models.
Anyhow, once we do this, here is an excerpt of a model we get:
```
MapType0Select -> {
  else -> T@U!val!0
}
...
MapType1Select -> {
  else -> T@U!val!0
}
```
Since the arity of these functions cannot be determined, on this line https://github.com/boogie-org/boogie/blob/master/Source/Model/ModelParser.cs#L279 it will be set to 1 (hardcoded) for both of them.
So they will both end up mapping to the same name `[1]`.
In turn, this will trigger an exception here: https://github.com/boogie-org/boogie/blob/master/Source/Model/ModelParser.cs#L304

More generally, as you already mentioned elsewhere, the number in the name of these functions has nothing to do with arity. That is my understanding as well. So then, as you also mentioned elsewhere, even if the arity of two functions, say MapType0Select and MapType1Select, can be determined, it could be for example 3 in both cases. Then, this code will replace both of these functions with `[3]`. Given that the code is using these maps to check whether an arity already exists, it seems that replacing multiple functions from the model with the same `[arity]` name was the intended behavior.

I hope this helps to get to the bottom of this. Thoughts?

Personally, the more I think about this, I am wondering if this model post-processing code should not be here in the first place. In its current implementation it is obscuring what is actually happening in the model.

zvonimir:
Another quick comment. It seems that when there is no polymorphism in a program, these maps with names such as `MapType0Select` do not even get generated in the first place. Which makes the purpose of this code even more unclear to me.

shazqadeer:
> Another quick comment. It seems that when there is no polymorphism in a program, these maps with names such as MapType0Select do not even get generated in the first place. Which makes the purpose of this code even more unclear to me.

As I explained in one of my comments, the source of these pesky function names is type encoding, which will of course not happen for monomorphic programs.  However, Dafny uses polymorphic Boogie for which type encoding is relevant.

shazqadeer:
Here is my proposal to fix the problem:

- Encode the arity of the function in the name itself.  So, for example, MapType0Select with arity 3 would be named MapType0Select3.  This will take the guesswork out of generating the correct arity during renaming.
- The renaming should use the pair of numbers in the name to create unique function names.  The selectFunctions and storeFunctions dictionaries would be keyed by the number pair.  If everybody loves the naming suggestion by @RustanLeino above, we could go with `<0>[]` for MapType0Select with arity 1, `<42>[,,]` for MapType42Select with arity 3, <67>[:=] for MapType67Store with arity 1, and `<0>[,,:=]` for MapType0Store with arity 3.  The purpose should be clearly explained with a comment so the reader does not go crazy.

shazqadeer:
Finally, I want to note that I am dubious about the value of the renaming.  If everybody loves the renaming to make the function name nicer, I am fine.  But I want to point out that there is this cost of some unrelated complexity in the model parsing code.

zvonimir:
Exactly. This is what I was just about to write, and what I already mentioned above.
I doubt that this renaming is helping much in making models more readable. I would guess that only hard-core experts look at the models in the first place, and for such people I would again guess that renaming is probably more of a nuisance than help, since the parsed model does not match the generated VC any more in terms of map names.
My suggestion is that we remove this renaming business, if @RustanLeino is ok with that. It seems that there are no clients at this point that rely on it.

Dargones:
Just checked that Boogie calls ModelParser **before** printing out the model to file. This means that turning off the renaming in Boogie will necessitate a fix to Dafny [here](https://github.com/dafny-lang/dafny/blob/0fa9f99df7d91bc059b63f2d39c44baca958ba67/Source/DafnyServer/boogie/DafnyProvider.cs#L47-L48) (I mentioned before that I wasn't sure how the `[2]` function in Dafny is related to the issue at hand but I am now certain that it is generated during the renaming). Since I am already working on this part of Dafny Server and will be patching and augmenting this code anyway, I will be willing to resolve any issues that arise from ceasing the renaming process. Other than that, I too think that stopping the renaming might be optimal.

Dargones:
Closing this PR because #418 fixes the problem. Thank you @RustanLeino @zvonimir @shazqadeer for all your help with the issue!

PR #418, opened 2021-07-21
Removed function renaming in model parsing

shazqadeer:
Resolves PR #417 which should be discarded.

### DIFF
--- a/Source/Model/ModelParser.cs
+++ b/Source/Model/ModelParser.cs
@@ -174,8 +174,6 @@ namespace Microsoft.Boogie
 
     internal override void Run()
     {
-      var selectFunctions = new Dictionary<int, Model.Func>();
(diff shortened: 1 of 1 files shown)


shazqadeer:
FYI: @Dargones

Dargones:
Just checked and this fixes the issue that I was having, thank you, @shazqadeer ! I do believe this will necessitate a fix to Dafny [here](https://github.com/dafny-lang/dafny/blob/0fa9f99df7d91bc059b63f2d39c44baca958ba67/Source/DafnyServer/boogie/DafnyProvider.cs#L47-L48) but I will be working on that file anyway.

zvonimir (approved):
Looks good to me.

zvonimir:
Thanks @shazqadeer! This just unblocked us.

PR #419, opened 2021-07-24
Fix inconsistency implementation of Vec library

shazqadeer:
- @cbarrettfb detected an inconsistency in the implementation of the Vec library.  This PR fixes the problem by rewriting the buggy definitions of vector operations in LibraryDefinitions.bpl
- While doing this work, some bugs were discovered in monomorphization and pool-based QI; they were fixed.


shazqadeer:
@DavidLDill: FYI

PR #423, opened 2021-07-26
Split redesign

typerSniper:
The `split_here` attribute (on predicate commands like asserts, assumes) is used for verifying parts of an implementation separately (but not independently).  This PR does so as follows:
Picture a Boogie implementation at the stage where all the back-edges are gone and all the specs have been inlined as asserts and assumes. The PR "checks" every assert with the closest `split_here` command that dominates it. If there is none, then the assert is "checked" with a default split that is unique and consists of all asserts that aren't dominated by a `split_here`. To do so, I assign every assert to some split and turn all other asserts to assumes. Thus a given split only checks its assigned assertions; others are assumed. 

This makes the checking separate but not really independent --- the assumes can influence Z3's decisions still. To curb this, I clean up each split and remove all assumptions that do not have a successor assert.


Source/VCGeneration/VCGen.cs
```diff
@@ -28,6 +28,25 @@ public VCGen(Program program, string /*?*/ logFilePath, bool appendLogFile, List
       this.logFilePath = logFilePath;
     }
 
+    public static Graph<Block> BlocksToDag(List<Block> blocks)
```
shazqadeer, line 31: It is a great service to implement a method for creating a Graph from blocks.  But if the list of blocks has cycles, the output would not be a DAG.  This should not matter for this method whose only job is to create a Graph.  Please rename the method.

Also, there are places in the Boogie code where the same computation is reimplemented. It would be great to refactor the code so that all those places target this one method that you have introduced.  See, for example, the  method ComputeLiveVariables in DeadVarElim.cs.

shazqadeer: You might also consider splitting your PR into two PRs, one for the block->graph business and one for the splitting business.

shazqadeer (commented):
@typerSniper : Thanks a lot for the excellent work on this PR.  I have left a few comments.

Source/VCGeneration/Split.cs
```diff
@@ -868,6 +1036,65 @@ public Counterexample ToCounterexample(ProverContext context)
...
+          bool isAssumeFalse (Cmd c) { return c is AssumeCmd ac && ac.Expr is LiteralExpr le && !le.asBool; }
+          var firstFalseIdx = b.Cmds.FindIndex(c => isAssumeFalse(c));
+          if (firstFalseIdx != -1)
+          {
+            b.Cmds = b.Cmds.Take(firstFalseIdx + 1).ToList();
+            b.TransferCmd = (b.TransferCmd is GotoCmd) ? new ReturnCmd(b.tok) : b.TransferCmd;
+          }
+        }
+
+        void DeleteSubSumptions (Block b)
```
RustanLeino, line 1052: Don't capitalize the second `S` in `Subsumptions`

Source/VCGeneration/Split.cs
```diff
@@ -868,6 +1036,65 @@ public Counterexample ToCounterexample(ProverContext context)
...
+        {
+          bool isAssumeSubsumption (Cmd c) { return c is AssumeCmd ac && QKeyValue.FindBoolAttribute(ac.Attributes, "subsumption"); }
+          b.Cmds = b.Cmds.Where(c => !isAssumeSubsumption(c)).ToList();
+        }
+
+        bool ContainsAssert(Block b)
+        {
+          bool isNonTrivialAssert (Cmd c) { return c is AssertCmd ac && !(ac.Expr is LiteralExpr le && le.asBool); }
+          return b.Cmds.Aggregate(false, (containsAssert, c) => containsAssert || isNonTrivialAssert(c));
```
RustanLeino, line 1061: Is there no `.Exists` or `.Any` method? If there is, then I think that would be easier to understand than `.Aggregate` with the extra parameter `containsAssert`.

Source/VCGeneration/Split.cs
```diff
@@ -868,6 +1036,65 @@ public Counterexample ToCounterexample(ProverContext context)
...
+          {
+            b.Cmds = b.Cmds.Take(firstFalseIdx + 1).ToList();
+            b.TransferCmd = (b.TransferCmd is GotoCmd) ? new ReturnCmd(b.tok) : b.TransferCmd;
+          }
+        }
+
+        void DeleteSubSumptions (Block b)
+        {
+          bool isAssumeSubsumption (Cmd c) { return c is AssumeCmd ac && QKeyValue.FindBoolAttribute(ac.Attributes, "subsumption"); }
+          b.Cmds = b.Cmds.Where(c => !isAssumeSubsumption(c)).ToList();
```
RustanLeino, line 1055: `{:subsumption 0}` is something that can be placed on an `AssertCmd`. This code looks for `:subsumption` placed on an `AssumeCmd`. I'm guessing this comes about because an `AssertCmd` has been turned into an `AssumeCmd`. Is it possible to instead never create such an `AssumeCmd` from an `assert {:subsumption 0} ...`?

RustanLeino: Also, I think it's best to also check for the the `0` argument of `:subsumption`, even though I don't think anyone ever uses `1` or `2`.

Source/VCGeneration/Split.cs
```diff
@@ -868,6 +1036,65 @@ public Counterexample ToCounterexample(ProverContext context)
...
+        todo.Push(blocks[0]);
+        while(todo.Count() > 0)
+        {
+          var currentBlock = todo.Peek();
+          var pop = peeked.Contains(currentBlock);
+          peeked.Add(currentBlock);
+          var interesting = false;
+          var exit = currentBlock.TransferCmd as GotoCmd;
+          if (exit != null && !pop)
+            exit.labelTargets.ForEach(b => todo.Push(b));
```
RustanLeino, line 1079: Put `{` `}` around the then branch.

Source/VCGeneration/Split.cs
```diff
@@ -868,6 +1036,65 @@ public Counterexample ToCounterexample(ProverContext context)
...
+
+        bool ContainsAssert(Block b)
+        {
+          bool isNonTrivialAssert (Cmd c) { return c is AssertCmd ac && !(ac.Expr is LiteralExpr le && le.asBool); }
+          return b.Cmds.Aggregate(false, (containsAssert, c) => containsAssert || isNonTrivialAssert(c));
+        }
+
+        // return blocks;
+        blocks.ForEach(b => DeleteSubSumptions(b));
+        blocks.ForEach(b => DeleteFalseGotos(b)); // make blocks ending in assume false leaves of the CFG-DAG -- this is probably unnecessary, may have been done previously
```
RustanLeino, line 1066: Might there be non-assert blocks that `goto` a block ending in `assume false`? If so, it would be nice to propagate the `assume false` to predecessor blocks. (Perhaps this is done elsewhere.)

Source/VCGeneration/Split.cs
```diff
@@ -736,6 +737,173 @@ public Counterexample ToCounterexample(ProverContext context)
         throw new cce.UnreachableException();
       }
 
+      private static void PrintSet<T> (HashSet<T> s) {
+        foreach(T i in s)
+          Console.WriteLine(i);
+      }
+
+      private static Dictionary<Block, Block> ImmediateDominator(List<Block> blocks)
```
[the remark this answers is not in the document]
typerSniper: I did not understand why `immediateDominatorMap` returns a map from Block to a list of Blocks. Probably because it computes all dominators? So I would need to get the *last* dominator from it, topologically speaking.

Also, the algorithm in `Graph.cs` is for general graphs and I have a DAG here; DAGs make things simpler and more efficient (only need one pass through the graph).

That said, I don't have a good sense of how important the difference is. So if you think it won't matter, I can use the one in `Graph.cs`.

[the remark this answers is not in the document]
shazqadeer: Please move your fast algorithm to Graph.cs as a public method. Document its difference with the other one.  Then we can proceed to landing this PR.

shazqadeer:
I noticed that there are hardly any tests for :split_here.  There are only a few occurrences in the Test/inst folder. I noticed that when :split_here is used with /trace, the output indicates how many splits happened.  Perhaps you can use that as some indication that the split is indeed happening.  My main point is that we need some tests.

Source/VCGeneration/Split.cs
typerSniper: `VCGen.AssertTurnedIntoAssume` already does it! No need for this piece of code.

typerSniper:
I added a test for checking the number of splits. Its a bit hacky---I dump the splits into files and check the names of the generated files. I do this using gratuitious diffs because I don't want to assume additional commands from the OS.

Source/Graph/Graph.cs
```diff
@@ -586,6 +586,49 @@ public List<Node> SuccessorsAsList(Node n)
       }
     }
 
+    // this function gives a simpler way to compute dominators but it assmumes the graph is a DAG.
```
shazqadeer, line 589: "this function" -> "This method"

"in the topological order" -> "in topological order"

remove period after "order"

"The function" -> "This method"

Source/Graph/Graph.cs
```diff
@@ -586,6 +586,49 @@ public List<Node> SuccessorsAsList(Node n)
...
+          s.UnionWith(dominators[predecessors.First()]);
+          predecessors.ForEach(v => s.IntersectWith(dominators[v]));
+        }
+        s.Add(u);
+        dominators[u] = s;
+      }
+      return dominators;
+    }
+
+    // Only use this for DAGs because it uses DominatorsFast() for computing dominators
```
shazqadeer, line 614: "Only use this" -> "Use this method only"

Source/VCGeneration/VCGen.cs
shazqadeer: I thought the PR that introduced this general method already landed. Why is this method showing up as a diff?

Source/VCGeneration/VCGen.cs
```diff
@@ -911,7 +929,7 @@ void SplitAndVerify(Implementation impl, Dictionary<TransferCmd, ReturnCmd> goto
               {
                 if (prover_failed)
```
shazqadeer, line 930: This is another annoying divergence from the convention of using camelCase for variable names.  Since you are modifying this file anyway, could you please take a shot at renaming names like "prover_failed" to "proverFailed" (and others).  But I won't insist on this change for this PR.  Perhaps you would consider doing this change in a separate PR.

typerSniper: Let's merge this in and I will soon create another PR for variable renaming.

Source/VCGeneration/VCGen.cs
shazqadeer: @typerSniper : Could you please respond to my question about BlocksToDAG so we can proceed to landing this PR?

shazqadeer (approved):

PR #425, opened 2021-07-28
always_assume attribute for specs 

typerSniper:
feat: `always_assume` attribute on `requires` and `ensures`.  On a requires, this lets the caller assume the pre-condition, instead of proving it. On an ensures, this lets the procedure's implementation assume the post-condition (instead of proving it).

### DIFF
--- a/Source/Core/Absy.cs
+++ b/Source/Core/Absy.cs
@@ -3675,7 +3675,8 @@ namespace Microsoft.Boogie
 
     public QKeyValue Attributes { get; set; }
 
-    public bool CanAlwaysAssume () {
+    public bool CanAlwaysAssume ()
+    {
       return QKeyValue.FindBoolAttribute(this.Attributes, "alwaysAssume");
     }
 
--- a/Source/Core/Absy.cs
+++ b/Source/Core/Absy.cs
@@ -3548,7 +3548,7 @@ namespace Microsoft.Boogie
 
     public bool CanAlwaysAssume()
     {
-      return QKeyValue.FindBoolAttribute(Attributes, "alwaysAssume");
+      return QKeyValue.FindBoolAttribute(Attributes, "always_assume");
     }
 
 
@@ -3677,7 +3677,7 @@ namespace Microsoft.Boogie
 
     public bool CanAlwaysAssume ()
     {
-      return QKeyValue.FindBoolAttribute(this.Attributes, "alwaysAssume");
+      return QKeyValue.FindBoolAttribute(this.Attributes, "always_assume");
     }
 
     public Ensures(IToken token, bool free, Expr /*!*/ condition, string comment, QKeyValue kv)
--- a/Source/Core/Absy.cs
+++ b/Source/Core/Absy.cs
@@ -3548,7 +3548,7 @@ namespace Microsoft.Boogie
 
     public bool CanAlwaysAssume()
     {
-      return QKeyValue.FindBoolAttribute(Attributes, "always_assume");
+      return Free && QKeyValue.FindBoolAttribute(Attributes, "always_assume");
     }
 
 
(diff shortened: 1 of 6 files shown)


shazqadeer:
Please provide a PR description.

Source/Core/Absy.cs
```diff
@@ -3546,6 +3546,12 @@ public String ErrorMessage
       get { return QKeyValue.FindStringAttribute(Attributes, "msg"); }
     }
 
+    public bool CanAlwaysAssume()
+    {
+      return QKeyValue.FindBoolAttribute(Attributes, "alwaysAssume");
```
shazqadeer, line 3551: I think the convention would dictate naming the attribute "always_assume".

shazqadeer:
Is always_assume supposed to be put only on free requires and ensures?

shazqadeer:
It would be helpful if you provided motivation for this attribute in the PR description.  I can imagine some usefulness for always_assume on a requires but it appears useless on an ensures.

typerSniper:
> Is always_assume supposed to be put only on free requires and ensures?

No. I didn't see a good reason to disallow it on non-free specs. 

Νοte, `free requires {:always_assume} P` is equivalent to `requires {:always_assume} P`. Afaik, for Dafny we could turn all `free requires` to `requires {:always_assume}` (@RustanLeino?). 

But the use case today would only put it on `free` specs. So imposing this restriction is not blocking and I don't feel strongly either way. 

> It would be helpful if you provided motivation for this attribute in the PR description. I can imagine some usefulness for always_assume on a requires but it appears useless on an ensures.

It maybe useful when a procedure is checked. For example:
```
Procedure X
  free ensures Q
  ensures P 
{
}
```
When checking X, Boogie would forget about `Q` entirely and try to prove `P` directly. However, in
```
Procedure X
  free ensures {:always_assume} Q
  ensures P 
{
}
```
Boogie would try to prove `Q ==> P`---because I insert an `assume Q`  which then is used for the subsequent `assert P` (after post-conditions are injected). 

This is useful if some conditions hold after an implementation, but you don't want to prove them and want to use them for other post-conditions. In Dafny, this is true of `canCalls`. I want to use `P#canCall` for proving `P` but I can't do so right now. One clumsy workaround is to change X to 

```
procedure X
  free ensures Q
  ensures Q ==> P 
{
}
```

I don't like this very much because if you have multiple such Q's and P's, the spec gets bigger and uglier. 
(I will add all this to the description tomorrow)

shazqadeer:
It sounds like putting :always_assume on a requires automatically "adds" a free on it.  Same for ensures.  Is my understanding correct?  If yes, I think this behavior is confusing.  I would recommend leaving the meaning of ordinary requires and ensures unchanged and allowing the user to further customize free requires/ensures with :always_assume.

Please make sure that you have test cases that demonstrate the benefit of :always_assume on both free requires and free ensures.

Also, say a bit more in the documentation of :always_assume to explain the contrast with ordinary free specifications.

typerSniper:
I've updated the code to ignore the attribute on non-free specs. Also, I changed the description to make it clear what happens with and without the attribute. As for test cases, I added `Test/test0/AlwaysAssume.bpl` that demonstrates the effect of the attribute on both requires and free ensures. The test case itself though is not very interesting.

Source/Core/Absy.cs
```diff
@@ -3669,6 +3675,11 @@ public String ErrorMessage
 
     public QKeyValue Attributes { get; set; }
 
+    public bool CanAlwaysAssume ()
+    {
+      return QKeyValue.FindBoolAttribute(this.Attributes, "always_assume");
```
shazqadeer, line 3680: You can also check that the requires/ensures is Free within this method so that the caller does not have to check for Free.

shazqadeer (approved):

shazqadeer (approved):

shazqadeer:
@typerSniper :  Thanks for excellent work on this PR.   Shall I land it?

typerSniper:
Yeah it looks ready.

RustanLeino:
Yes, please. Thanks for your review, @shazqadeer . This PR fixes an asymmetry in how procedure specifications are handled, which from time to time has caused Dafny postconditions not to prove without duplicating the postcondition as an assertion at the end of the method body.

PR #426, opened 2021-07-28
Focus

typerSniper:
feat: focus attribute on predicate commands (asserts and assumes). It provides another way to split up verification work into smaller chunks. 

Think of a Boogie implementation after its a DAG (no back edges). Consider some Block `b` which has an assert with attribute `focus`.  The attribute achieves the property: Block `b` only interacts with its ancestors and descendants. Verification Checks that involve Block `b`  are not affected by "non-ancestor-descendant" blocks (they don't even show up in the Z3 formula). Moreover, other blocks are totally oblivious to Block `b`. 

Conceptually, you can think that verification is split into two parts: first, where all the control flow goes through `b` and second, where block `b` is not even reachable. Alternatively, you can think of it like first placing `assume false` on all _other_ paths, verifying the DAG, then placing `assume false` on Block `b`, and then verifying again.

There is an interesting design decision to be made when there are multiple blocks with the `focus` attribute---in what order do we process the _foci_? Perhaps surprisingly, if two blocks are not ancestor-descendant then it doesn't matter which one you process first. However, it does matter if two foci are on the same path. If you process the ancestor first, you get more splits (even exponential, I believe). However, if you process the descendant first, you get a linear number of splits. I believe exponential splits are better because they are more true to the semantics: a focussed block is always hidden to its siblings.  We can switch between these two choices by setting the variable `reallyFocus` to `true` or `false` in function `FocusImpl` in `Split.cs`.


Source/Core/CommandLineOptions.cs
```diff
@@ -1786,6 +1786,11 @@ identity function.
      {:subsumption n}
        Overrides the /subsumption command-line setting for this assertion.
 
+     {:focus}
+      It deletes all paths that don't have the containing block, assumes code leading to the block,
```
shazqadeer, line 1790: Documentation for other attributes begin at the column right after the :.

Splits verification into two problems. First problem deletes all paths that do not have the focus block. Second problem considers the paths deleted in the first problem and does not contain either the focus block or any block dominated by it.

Source/VCGeneration/Split.cs
```diff
@@ -1071,6 +1066,159 @@ bool ContainsAssert(Block b)
...
+          return blocks.Where(blk => blk.Cmds.Where(c => isFocusCmd(c)).Count() != 0).ToList();
+        }
+        var dag = Program.GraphFromImpl(impl);
+        var topoSorted = dag.TopologicalSort().ToList();
+        // If reallyFocus is set to true,
+        // foci are processed in a top-down fashion --- i.e., if two foci are on the same path,
+        // the ancestor is processed first.
+        // On the other hand, if reallyFocus is false,
+        // foci are processed in a bottom-up fashion --- i.e., the descendant is processed first.
+        bool reallyFocus = true;
```
shazqadeer, line 1084: Any particular reason why you chose not to make "reallyFocus" specifiable by a Boogie user, potentially by making it a command-line parameter?  Perhaps you just want to experiment for a while before really deciding whether this option is worth it.

Source/VCGeneration/Split.cs
```diff
@@ -1071,6 +1066,159 @@ bool ContainsAssert(Block b)
...
+          } else {
+            return (topoSorted.IndexOf(b1) < topoSorted.IndexOf(b2)) ^ reallyFocus ? 1 : -1;
+          }
+        }
+        List<Block> focusBlocks = getFocusBlocks(impl.Blocks);
+        if(focusBlocks.Count() == 0) {
+          return null;
+        }
+
+        focusBlocks.Sort(compareBlocks); // if reallyFocus is true, blocks are sorted according to the topological order; otherwise they are placed in reverse topo order.
```
shazqadeer, line 1097: The comment is long.  Put it in a block above the code.

Source/VCGeneration/Split.cs
```diff
@@ -1071,6 +1066,159 @@ bool ContainsAssert(Block b)
...
+            if (visited.Contains(b)) continue;
+            var related = direction ? dag.Successors(b) : dag.Predecessors(b);
+            related.Where(b => !visited.Contains(b)).ToList().ForEach(b => todo.Push(b));
+            visited.Add(b);
+          }
+          return visited;
+        }
+
+        // finds all the blocks dominated by focusBlock in the subgraph
+        // which only contains vertices of subGraph.
```
shazqadeer, line 1115: subGraph ->subgraph

shazqadeer:
A Boogie use may put :split_here and :focus in an arbitrary fashion.  Is your implementation guaranteed to be sound for any such placement?

typerSniper:
> A Boogie use may put :split_here and :focus in an arbitrary fashion. Is your implementation guaranteed to be sound for any such placement?

Yes, it is designed to stay sound. However, one needs to be careful because it could potentially generate a very large number of splits.

Test/test0/Focus.bpl
```diff
@@ -0,0 +1,33 @@
+// RUN: %boogie /vcsDumpSplits "%s" > "%t"
+// RUN: %diff "%s.expect" "%t"
+// RUN: %diff Ex.split.-1.bpl Ex.split.-1.bpl
```
shazqadeer, line 3: What does this command do?  It seems to be diffing a file against itself.

Source/Core/CommandLineOptions.cs
```diff
@@ -2097,6 +2104,9 @@ be n (default 500)
                 only for monomorphic programs.
   /reflectAdd   In the VC, generate an auxiliary symbol, elsewhere defined
                 to be +, instead of +.
+  /laxFocus     Process foci in a bottom-up fashion. This way only generates
```
shazqadeer, line 2107: /relaxFocus?

Source/VCGeneration/Split.cs
```diff
@@ -1071,6 +1066,169 @@ bool ContainsAssert(Block b)
...
+      {
+        bool IsFocusCmd(Cmd c) {
+          return c is PredicateCmd p && QKeyValue.FindBoolAttribute(p.Attributes, "focus");
+        }
+        List<Block> GetFocusBlocks(List<Block> blocks) {
+          return blocks.Where(blk => blk.Cmds.Where(c => IsFocusCmd(c)).Count() != 0).ToList();
+        }
+        var dag = Program.GraphFromImpl(impl);
+        var topoSorted = dag.TopologicalSort().ToList();
+        // If reallyFocus is set to true,
```
shazqadeer, line 1079: The comment here and on line 1097 are repetitive.  Please consolidate in one place.

Source/VCGeneration/Split.cs
```diff
@@ -1071,6 +1066,169 @@ bool ContainsAssert(Block b)
...
+        // foci are processed in a top-down fashion --- i.e., if two foci are on the same path,
+        // the ancestor is processed first.
+        // On the other hand, if reallyFocus is false,
+        // foci are processed in a bottom-up fashion --- i.e., the descendant is processed first.
+        bool reallyFocus = !CommandLineOptions.Clo.LaxFocus;
```
shazqadeer, line 1084: The use of the local variable "reallyFocus" now seems gratuitous.

Source/Core/CommandLineOptions.cs
shazqadeer: Process foci in a bottom-up fashion, guaranteeing a linear number of splits are generated. 
The default (top-down) is more aggressive and may create an exponential number of splits.

shazqadeer (approved):

Test/test0/Focus.bpl
typerSniper: I just wanted to check that this file is generated and I did not want to assume other OS level commands.

Source/Core/CommandLineOptions.cs
shazqadeer: Please update the documentation also:

Process foci in a bottom-up fashion, guaranteeing a linear number of splits are generated.
The default (top-down) is more aggressive and may create an exponential number of splits.

PR #429, opened 2021-08-09
Replace using statements with using declarations in solution

keyboardDrummer:
This PR replaces using statements with using declarations to reduce unhelpful syntax. More information is in the release notes of the feature: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-8.0/using#motivation

### DIFF
--- a/Source/Predication/SmartBlockPredicator.cs
+++ b/Source/Predication/SmartBlockPredicator.cs
@@ -328,7 +328,7 @@ namespace Microsoft.Boogie
               var condPred = FreshPredicate(ref predCount);
               predMap[block.Item1] = condPred;
               defMap[block.Item1] = condPred;
-              var headerIterator = headerDominance.GetEnumerator();
+              using var headerIterator = headerDominance.GetEnumerator();
               // Add the predicate to the loop header H that dominates the node (if one
               // exists) such that H does not dominate another header which also dominates
               // the node. Since predicates are owned by loop headers (or the program entry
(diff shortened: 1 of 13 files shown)


Source/Predication/SmartBlockPredicator.cs
```diff
@@ -328,7 +328,7 @@ Variable FreshPredicate(ref int predCount)
               var condPred = FreshPredicate(ref predCount);
               predMap[block.Item1] = condPred;
               defMap[block.Item1] = condPred;
-              var headerIterator = headerDominance.GetEnumerator();
+              using var headerIterator = headerDominance.GetEnumerator();
```
shazqadeer, line 331: What is the difference in behavior as a result of this change?

shazqadeer:
What is the primary motivation for this PR?

Source/Predication/SmartBlockPredicator.cs
keyboardDrummer: I'm guessing no practical difference, but now the method satisfies the contract of `GetEnumerator`, which is that it should be disposed after it's used. This isn't required when using `foreach` to enumerate, and I'm not sure why foreach is not used here.

keyboardDrummer:
> What is the primary motivation for this PR?

The motivation is to reduce the amount of syntax, the same as why using declarations were introduced into the language: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-8.0/using#motivation

shazqadeer (approved):

PR #431, opened 2021-08-10
Clarify release instructions

RustanLeino:
The `git push --tags` command is discouraged, since it will push _all_ tags on the local machine. For example, this fails for me, which leaves everything in a state I don't understand.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ To push a new version to nuget, perform the following steps:
 
 - Update the version (e.g., x.y.z) in `Source/Directory.Build.props` and commit the change locally
 - `git tag vx.y.z`
-- `git push origin vx.y.z` where `origin` denotes the remote on the github.com
+- `git push origin master vx.y.z` where `origin` denotes the remote on github.com
 
 The [CI workflow](.github/workflows/test.yml) will build and push the packages.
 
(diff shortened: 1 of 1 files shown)


shazqadeer (approved):

README.md
```diff
@@ -107,9 +107,9 @@ See the [Unit test documentation](Source/UnitTests/README.md)
 The current version of Boogie is noted in a [build property](Source/Directory.Build.props).
 To push a new version to nuget, perform the following steps:
 
-- Update the version (e.g., x.y.z) and commit the change
-- git tag vx.y.z
-- git push --tags
+- Update the version (e.g., x.y.z) in `Source/Directory.Build.props` and commit the change locally
+- `git tag vx.y.z`
+- `git push origin vx.y.z` where `origin` denotes the remote on the github.com
```
shazqadeer, line 112: Consider removing `the` before github.com.

bkragl: I think you need to include the master branch in the push command, i.e., `git push origin master vx.y.z`. Otherwise the commit will only be "reachable" via the tag, but others won't see it (and develop on top of it) on the master branch.

bkragl (changes requested):

README.md
RustanLeino: Ouch. Thanks for the suggestion. (I probably made this mistake when trying to get v2.9.3 out.) I made the edit.

PR #432, opened 2021-08-18
Refactoring some of the code 

typerSniper:
This PR does three things:

1) Prune dead variables. After focussing and splitting, fewer variables are live. The dead variables and their corresponding where clauses are now removed. This removal only happens if the `/pruneFunctionsAndAxioms` flag is set. It changes the behavior of the following program:
```
var x: int where false 

if (*) {
 assert x > 0;
 assert false;
} else {
 assume {:focus} true;
 assert false;
}
```

The code above generates two splits: one with only the `if` branch and another with only the `else` branch. But `x` is treated to be live in both because dead variable elimination is done before focussing/splitting. So, in the earlier version, Boogie determines that `x` is live, so it assumes the `where` clause of `x`. This means it inserts `assume false` **before** the `if` and both assertions prove. 

Now, if pruning is enabled, dead variables are recomputed. So, we determine the variable `x` to be dead in the `else` split and delete its where clause (`assume false`). This means the assertion in `else` fails when pruning is enabled.

2) Refactor `FindAttribute`: The code around finding attributes looked a bit clunky: it had for loops everywhere for iterating through the attributes. I have refactored  `FindAttribute` to take a `property: QKeyValue -> bool` function; it applies this function to every attribute and returns the first one that satisfies it. 

3) Compose focussing and splitting: I made a mistake in composing focussing and splitting earlier. Nothing unsound but it was generating more splits than necessary. I have fixed it and refactored the code a bit to make it cleaner.


typerSniper:
These failing tests are changes in error traces. Can I update them?

shazqadeer:
> These failing tests are changes in error traces. Can I update them?

Go ahead.

typerSniper:
@shazqadeer , @RustanLeino: I've updated the expect files.

shazqadeer (approved):

PR #434, opened 2021-08-24
Add tests for parallel processing

keyboardDrummer:
Let some of the Boogie tests in the suite use multiple processors to enable testing the parallel processing features.

We turn off error tracing for these tests since there are multiple valid error traces for a single verification error, and which trace is shown by Boogie can depend on the order in which Boogie verifies the program which depends on the amount of threads used.

### DIFF
--- a/Test/lit.site.cfg
+++ b/Test/lit.site.cfg
@@ -104,6 +104,7 @@ if len(boogieParams) > 0:
 lit_config.note('Using Boogie: {}'.format(boogieExecutable))
 
 config.substitutions.append(('%boogie', boogieExecutable))
+config.substitutions.append(('%parallel-boogie', boogieExecutable + ' /vcsCores:2 /errorTrace:0'))
 
 # Add diff tool substitution
 if os.name == 'posix':
--- a/Test/aitest0/Intervals.bpl
+++ b/Test/aitest0/Intervals.bpl
@@ -1,4 +1,4 @@
-// RUN: %boogie -infer:j "%s" > "%t"
+// RUN: %parallel-boogie -infer:j "%s" > "%t"
 // RUN: %diff "%s.expect" "%t"
 const N: int;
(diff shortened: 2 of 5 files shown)


Test/lit.site.cfg
```diff
@@ -103,7 +103,8 @@ if len(boogieParams) > 0:
 # Inform user what executable is being used
 lit_config.note('Using Boogie: {}'.format(boogieExecutable))
 
-config.substitutions.append(('%boogie', boogieExecutable))
+config.substitutions.append(('%boogie', boogieExecutable + ' /vcsCores:2'))
+config.substitutions.append(('%sequential-boogie', boogieExecutable))
```
shazqadeer, line 107: I would prefer if you used boogie for the sequential version and parallel-boogie for the parallel version.

shazqadeer (commented):
I would have thought that there are a lot more tests with golden output containing error traces.  Are you sure that you got all of them?

keyboardDrummer:
> I would have thought that there are a lot more tests with golden output containing error traces. Are you sure that you got all of them?

Tests with only a single procedure don't seem to be affected by this change. Does that explain the low number of affected tests?

Test/lit.site.cfg
keyboardDrummer: I thought letting `boogie` be `parallel-boogie` would be nicer since it implies that parallel is the default, leading to more tests using the parallel feature. I can also imagine that we don't want that to be the default, but in that case I wonder which tests should we pick to use `parallel-boogie`, certainly not as many of them as we can, as I've done in this PR, right?

The downside of tests using `parallel-boogie` is that if some of the debug options are used the test can be unstable, although I'm guessing this is detected early on.

shazqadeer: @RustanLeino :  Please chime in.

PR #435, opened 2021-08-25
Move Checker common state so the cache can be hit

keyboardDrummer:
Many commits ago the field `VCGen.CheckerCommonState` was introduced ([ref](https://github.com/boogie-org/boogie/blame/1f7016e583f2264340385b480a4507e35133669d/Source/VCGeneration/Check.cs#L134)). At that time a single `VCGen` was used for an entire Boogie program ([ref](https://github.com/boogie-org/boogie/blob/1f7016e583f2264340385b480a4507e35133669d/Source/BoogieDriver/BoogieDriver.cs#L561)), so the `CheckerCommonState` could be shared between implementations. The current master branch creates a `VCGen` per Implementation so it seems the cache in `CheckerCommonState` never hits. 

### Changes
Move the `CheckerCommonState` there to enable the cache to function.


shazqadeer:
@keyboardDrummer : Should this PR be abaondoned?

PR #438, opened 2021-09-07
Don't automatically run StyleCop for any projects referencing Boogie projects

keyboardDrummer:
I ran into the issue that when referencing Boogie projects, instead of the Boogie NuGet packages, StyleCop was ran on non-Boogie packages. This change resolved that. [More information](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/2278).


shazqadeer (approved):
I don't understand StyleCop but the change looks harmless.  

FWIW, StyleCop support was added by @zvonimir

PR #439, opened 2021-09-08
Small step towards proof isolation for implementations

keyboardDrummer:
To enable stable proof generation, the presence of one Boogie implementation shouldn't effect what gets sent to the solver when verifying another Boogie implementation. 

### Changes

- Add a test for proof stability that looks at two variants of a tiny program. 
- Change the integer identifier that Boogie uses to reference AST nodes in the verifier to improve implementation proof isolation.


Source/Core/Absy.cs
```diff
@@ -63,18 +63,15 @@ public Absy(IToken tok)
     // We uniquely number every AST node to make them
     // suitable for our implementation of functional maps.
     //
-    public int UniqueId
-    {
-      get { return this.uniqueId; }
-    }
+    public int UniqueId => this.uniqueId;
 
     private const int indent_size = 2;
```
shazqadeer, line 68: This is unrelated to your diff.  But the name of this constant does not follow the naming convention.  I want to raise awareness so contributors can start cleaning up code as we go along.

Source/Core/IdMap.cs
```diff
@@ -0,0 +1,28 @@
+using System.Collections.Generic;
+
+namespace Microsoft.Boogie
+{
+  public class IdMap<T>
+  {
+    Dictionary<T, int> toId = new ();
+    Dictionary<int, T> fromId = new ();
+    private int counter = 10;
```
shazqadeer, line 9: Why 10?

shazqadeer (commented):
I understand that instead of using absy.UniqueId, you are now using your new idmap to assign a number to each absy node.  I don't understand how this helps with proof isolation.  Could you please expand your explanation in the PR summary?

Source/Core/Absy.cs
RustanLeino: Another comment on this: For a property getter like `UniqueId` that is backed by a private field (like `uniqueId`), I think you can just write `public int UniqueId { get; }` and drop the explicit declaration of the private field. Other than needing a property getting to implement a property getting in an `interface`, I've never understood the big fuss about having property getters--so, speaking for myself, I would just declare this as a field `public int UniqueId;` and be done with it.

RustanLeino (commented):
LGTM, once Shaz's comments have been addressed.

Source/Core/Absy.cs
keyboardDrummer: > This is unrelated to your diff. But the name of this constant does not follow the naming convention. I want to raise awareness so contributors can start cleaning up code as we go along.

Ideally we would let the build detect improper naming, possibly after a solution wide automated rename to let all names match the convention. I don't know how to set that up but I hope to look into it at some point.

> Another comment on this: For a property getter like UniqueId that is backed by a private field (like uniqueId), I think you can just write public int UniqueId { get; } and drop the explicit declaration of the private field.

This too might be something we can have the build check for. I'll leave this code for now since changing it isn't in scope of the PR.

Source/Core/IdMap.cs
keyboardDrummer: I don't have a good answer to that. `0` doesn't work, it seems like that's a reserved number in the control flow scheme but I didn't investigate to confirm that.

I'll change the code to explain the choice for 10.

Source/Core/Absy.cs
shazqadeer: > Ideally we would let the build detect improper naming, possibly after a solution wide automated rename to let all names match the convention. I don't know how to set that up but I hope to look into it at some point.
>  
> This too might be something we can have the build check for.

@zvonimir set up StyleCop for Boogie.  He might know how to get this done.

Source/Core/IdMap.cs
shazqadeer: Ok, I remember now.  0 is indeed reserved because it is the value assigned to the fictitious control-flow location at the beginning of an error trace.  Please include this note as a comment.

RustanLeino (dismissed):

Source/Core/IdMap.cs
keyboardDrummer: Done

Source/Core/Absy.cs
zvonimir: So this is the file to look at: https://github.com/boogie-org/boogie/blob/master/Source/StyleCop.ruleset
Here I disabled a bunch of checks that were causing tons of style checking errors to get generated. At that point I did not have time to fix all of those, and the plan was to merge basic checking, and get back to these and take care of them one by one.
Note that StyleCop is far from perfect, but is the best style checker for C# I managed to fine.
Anyhow, you want to look for something like this in that file:
```
<Rule Id="SA1310" Action="None" /> <!-- Field should not contain an underscore -->
```
Then you remove this line and see what happens. Then you do all the fixes.
I hope this makes sense.

shazqadeer (approved):

PR #440, opened 2021-09-09
Remove references to `CommandLineOptions.Clo` in `SMTLib`

keyboardDrummer:
This PR is the first in a series to remove the static variable `CommandLineOptions.Clo`. This variable has a few downsides:

1. When calling any function in Boogie, it's not clear from the function signature how it can be configured using `CommandLineOptions.Clo` and there's no easy way to figure this out. Relatedly, external callers of Boogie must initialise `CommandLineOptions.Clo`, but when they call a public Boogie API it's not obvious that they should.
1. Because `CommandLineOptions.Clo` is referenced from everywhere it has to live at the bottom of the project reference stack, in `Core`, but this means it can't refer to any interesting types defined higher in the stack and this leads to options typed `object` that have to be cast at a later time. 
    1. An example is the function `CommandLineOptions.TheProverFactory.SpawnProver` which returns `object` while in practice it returns a `ProverInterface` which is defined in the `VCGeneration` project that's not available from `Core`.
1. It's not possible to run Boogie in parallel with different command line options, since the options are shared between threads. Note this could also be solved by making `CommandLineOptions.Clo` thread-local.

This PR replaces references to the global variable `CommandLineOptions.Clo` that occur in the project `SMTLib`.


Source/Provers/SMTLib/SMTLibLineariser.cs
```diff
@@ -26,31 +26,50 @@ public class LineariserOptions
   // Lineariser for expressions. The result (bool) is currently not used for anything
   public class SMTLibExprLineariser : IVCExprVisitor<bool, LineariserOptions /*!*/>
   {
-    public static string StoreOpName(VCExprNAry node)
+    public SMTLibOptions LibOptions { get; }
```
shazqadeer, line 29: Is it bad style to use SMTLibOptions for the name of the field also?

shazqadeer (approved):

shazqadeer:
Here is my understanding of this PR:  It factors out a subset of the flags in CommandLineOptions into a separate interface called SMTLibOptions which is now implemented by CommandLineOptions.  These factored out options are presumably the ones that are referred to in the code that does the SMT encoding.  So far, so good.  How this PR contributes to the overall mission charted out in the PR summary remains a mystery to me.

zvonimir:
These pull requests and changes are almost certainly going to have major effects on Corral (https://github.com/boogie-org/corral/) downstream.
Would it be possible to update Corral as well along the way (since you know the best what refactoring you are exactly doing)?

keyboardDrummer:
> So far, so good. How this PR contributes to the overall mission charted out in the PR summary remains a mystery to me.

I'm not sure where the mystery comes from. The mission can be summarized as removing the static variable `CommandLineOptions.Clo`, for which we first need to remove all references to it. This PR removes all references to it that occur in the SMTLib project.

One step that this PR doesn't take is to place `SMTLibOptions` in `SMTLib` instead of in `Core`, which is required to remove downside (2) listed in the details. This change requires another refactoring that I'm leaving for another PR.

keyboardDrummer:
> These pull requests and changes are almost certainly going to have major effects on Corral (https://github.com/boogie-org/corral/) downstream.

I don't think this PR has an effect on downstream projects, but indeed the stated goal of removing `CommandLineOptions.Clo` will, because downstream projects might refer to it. The migration fix that downstream projects can do is to declare their own `DownStreamOptions.Clo` and refer to that instead of `CommandLineOptions.Clo`. This is an easy change. After it downstream projects can decide whether to remove references to static variables as well or keep it the way they have it.

In the PR that actually removes `CommandLineOptions.Clo`, I can include an upgrade guide in a release notes file, but don't worry it'll be tiny. The alternative of only making fully backwards compatible changes in Boogie would I think be a major blocker in evolving Boogie.

> Would it be possible to update Corral as well along the way (since you know the best what refactoring you are exactly doing)?

I can help if Corral upgrades to a Boogie version that requires it.

Source/Provers/SMTLib/SMTLibLineariser.cs
keyboardDrummer: I don't think so. I would have preferred to name it `Options` but that was often already taken by a field containing `ProverOptions`. I think ideally `SMTLibOptions` and `ProverOptions` would be related somehow and then we'd only need a single `Options` field, but I've left that out of scope for this PR.

shazqadeer:
> > So far, so good. How this PR contributes to the overall mission charted out in the PR summary remains a mystery to me.
> 
> I'm not sure where the mystery comes from. The mission can be summarized as removing the static variable `CommandLineOptions.Clo`, for which we first need to remove all references to it. This PR removes all references to it that occur in the SMTLib project, so it's a step in that direction.
> 
> One step that this PR doesn't take is to place `SMTLibOptions` in `SMTLib` instead of in `Core`, which is required to remove downside (2) listed in the PR description. That change requires another refactoring that I'm leaving for another PR.

I am trying to understand what the final product after all the refactoring will look like.  It sounds like the class CommandLineOptions will remain and will implement a collection of different interfaces, each presumably for a different component in Boogie.  The static field clo in CommandLineOptions will go away and a separate instance of CommandLineOptions will be created to initialize an instance of some component.  Is my understanding correct?

shazqadeer:
> The alternative of only making fully backwards compatible changes in Boogie would I think be a major blocker in evolving Boogie.

I agree.

Source/Provers/SMTLib/SMTLibLineariser.cs
shazqadeer: @bkragl :  FYI.  Your input about ProverOptions and SMTLibOptions and potential unification would be valuable since you spent a ton of time earlier cleaning up the Boogie-Prover interface.

keyboardDrummer:
> I am trying to understand what the final product after all the refactoring will look like. It sounds like the class CommandLineOptions will remain and will implement a collection of different interfaces, each presumably for a different component in Boogie. The static field clo in CommandLineOptions will go away and a separate instance of CommandLineOptions will be created to initialize an instance of some component. Is my understanding correct?

I think so yes. The definition of `CommandLineOptions` will move to `ExecutionEngine`. The root calls into boogie like many of the methods in `ExecutionEngine.cs` will take an instance of `CommandLineOptions`.

Note that this PR has, IMO, a positive effect regardless of future PRs, since it makes it easier to see what options can be passed to SMTLib.

PR #442, opened 2021-09-21
Prune constants

keyboardDrummer:
Reviewing tip: you can use [this link](https://github.com/boogie-org/boogie/pull/442/files/f397542bd6604b30b8467d24dbdcdf63d2fc3487..e84a1dc8d8ff592e909ca87278701750db49adb7) to see the changes except the first move refactoring commit.

### Changes
- Add the ability for Boogie to prune constants, in addition to the existing ability to prune axioms and functions.
- Change the complexity of the code that creates the reachability graph for pruning, to be linear instead of quadratic in program size
- Slightly simplify the pruning code so it uses one `incoming` field instead of two.

### Testing
- Add a test that verifies constants, functions, axioms and types are isolated when they're part of independent proofs.

### DIFF
--- a/Source/Core/Prune/Prune.cs
+++ b/Source/Core/Prune/Prune.cs
@@ -100,11 +100,16 @@ namespace Microsoft.Boogie
       blocks.ForEach(blk => blk.Cmds = blk.Cmds.Where(c => !DeadWhereAssumption(c)).ToList());
     }
 
-    public static IEnumerable<Declaration> PruneDecl(Program p, List<Block> blocks)
+    /*
+     * Global variables, type constructor declarations, type synonyms, procedures and implementations are not pruned, but none of them produce solver commands.
+     * See Checker.Setup for more information.
+     * Data type constructor declarations are not pruned and they do affect VC generation.
+     */
+    public static IEnumerable<Declaration> GetLiveDeclarations(Program program, List<Block> blocks)
     {
-      if (p.DeclarationDependencies == null || blocks == null || !CommandLineOptions.Clo.PruneFunctionsAndAxioms)
+      if (program.DeclarationDependencies == null || blocks == null || !CommandLineOptions.Clo.PruneFunctionsAndAxioms)
       {
-        return p.TopLevelDeclarations;
+        return program.TopLevelDeclarations;
       }
 
       BlocksVisitor blocksNode = new BlocksVisitor(blocks);
@@ -112,8 +117,8 @@ namespace Microsoft.Boogie
       TrimWhereAssumes(blocks, blocksNode.RelVars);
 
       // an implementation only has outgoing edges.
-      var reachableDeclarations = GraphAlgorithms.FindReachableNodesInGraphWithMergeNodes(p.DeclarationDependencies, blocksNode.outgoing).ToHashSet();
-      var result = p.TopLevelDeclarations.Where(d => d is not Constant && d is not Axiom && d is not Function || reachableDeclarations.Contains(d));
+      var reachableDeclarations = GraphAlgorithms.FindReachableNodesInGraphWithMergeNodes(program.DeclarationDependencies, blocksNode.outgoing).ToHashSet();
+      var result = program.TopLevelDeclarations.Where(d => d is not Constant && d is not Axiom && d is not Function || reachableDeclarations.Contains(d));
       return result;
     }
   }
--- a/Source/UnitTests/ExecutionEngineTests/ProofIsolation.cs
+++ b/Source/UnitTests/ExecutionEngineTests/ProofIsolation.cs
@@ -56,13 +56,13 @@ function age(Wicket) returns (int);
 axiom age(w) == 7;
 
 var favorite: Wicket;
+var alternative: Wicket;
 
 type Barrel a;
 type RGBColor;
 const unique red: RGBColor;
 const unique green: RGBColor;
 const unique blue: RGBColor;
-axiom ( forall ce: RGBColor :: ce == red || ce == green || ce == blue );
 
 const m: [Barrel Wicket] Wicket;
 const n: <a> [Barrel a] a;
@@ -76,6 +76,7 @@ procedure M(x: int, coloredBarrel: Barrel RGBColor)
   ensures age(favorite) == 42; 
 {
   var y: ComplicatedInt;
+  favorite := alternative;
   assert y == 3;
 }
 ";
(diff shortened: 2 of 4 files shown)


Source/Core/Prune/Prune.cs
```diff
@@ -10,21 +11,44 @@ private static bool ExcludeDep(Declaration d)
...
+          object source;
+          if (incomingTuple.Length == 0) {
+            continue;
+          }
+
+          if (incomingTuple.Length == 1) {
+            source = incomingTuple.First();
+          } else {
+            foreach (var mergeIncoming in incomingTuple) {
+              var mergeIncomingTargets = edges.GetOrCreate(mergeIncoming, () => new());
```
RustanLeino, line 36: This notation is very concise. Nice!

Source/Core/Prune/Util.cs
```diff
@@ -0,0 +1,30 @@
...
+      if (dictionary.TryGetValue(key, out var result)) {
+        return result;
+      }
+
+      result = createValue();
+      dictionary[key] = result;
+      return result;
+    }
+    
+    public static uint SafeMult(uint a, uint b) {
```
RustanLeino, line 19: The name "safe" here seems wrong to me. The method method is anything but safe, because it may not multiply at all. Instead, it returns the minimum of `MaxValue` and `a * b`. A name like `ApproximateMult` may be better, and adding a descriptive comment also seems good.

Source/Core/Prune/Util.cs
```diff
@@ -0,0 +1,30 @@
...
+      uint result = UInt32.MaxValue;
+      try {
+        checked {
+          result = a * b;
+        }
+      } catch (OverflowException) { }
+
+      return result;
```
RustanLeino, line 27: I think this would be clearer as

```
try {
  return a * b;
} catch (OverflowException) {
  return UInt32.MaxValue;
}
```

Source/Graph/Graph.cs
```diff
@@ -1106,6 +1106,36 @@ public ICollection<Node> Reachable()
...
+
+    /**
+     * A merge node is a node that has multiple incoming edges, and which cannot be traversed unless all incoming edges have been traversed.
+     * A merge node is represented by an object of type IReadOnlySet{object}
+     */
+    public static IEnumerable<object> FindReachableNodesInGraphWithMergeNodes(Dictionary<object, List<object>> edges, IEnumerable<object> roots)
+    {
+      var todo = new Stack<object>(roots);
+      var visitedEdges = new HashSet<object>();
+      while(todo.Any())
```
RustanLeino, line 1118:
```suggestion
      while (todo.Any())
```

Source/Graph/Graph.cs
```diff
@@ -1106,6 +1106,36 @@ public ICollection<Node> Reachable()
...
+      {
+        var node = todo.Pop();
+        if (visitedEdges.Contains(node)) continue;
```
RustanLeino, line 1121: Put the "then" branch (that is, the `continue;` statement) into curly braces.

Source/Graph/Graph.cs
```diff
@@ -1106,6 +1106,36 @@ public ICollection<Node> Reachable()
...
+        
+        if (node is IReadOnlySet<object> mergeNode) {
+          if (!visitedEdges.IsSupersetOf(mergeNode)) 
+            continue;
```
RustanLeino, line 1125: Use curly braces around the "then" branch.

RustanLeino (changes requested):
Oddly enough, I didn't see where the new processing for constants was done. In all places where I saw branching for constants in this PR, the previous code already had the same case. Where did I miss the new pruning itself?

Source/Graph/Graph.cs
keyboardDrummer: In the file `Source/StyleCop.ruleset` there's a line `<Rule Id="SA1503" Action="None" /> <!-- Braces should not be omitted -->` which makes it possible to leave out the braces around the continue. There's over a hundred location in Boogie where the braces are currently skipped.

I suggest we use the `StyleCop.ruleset` file as a guideline for styling. If we want to enforce curly braces everywhere, let's create a PR for that, where we update `StyleCop.ruleset` and all existing wrong cases.

Source/Core/Prune/Util.cs
keyboardDrummer: This was existing code copied from somewhere else, but I'll change it.

Source/Core/Prune/Util.cs
keyboardDrummer: Renamed it, not this was an existing method that I moved because it was in another class called Util.

keyboardDrummer:
> Oddly enough, I didn't see where the new processing for constants was done. In all places where I saw branching for constants in this PR, the previous code already had the same case. Where did I miss the new pruning itself?

Previously, the reachability graph computed only contained axioms and functions, so the traversal could not detect which constants were traversed. If you look a [the old](https://github.com/boogie-org/boogie/blob/master/Source/Core/Prune.cs#L243) edge computation:

```
    public static Dictionary<DependencyEvaluator, List<DependencyEvaluator>> InitializeEdges(Program p)
    {
      if (!CommandLineOptions.Clo.PruneFunctionsAndAxioms)
      {
        return null;
      }
      var nodes = p.Axioms.Select(ax => (DependencyEvaluator)new AxiomVisitor(ax)).ToList();
      nodes.ForEach(axv => ((AxiomVisitor)axv).Visit(axv.node));
      var functionNodes = p.Functions.Select(f => (DependencyEvaluator)new FunctionVisitor(f)).ToList();
      functionNodes.ForEach(fv => ((FunctionVisitor)fv).Visit(fv.node));
      nodes.AddRange(functionNodes);
      nodes.ForEach(u => u.incoming = u.incoming.Where(i => u.node == i || !ExcludeDep(i)).ToHashSet());
      nodes.ForEach(u => u.outgoing = u.outgoing.Where(i => !ExcludeDep(i)).ToHashSet());
      var edges = new Dictionary<DependencyEvaluator, List<DependencyEvaluator>>();
      nodes.ForEach(u => edges[u] = nodes.Where(v => DependencyEvaluator.depends(u, v)).ToList());
      return edges;
    }
```

You can see that it computes edges only from and to `nodes`, which contains only axioms and function. It assumes that there's only a single hidden node between an axiom or function. The call to `DependencyEvaluator.depends(u, v)` sees if two axioms/functions are connected by such a hidden node.

There was also a quadratic complexity issue arising from the double iteration over `nodes`.

RustanLeino (approved):

PR #443, opened 2021-09-22
Turn on 'Braces should not be omitted' and resolve existing violations

keyboardDrummer:
This change was made because of this comment: https://github.com/boogie-org/boogie/pull/442#discussion_r713403442

### DIFF
--- a/Source/StyleCop.ruleset
+++ b/Source/StyleCop.ruleset
@@ -1,101 +1,100 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <RuleSet Name="StyleCop rules for Boogie" Description="Custom Rules" ToolsVersion="16.0">
   <Rules AnalyzerId="StyleCop.Analyzers" RuleNamespace="StyleCop.Analyzers">
-    <Rule Id="SA0001" Action="None" />
-    <Rule Id="SA1000" Action="None" />
-    <Rule Id="SA1001" Action="None" />
-    <Rule Id="SA1003" Action="None" />
-    <Rule Id="SA1005" Action="None" />
-    <Rule Id="SA1008" Action="None" />
-    <Rule Id="SA1009" Action="None" />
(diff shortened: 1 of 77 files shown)


zvonimir (approved):
Looks good to me. Thanks for doing this cleanup!

Source/StyleCop.ruleset
```diff
@@ -1,101 +1,101 @@
...
+    <Rule Id="SA1404" Action="None" />
+    <Rule Id="SA1405" Action="None" />
+    <Rule Id="SA1407" Action="None" />
+    <Rule Id="SA1408" Action="None" />
+    <Rule Id="SA1411" Action="None" />
+    <Rule Id="SA1413" Action="None" />
+    <Rule Id="SA1500" Action="None" />
+    <Rule Id="SA1501" Action="None" />
+    <Rule Id="SA1502" Action="None" />
+    <Rule Id="SA1503" Action="Error" />
```
zvonimir, line 80: Just a quick comment. I think that by default whichever rule is omitted here, it gets checked.
So why would you leave this one in (and set `Action` to `Error`) instead of simply omitting it?

zvonimir: Also, I found comments next to each rule to be kind of useful when trying to quickly figure out why something is failing. Why did you remove them? Thx!

keyboardDrummer: > Just a quick comment. I think that by default whichever rule is omitted here, it gets checked.
So why would you leave this one in (and set Action to Error) instead of simply omitting it?

The default is to emit a warning, not an error. But on second thought I think a warning is better. We can configure the PR build to treat warnings as errors, and then during development a warning won't block running the program.

> Also, I found comments next to each rule to be kind of useful when trying to quickly figure out why something is failing. Why did you remove them? Thx!

Oops, that was unintentional, thanks for spotting it!

shazqadeer (approved):

PR #449, opened 2021-11-09
Fix pruning bugs related to trigger

keyboardDrummer:

### Changes
- Fix bug in the I1 testcase in `Prune.bpl`, which prevented the test from failing when pruning wasn't working
- Fix several bugs in pruning related to triggers

### DIFF
--- a/Source/Core/Prune/DependencyEvaluator.cs
+++ b/Source/Core/Prune/DependencyEvaluator.cs
@@ -25,7 +25,7 @@ namespace Microsoft.Boogie
     public List<Declaration[]> incomingSets = new();
     public HashSet<Type> types = new();
     
-    private static bool ExcludeDep(Declaration declaration)
+    public static bool ExcludeDep(Declaration declaration)
     {
       return declaration.Attributes != null && QKeyValue.FindBoolAttribute(declaration.Attributes, "exclude_dep");
     }
@@ -39,7 +39,9 @@ namespace Microsoft.Boogie
 
     protected void AddOutgoing(Declaration newOutgoing)
     {
-      outgoing.Add(newOutgoing);
+      if (!ExcludeDep(newOutgoing)) {
+        outgoing.Add(newOutgoing);
+      }
     }
 
     protected void AddIncoming(Declaration[] declarations)
--- a/Source/Core/Prune/DependencyEvaluator.cs
+++ b/Source/Core/Prune/DependencyEvaluator.cs
@@ -39,9 +39,7 @@ namespace Microsoft.Boogie
 
     protected void AddOutgoing(Declaration newOutgoing)
     {
-      if (!ExcludeDep(newOutgoing)) {
-        outgoing.Add(newOutgoing);
-      }
+      outgoing.Add(newOutgoing);
     }
 
     protected void AddIncoming(Declaration[] declarations)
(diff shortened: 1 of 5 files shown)


Source/Graph/Graph.cs
```diff
@@ -1182,13 +1182,12 @@ public static IEnumerable<object> FindReachableNodesInGraphWithMergeNodes(Dictio
...
-        if (node is IReadOnlySet<object> mergeNode) {
-          if (!visitedEdges.IsSupersetOf(mergeNode)) {
+        if (node is IEnumerable<object> objects) {
+          if (!visitedEdges.IsSupersetOf(objects)) {
             continue;
           }
-        } else {
-          visitedEdges.Add(node);
         }
+        visitedEdges.Add(node);
```
MikaelMayer, line 1190: Accurate catching of a bug, where merge nodes would not be added if they were visitable.
Since the check is now against IEnumerable instead of IReadOnlySet, please update the comment accordingly.

Source/Core/Prune/DependencyEvaluator.cs
```diff
@@ -25,7 +25,7 @@ public class DependencyEvaluator : ReadOnlyVisitor
     public List<Declaration[]> incomingSets = new();
     public HashSet<Type> types = new();
     
-    private static bool ExcludeDep(Declaration declaration)
+    public static bool ExcludeDep(Declaration declaration)
```
MikaelMayer, line 28: What does it fix that private became public here? I suggest you add a comment to the function about why it needs to be public of it is not testable (e.g. for APIs), in case someone later wants to revert it.

MikaelMayer (changes requested):

Source/Core/Prune/DependencyEvaluator.cs
MikaelMayer: Ok, if you don't intend to merge this PR first since you are including it in #450, which makesuse of ExcludeDep, you can ignore my comment.

Test/test0/Prune.bpl
```diff
@@ -20,7 +20,7 @@ axiom (forall x: int ::
 
 procedure I1(x : int) returns (y: int)
   requires R(x);
-  ensures Q(f1(y)); // this post-condition doesn't prove because f1 is attributed as exclude_dep and
+  ensures Q(f1(x)); // this post-condition doesn't prove because f1 is attributed as exclude_dep and
                     // is thus removed from the outgoing set of I1.
                     // This makes the axiom on line 23 unreachable from I1, which is thus pruned away.
```
fabiomadge, line 25: 33?

fabiomadge, line 25:
```suggestion
                    // This makes the axiom on line 33 unreachable from I1, which is thus pruned away.
```

fabiomadge: I might be misunderstanding the wording here, but don't see an axiom in L23.

Source/Core/Prune/DependencyEvaluator.cs
fabiomadge: Pretty sure it can be `private` in #450 as well.

fabiomadge (approved):

robin-aws (approved):

PR #452, opened 2021-11-11
Allow keeping SMT commands unchanged when Boogie identifiers are renamed

keyboardDrummer:

This PR includes changes https://github.com/boogie-org/boogie/pull/453

SMT solvers in practice can show significantly different behavior in terms of total runtime and `check-sat` result when identifiers in the input are renamed. Such changes in solver output are often surprising to the Boogie programmer, so this PR introduces an option to normalize names in the Boogie program.

### Changes
- Enables normalizing any names that occur in the Boogie program when translating it into SMT commands.
- Remove unused code that was in SMTLibNamer and UniqueNamer

### Testing
- Add a test that checks that renaming various elements in a Boogie program does not effect the solver input.
- Add the `/normalizeNames:1` option to all tests that inspect the SMT model, to verify that names in the parsed SMT model are correctly translated back to their original name.


Source/UnitTests/ExecutionEngineTests/ProofIsolation.cs
```diff
@@ -11,6 +11,86 @@ namespace ExecutionEngineTests
...
+var alternative2: Wicket2;
+
+type Barrel2 a;
+type RGBColor2;
+const unique red2: RGBColor2;
+const unique green2: RGBColor2;
+const unique blue2: RGBColor2;
+
+const m2: [Barrel2 Wicket2] Wicket2;
+const n2: <a> [Barrel2 a] a;
```
MikaelMayer, line 65: Can you also rename the type parameter `a`, the procedure `x`, the input parameter `x` and the variable `y` ?

Source/UnitTests/ExecutionEngineTests/ProofIsolation.cs
```diff
@@ -11,6 +11,86 @@ namespace ExecutionEngineTests
...
+      CommandLineOptions.Clo.Parse(new string[]{});
+      CommandLineOptions.Clo.DiscardNames = true;
+      ExecutionEngine.printer = new ConsolePrinter();
+      
+      CommandLineOptions.Clo.ProcsToCheck.Add("M");
+      var proverLog1 = GetProverLogForProgram(procedure1);
+      var proverLog2 = GetProverLogForProgram(procedure2);
+      Assert.AreEqual(proverLog1, proverLog2);
+      
+      Console.Out.Write("");
```
MikaelMayer, line 91: Since this statement seems to do nothing, can you write a comment about why it is necessary? .E.g. "Flushes cache", or "Indicate to external service that we are done".

Source/VCGeneration/Checker.cs
```diff
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics.Contracts;
+using System.Linq;
```
MikaelMayer, line 4: Why is this import necessary since nothing else changed in the file?

MikaelMayer (approved):

MikaelMayer (changes requested):
Just minor changes.

shazqadeer:
@MikaelMayer :  Removing a lock does not appear a minor change to me.

shazqadeer:
> This PR enables discarding any names that occur in the Boogie program when translating it into SMT commands.

I could not understand the sentence above in the PR description.  Perhaps use extra sentences.

shazqadeer:
@RustanLeino :  I would appreciate your input on this PR.

Source/Provers/SMTLib/SmtLibNameUtils.cs
```diff
@@ -64,54 +69,20 @@ public class SMTLibNamer : UniqueNamer
 
     static HashSet<string> reservedSmtWords;
     static bool[] validIdChar;
-    static bool symbolListsInitilized;
 
-    static void InitSymbolLists()
+    static SmtLibNameUtils()
     {
-      lock (reservedSmtWordsList)
+      reservedSmtWords = new HashSet<string>();
```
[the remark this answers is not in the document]
keyboardDrummer: I moved this code to a static constructor and those are thread-safe by default.

Source/UnitTests/ExecutionEngineTests/ProofIsolation.cs
keyboardDrummer: Good call :)

Source/UnitTests/ExecutionEngineTests/ProofIsolation.cs
keyboardDrummer: I was a debugging remainder

Source/VCGeneration/Checker.cs
keyboardDrummer: It's not. Removed it.

keyboardDrummer:
> > This PR enables discarding any names that occur in the Boogie program when translating it into SMT commands.
> 
> I could not understand the sentence above in the PR description. Perhaps use extra sentences.

Done.

shazqadeer:
- The PR title talks about renaming which is an accurate description of the goals of the PR.  However, the PR description talks about discarding which is misleading and confusing.
- The option name DiscardNames is confusing and should be renamed to something that indicates that variables in the input are being renamed according to some scheme to product the SMT output.

I see that you are renaming names as n0, n1, ...  Did I get that right?  Is it still possible that a minor change to an input file would cause the names assigned to variables x and y to shift?  I imagine that the actual assignment of names to x and y would depend on some iteration order.  Changes to names other than x and y could cause the iteration order to change.

keyboardDrummer:
> * The PR title talks about renaming which is an accurate description of the goals of the PR.  However, the PR description talks about discarding which is misleading and confusing.
> * The option name DiscardNames is confusing and should be renamed to something that indicates that variables in the input are being renamed according to some scheme to product the SMT output.
> 
> I see that you are renaming names as n0, n1, ... Did I get that right? Is it still possible that a minor change to an input file would cause the names assigned to variables x and y to shift? I imagine that the actual assignment of names to x and y would depend on some iteration order. Changes to names other than x and y could cause the iteration order to change.

Good callout. The steps I'm adding to Boogie are these:

- Remove top level declarations that aren't used for the verification condition that's about the be verified
- Reorder top level declarations based on a content hash
- [Optionally] Rename top level declarations based on their order.

The combination of these allows solver input to remain unchanged, when changes unrelated to the verification condition being solved are made, and when declarations are reordered or renamed.

However, you are right that adding something that isn't pruned away might change the names of all the other declarations in the program, due to the naming by order system. One question is whether we should always try to minimise the amount of changes in solver input, or focus solely on providing zero solver input changes when possible. I'm doing the latter since that's the only thing which guarantees no change in solver behavior.

If you're adding something that cannot be pruned away, then a small change to solver input is unavoidable. That the renaming scheme makes this a bigger change is something I'm accepted, given the benefit of getting zero solver input changes when only doing a rename.

RustanLeino:
I think

>  focus solely on providing zero solver input changes when possible

is a good goal to focus on.

Source/Provers/SMTLib/SmtLibNameUtils.cs
```diff
@@ -64,54 +69,20 @@ public class SMTLibNamer : UniqueNamer
...
+        reservedSmtWords.Add(w);
       }
-    }
 
-    static string AddQuotes(string s)
-    {
-      var allGood = true;
-
-      foreach (char ch in s)
+      validIdChar = new bool[255];
```
RustanLeino, line 81: Why 255? I'm guessing it has to do with 255 being close to 2^8. But then, why not use all 256? Or is 255 just an arbitrary "pretty small" number?

Source/UnitTests/ExecutionEngineTests/ProofIsolation.cs
```diff
@@ -11,6 +11,111 @@ namespace ExecutionEngineTests
   [TestFixture]
   public class ProofIsolation
   {
+    [Test()]
+    public void TestNameDiscarding()
+    {
+      var procedure1 = @"
+type Wicket;
+const w : Wicket;
```
RustanLeino, line 19: Remove space after `:`

Source/Core/CommandLineOptions.cs
```diff
@@ -543,7 +543,8 @@ void ObjectInvariant2()
     public int VerifySnapshots = -1;
     public bool VerifySeparately = false;
     public string PrintFile = null;
-
+    public bool EmitDebugInformation { get; set;  } = true;
```
RustanLeino, line 546: If this is configurable, we should also have a command-line switch for it.

RustanLeino (changes requested):
I like this option. I have some questions that the code I read didn't answer for me:

* Are only top-level declarations renamed? Or does the renaming also apply to local variables, quantified variables, and incarnation variables?
* Boogie performs "lambda lifting", where it creates a function for each `lambda` binder in the given program. Do these play a part of the renaming?
* Under the renaming option, is the output affected? That is, will Boogie rename things back before given back any results to its callers? For example, if you want to use the counterexample display in the Dafny IDE, then will it still know what variables to ask for value for? (If the output is affected, then I have doubts about how useful this PR actually is to Boogie clients.)

Source/Provers/SMTLib/SmtLibNameUtils.cs
keyboardDrummer: I don't know, I didn't write this code, just moved it.

Source/Core/CommandLineOptions.cs
keyboardDrummer: Added it.

RustanLeino:
> I could look into why those names are maintained, but we can also leave it for later.

Don't worry about the types `T@T` and `T@U` and the function `tickleBool`, etc. These are always added by Boogie, so they will be the same for any input.

RustanLeino:
> The output was effected but I've since changed the ModelParser so it automatically reverts the renaming when recovering a model from Z3, this makes the counterexample display unchanged by this PR.

Excellent! Thanks.

RustanLeino (dismissed):

MikaelMayer (approved):

Source/Core/CommandLineOptions.cs
```diff
@@ -1656,6 +1687,14 @@ protected override bool ParseOption(string name, CommandLineOptionEngine.Command
         case "kInductionDepth":
           ps.GetNumericArgument(ref KInductionDepth);
           return true;
+        
+        case "emitDebugInformation":
+          ps.GetNumericArgument(ref emitDebugInformation);
+          return true;
+        
+        case "discardNames":
```
shazqadeer, line 1695: I find the option name "discardNames" confusing.  I have indicated several times before in my feedback that the verb "discard" is inappropriate for what this PR is trying to achieve.

shazqadeer (changes requested):

Source/Core/CommandLineOptions.cs
keyboardDrummer: Hey, sorry for not responding to your earlier comment.

Do you have a suggestion for a different name? Your earlier comment seems to indicate something with `rename`, but I find that `renameNames` does not say enough about what the option does.

The goal of the option is to get verification output that does not depend on original program names and is deterministic. I picked `discardNames` to indicate that the original program names are not used, leaving the reader to conclude that they must be replaced using some fixed naming scheme. 

Here are some ideas for other names:
`replaceNames`, `discardAndReplaceNames`, `ignoreNames`, or `renameBasedOnContent`

Do you have a preference or suggestion?

bkragl: Maybe `normalizeNames`?

shazqadeer: normalizeNames appears to be the best option so far.

keyboardDrummer: Great, made the change :)

shazqadeer (approved):

PR #453, opened 2021-11-11
Enable turning off skolemid

keyboardDrummer:
To prevent verification behavior from changing unexpectedly, it can help not to emit skolemid and qid, since these contain an incrementally assigned number and a source code location, respectively, that can easily change for the entire SMT program when small changes occur in the input Boogie program.

### Testing

- Add a tests that inspects whether the solver logs contain the right information when `EmitDebugInformation` is either turned on or off.
- Manually tried the `/emitDebugInformation:1` and `/emitDebugInformation:0` command line options.

### DIFF
--- a/Source/Core/CommandLineOptions.cs
+++ b/Source/Core/CommandLineOptions.cs
@@ -205,6 +205,20 @@ namespace Microsoft.Boogie
       /// If there is one argument and it is a non-negative integer, then set "arg" to that number and return "true".
       /// Otherwise, emit error message, leave "arg" unchanged, and return "false".
       /// </summary>
+      public bool GetNumericArgument(ref bool arg)
+      {
+        int intArg = 0;
+        var result = GetNumericArgument(ref intArg, x => x < 2);
+        if (result) {
+          arg = intArg != 0;
+        }
+        return result;
+      }
+      
+      /// <summary>
+      /// If there is one argument and it is a non-negative integer, then set "arg" to that number and return "true".
+      /// Otherwise, emit error message, leave "arg" unchanged, and return "false".
+      /// </summary>
       public bool GetNumericArgument(ref int arg)
       {
         //modifies nextIndex, encounteredErrors, Console.Error.*;
@@ -543,8 +557,16 @@ namespace Microsoft.Boogie
     public int VerifySnapshots = -1;
     public bool VerifySeparately = false;
     public string PrintFile = null;
-    public bool EmitSkolemIdAndQId = true;
-    
+
+    /**
+     * Whether to emit {:qid}, {:skolemid} and set-info :boogie-vc-id
+     */
+    public bool EmitDebugInformation
+    {
+      get => emitDebugInformation;
+      set => emitDebugInformation = value;
+    }
+
     public int PrintUnstructured {
       get => printUnstructured;
       set => printUnstructured = value;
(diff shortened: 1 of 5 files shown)


MikaelMayer (dismissed):

Source/Core/CommandLineOptions.cs
```diff
@@ -543,7 +543,8 @@ void ObjectInvariant2()
     public int VerifySnapshots = -1;
     public bool VerifySeparately = false;
     public string PrintFile = null;
-
+    public bool EmitSkolimIdAndQId = true;
```
shazqadeer, line 546: Skolim -> Skolem

shazqadeer: Please make this change everywhere.

Source/Provers/SMTLib/SMTLibLineariser.cs
```diff
@@ -459,7 +459,7 @@ public bool Visit(VCExprQuantifier node, LineariserOptions options)
         if (hasAttrs)
         {
           wr.Write("\n");
-          if (info.qid != null)
+          if (info.qid != null && CommandLineOptions.Clo.EmitSkolemIdAndQId)
```
shazqadeer, line 462: It looks like it is possible for info.qid to be null in which case nothing will be emitted.  So, there already appears to be a way to control whether something is emitted.  Then what is the purpose of the new option?

keyboardDrummer: > So, there already appears to be a way to control whether something is emitted.

Judging by the code's, there's not. There's only a few edge cases in which it's null but you can't trigger it to be null everywhere.

RustanLeino: I think `info.qid` is set only if the Boogie program uses the `{:qid ...}` attribute.

RustanLeino (changes requested):
I like the option, but I think the option should also be available from the command line.

Source/Provers/SMTLib/SMTLibLineariser.cs
keyboardDrummer: You can see [here](https://github.com/boogie-org/boogie/blob/master/Source/VCExpr/Boogie2VCExpr.cs#L690) that Boogie generates a `{:qid}` if none is specified:

```
      if (qid == null && vars.Count != 0)
      {
        // generate default name (line:column position in .bpl file)
```

Also Dafny does not generate any `{:qid}` attributes but the resolving SMT logs still contain them.

keyboardDrummer:
> I like the option, but I think the option should also be available from the command line.

Why? What's a command-line use-case, besides testing Boogie?

RustanLeino:
> > I like the option, but I think the option should also be available from the command line.
> 
> What would be the use-case for that? I'm only aware of the testing use-case for using Boogie's command line.
> 
> I think downstream tools should explicitly decide whether to expose this option through the command line instead of automatically inheriting it from Boogie.

The use case is that a Dafny developer (like ourselves) is debugging something. We then use the `/print` flag to output the Boogie program. Then, we want to run `boogie` directly on the generated Boogie program, which means we need to be able to invoke `boogie` with the same settings as Dafny sets internally.

keyboardDrummer:
> The use case is that a Dafny developer (like ourselves) is debugging something. We then use the `/print` flag to output the Boogie program. Then, we want to run `boogie` directly on the generated Boogie program, which means we need to be able to invoke `boogie` with the same settings as Dafny sets internally.

And I guess changing the Boogie code by hand is much faster than doing so in the Boogie generating code of the downstream tool, right? I'll add the option.

RustanLeino (approved):

shazqadeer (approved):

PR #454, opened 2021-11-14
Content hash based reordering

keyboardDrummer:
Reorders top level axioms based on a hash of their content. This enables reordering Boogie members without effecting solver input.

Based on #452


Source/Core/SMTLibOptions.cs
```diff
@@ -20,5 +20,13 @@ public interface SMTLibOptions
     bool SIBoolControlVC { get; }
     bool RestartProverPerVC { get; }
     bool Trace { get; }
+    
+    /**
+     * Setting this to true will rename all identifiers in the Boogie program to a generated name that does not depend on the original name.
```
RustanLeino, line 25: "all identifiers" refers to both global and local names, right?

Source/Core/SMTLibOptions.cs
```diff
@@ -20,5 +20,13 @@ public interface SMTLibOptions
...
+     * Discarding the original names is useful to prevent the solver input from changing when identifiers are renamed in the
+     * Boogie program, which prevents unexpected changes in solver output.
+     */
+    bool DiscardNames { get; }
```
RustanLeino, line 29: Is there any reason _not_ to always do this transformation? It seems good for everyone, for perhaps this should just always be on. (Or, if this causes the output to use the generated names, then I'm doubtful that anyone could afford to use it.)

RustanLeino (dismissed):
I love this feature.

Well, I _think_ I love it. I'm basing that love on the believe that this does not change the output. Is that right? For example, with this feature engaged, will the Dafny counterexample display in the IDE continue working without change?

Also, I'm clicking "Approve", but I expect the changes to this PR's dependent PRs to be addressed first.

Source/Core/SMTLibOptions.cs
keyboardDrummer: Jep

RustanLeino (dismissed):

shazqadeer:
In PR #452, the goal of this PR is described as:

> Reorder top level declarations based on a content hash.

However, the description of this PR says:

> Reorders top level axioms based on a hash of their content.

Are you reordering only axioms or all declarations?  Please make the descriptions consistent.

keyboardDrummer:
> In PR #452, the goal of this PR is described as:
> 
> > Reorder top level declarations based on a content hash.
> 
> However, the description of this PR says:
> 
> > Reorders top level axioms based on a hash of their content.
> 
> Are you reordering only axioms or all declarations? Please make the descriptions consistent.

I'm only reordering the axioms based on a content hash. Reordering the axioms turns out to be enough since the order in which top level functions and constants are translated to SMT is not based on their declaration order, but by the order in which they're discovered when traversing the axioms.

shazqadeer (approved):

robin-aws (approved):
Re-approving after previous approvals were dismissed by resolving merge conflicts.

shazqadeer (approved):

PR #458, opened 2021-11-23
Improved documentation about errorLimit

shazqadeer:


Source/Core/CommandLineOptions.cs
```diff
@@ -2404,7 +2404,8 @@ be n (default 500)
 
   /errorLimit:<num>
                 Limit the number of errors produced for each procedure
-                (default is 5, some provers may support only 1)
+                (default is 5, some provers may support only 1).
+                Set num to 0 to find as many assertion failures as possible.
```
zvonimir, line 2408: Is `0` the default for "infinity" for other options as well?
Or is it MAX_INT?

shazqadeer: I don't know.  I doubt there is any uniform policy being adhered to.

zvonimir (approved):
Looks good to me.

PR #459, opened 2021-11-26
feat: Robustness against prover crash on (reset)

MikaelMayer:
This PR provides backward compatible support for Z3 4.8.5 on Windows, which sometimes would crash when receiving the `(reset)` command - this happened in the Dafny test suite (https://github.com/dafny-lang/dafny/pull/1540).
This bug was corrected in Z3 4.8.8, but I believe this feature can be our defense-in-depth strategy for dealing with any unexpected Z3 reset crash in the future.

## Changes

- `SMTLibProcessTheoremProver.cs` Added a check right after the `(reset)` command via a `HasProverDied()` command to the prover. If it was not the first (reset) sent, it restarts a new fresh process.
- `SMTLibProcess.cs` Added the method `HasProverDied()` which invokes `PingPong` and returns the exception if one was thrown.

### DIFF
--- a/Source/Provers/SMTLib/SMTLibProcess.cs
+++ b/Source/Provers/SMTLib/SMTLibProcess.cs
@@ -125,6 +125,20 @@ namespace Microsoft.Boogie.SMTLib
       return sx != null && sx.Name == ":name";
     }
 
+    public ProverDiedException HasProverDied()
+    {
+      try
+      {
+        PingPong();
+      }
+      catch (ProverDiedException e)
+      {
+        return e;
+      }
+
+      return null;
+    }
+    
     public void PingPong()
     {
       Ping();
--- a/Source/Provers/SMTLib/SMTLibProcess.cs
+++ b/Source/Provers/SMTLib/SMTLibProcess.cs
@@ -125,7 +125,7 @@ namespace Microsoft.Boogie.SMTLib
       return sx != null && sx.Name == ":name";
     }
 
-    public ProverDiedException HasProverDied()
+    public ProverDiedException GetExceptionIfProverDied()
     {
       try
       {
--- a/Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
+++ b/Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
@@ -47,6 +47,7 @@ namespace Microsoft.Boogie.SMTLib
     private VCExpressionGenerator gen;
     protected readonly SMTLibProverOptions options;
     private bool usingUnsatCore;
+    private int resetCount = 0;
 
     [ContractInvariantMethod]
     void ObjectInvariant()
@@ -129,9 +130,9 @@ namespace Microsoft.Boogie.SMTLib
       }
     }
 
-    void SetupProcess()
+    void SetupProcess(bool recover = false)
     {
-      if (Process != null)
+      if (Process != null && !recover)
       {
         return;
       }
@@ -145,8 +146,7 @@ namespace Microsoft.Boogie.SMTLib
       if (Process != null && Process.NeedsRestart)
       {
         Process.Close();
-        Process = null;
-        SetupProcess();
+        SetupProcess(true);
         Process.Send(common.ToString());
       }
     }
@@ -628,6 +628,7 @@ namespace Microsoft.Boogie.SMTLib
       {
         this.gen = gen;
         SendThisVC("(reset)");
+        RecoverIfProverCrashedAfterReset();
         SendThisVC("(set-option :" + Z3.RlimitOption + " 0)");
 
         if (0 < common.Length)
@@ -641,6 +642,23 @@ namespace Microsoft.Boogie.SMTLib
         }
 
         HasReset = true;
+      }
+    }
+
+    private void RecoverIfProverCrashedAfterReset()
+    {
+      resetCount += 1;
+      if (Process.HasProverDied() is { } e)
+      {
+        if (resetCount == 1)
+        {
+          // First-time reset cannot be recovered
+          throw e;
+        }
+
+        Process.Close();
+        SetupProcess(true);
+        resetCount = 1;
       }
     }
 
--- a/Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
+++ b/Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
@@ -47,7 +47,6 @@ namespace Microsoft.Boogie.SMTLib
     private VCExpressionGenerator gen;
     protected readonly SMTLibProverOptions options;
     private bool usingUnsatCore;
-    private int resetCount = 0;
 
     [ContractInvariantMethod]
     void ObjectInvariant()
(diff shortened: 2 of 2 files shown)


Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
```diff
@@ -129,9 +130,9 @@ private void SetupAxiomBuilder(VCExpressionGenerator gen)
       }
     }
 
-    void SetupProcess()
+    void SetupProcess(bool recover = false)
```
keyboardDrummer, line 133: It seems like you don't need the additional `recover` argument if you run `Process = null;` before calling `SetupProcess` on line 660.

Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
```diff
@@ -644,6 +645,23 @@ public override void Reset(VCExpressionGenerator gen)
...
     }
 
+    private void RecoverIfProverCrashedAfterReset()
+    {
+      resetCount += 1;
+      if (Process.HasProverDied() is { } e)
+      {
+        if (resetCount == 1)
+        {
+          // First-time reset cannot be recovered
```
keyboardDrummer, line 655: Why not?

Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
```diff
@@ -644,6 +645,23 @@ public override void Reset(VCExpressionGenerator gen)
       }
     }
 
+    private void RecoverIfProverCrashedAfterReset()
+    {
+      resetCount += 1;
+      if (Process.HasProverDied() is { } e)
```
keyboardDrummer, line 651: I think I would find `is Exception e` easier to read than `is { } e`

Source/Provers/SMTLib/SMTLibProcess.cs
```diff
@@ -125,6 +125,20 @@ public bool IsPong(SExpr sx)
       return sx != null && sx.Name == ":name";
     }
 
+    public ProverDiedException HasProverDied()
```
keyboardDrummer, line 128: The name of this method implies that it returns a boolean.

Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
MikaelMayer: Because at this point, only `(reset)` was sent to the solver. if it crashes there, recovering it would mean spawning a new solver instance and sending it more `(reset)` command, which could loop indefinitely.
In practice this then-branch will never be entered, but it's there to prove termination (in the unlikely case a solver is corrupted for example)

Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
MikaelMayer: I followed here the principle Don't Repeat Yourself. Looking at this pattern used twice, It thought it would be better to refactor it:
`Process = null; SetupProcess();` ==> `SetupProcess(/*recover*/true)`
I see several reasons that this enhances the readability and maintainability
- Having a statement controlling the inner if of another hidden statement is error-prone and hard to follow.
- If Process is implemented in a different way down the road, we will have to change all statements before call sites
- Process is assigned twice in the first statement (once null, once the new process), but what we want here is just discard the previous process and only one assignment can be made.

Because of the purpose of this PR is not maintainability, if you tell me that you prefer the first version, I'd happily make the change if this gets this PR accepted more easily.

Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
MikaelMayer: That's was the IDE semi-auto correction (it underlined in yellow `ProverDiedException e` when I wrote it), which was annoying.
Thanks to your suggestion, I'm writing `Exception e` and it does not complain anymore and indeed it's easier to read.

Source/Provers/SMTLib/SMTLibProcess.cs
MikaelMayer: Method name changed to `GetExceptionIfProverDied`

Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
```diff
@@ -644,6 +645,23 @@ public override void Reset(VCExpressionGenerator gen)
...
+      {
+        if (resetCount == 1)
+        {
+          // First-time reset cannot be recovered
+          throw e;
+        }
+
+        Process.Close();
+        SetupProcess(true);
+        resetCount = 1;
```
keyboardDrummer, line 661: At the start of `RecoverIfProverCrashedAfterReset` this variable is incremented, so assigning 1 to it, would mean that next time it's already 2, right? If you want to reset this value shouldn't you set it to 0 ?

Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
keyboardDrummer: Why would another reset command be sent after the recovery?

Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
keyboardDrummer: I suggest inlining the

```
if (Process != null && !recover)
      {
        return;
      }
```

Part of the method, since only one call passes `recover == false`.

Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
MikaelMayer: Next time, it's already 2, right, but there is a good reason for that.
The code is catching errors that happen right after a `(reset)` command is issued, and no other errors. Currently, Boogie sends this command when the process starts for the first time, and that's the non-recoverable `(reset)` command that would be caught and thrown again.
Other than that, if a `(reset)` throw an exception after this case, it means that the prover already did some processing and it's really after that that the `(reset)` command failed. So we just want to restart the prover and not even send a `(reset)` command that would need to be checked again.

Sorry for the confusion, I could have just put a boolean there or just remove this counter alltogether (after all, if a reset at the beginning fails, there isn't anything that we could do)

Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
MikaelMayer: You're right, actually, no `(reset)` command is sent after the recover. I think I can simplify my code then.

Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
MikaelMayer: That's a good point. Since there is never a point where we call `SetupProcess` "just in case", I removed all ifs and just assume that if the process was there, it should be restarted anyway with this call.

keyboardDrummer (approved):

Issue #460, opened 2021-11-30
/xml feature breaks when verifying multiple VCs in parallel


robin-aws:
The `XmlSink` object isn't used in a thread safe way. For example:

```
> boogie Test/commandline/noProc.bpl /vcsCores:2 -xml:noProc.xml
Unhandled exception. System.AggregateException: One or more errors occurred. ('startTime' is a duplicate attribute name.) (The Writer is closed or in error state.) (The Writer is closed or in error state.) (The Writer is closed or in error state.) ('startTime' is a duplicate attribute name.) (The Writer is closed or in error state.) (The Writer is closed or in error state.) (The Writer is closed or in error state.)
 ---> System.Xml.XmlException: 'startTime' is a duplicate attribute name.
   at System.Xml.XmlWellFormedWriter.AddAttribute(String prefix, String localName, String namespaceName)
   at System.Xml.XmlWellFormedWriter.WriteStartAttribute(String prefix, String localName, String namespaceName)
   at System.Xml.XmlWriter.WriteAttributeString(String localName, String value)
   at Microsoft.Boogie.XmlSink.WriteStartMethod(String methodName, DateTime startTime) in /home/runner/work/boogie/boogie/Source/Core/Xml.cs:line 87
   at Microsoft.Boogie.ExecutionEngine.VerifyImplementation(Program program, PipelineStatistics stats, ErrorReporterDelegate er, String requestId, Dictionary`2 extractLoopMappingInfo, Implementation[] stablePrioritizedImpls, Int32 index, OutputCollector outputCollector, List`1 checkers, String programId) in /home/runner/work/boogie/boogie/Source/ExecutionEngine/ExecutionEngine.cs:line 1190
  ...

PR #461, opened 2021-11-30
Add splits to xml

robin-aws:
This information is currently included in the output of `/trace` but not the XML. Including it there too will enable more programmatic analysis.

### DIFF
--- a/Source/Core/Xml.cs
+++ b/Source/Core/Xml.cs
@@ -112,6 +112,7 @@ namespace Microsoft.Boogie
     public void WriteStartSplit(int splitNum, DateTime startTime)
     {
       Contract.Requires(splitNum > 0);
+      Contract.Requires(startTime != null);
       Contract.Requires(IsOpen);
       //modifies this.*;
       Contract.Ensures(IsOpen);
@@ -128,6 +129,7 @@ namespace Microsoft.Boogie
     public void WriteEndSplit(string outcome, TimeSpan elapsed)
     {
       Contract.Requires(outcome != null);
+      Contract.Requires(elapsed != null);
       Contract.Requires(IsOpen);
       //modifies this.*;
       Contract.Ensures(IsOpen);
--- a/Source/Core/Xml.cs
+++ b/Source/Core/Xml.cs
@@ -112,7 +112,6 @@ namespace Microsoft.Boogie
     public void WriteStartSplit(int splitNum, DateTime startTime)
     {
       Contract.Requires(splitNum > 0);
-      Contract.Requires(startTime != null);
       Contract.Requires(IsOpen);
       //modifies this.*;
       Contract.Ensures(IsOpen);
@@ -129,7 +128,6 @@ namespace Microsoft.Boogie
     public void WriteEndSplit(string outcome, TimeSpan elapsed)
     {
       Contract.Requires(outcome != null);
-      Contract.Requires(elapsed != null);
       Contract.Requires(IsOpen);
       //modifies this.*;
       Contract.Ensures(IsOpen);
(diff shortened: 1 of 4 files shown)


Source/Core/Xml.cs
```diff
@@ -109,6 +109,41 @@ public void WriteEndMethod(string outcome, DateTime endTime, TimeSpan elapsed)
       cce.EndExpose();
     }
 
+    public void WriteStartSplit(int splitNum, DateTime startTime)
+    {
+      Contract.Requires(splitNum > 0);
```
keyboardDrummer, line 114: Are these contracts really useful? I thought everything but `Contract.Assert` has no runtime or compile time effect, and I would argue even the `Contract.Assert` is more noisy than useful.

Source/VCGeneration/SplitAndVerifyWorker.cs
```diff
@@ -108,6 +108,10 @@ private void StartCheck(Split split, Checker checker)
           split.Stats, currentSplitNumber + 1, total, 100 * provenCost / (provenCost + remainingCost));
       }
 
+      if (options.XmlSink != null && DoSplitting) {
```
keyboardDrummer, line 111: Seems like something like an `interface SplitLogger { void StartSplit(..); void EndSplit(..)  }` with a ConsoleSplitLogger, XMLSplitLogger, and AggregateSplitLogger implementation could be useful to make the code a bit cleaner, but I can see that requires quite some refactoring which you may want to postpone.

keyboardDrummer (approved):

Source/VCGeneration/SplitAndVerifyWorker.cs
robin-aws: Totally agree, and yes not something I wanted to take on in this PR. Likely makes more sense to do that when fixing #460.

Source/Core/Xml.cs
robin-aws: Yup, I was just following the established pattern for now. I do find they are useful documentation even if they are never enforced.

As an aside, I was pleasantly surprised that the build rejected my attempt to add null checks that were unnecessary, but I'm now confused as to how the project is using the new nullability rules since I can't find any `<Nullable>` elements in the configuration.

robin-aws:
@keyboardDrummer neither of us noticed my test didn't pass in CI, but I've fixed it now - can I get a fresh approval?

@shazqadeer this is ready for your review now too.

keyboardDrummer (approved):

Source/VCGeneration/Split.cs
```diff
@@ -1412,6 +1412,11 @@ public void ReadOutcome(ref ConditionGeneration.Outcome curOutcome, out bool pro
             checker.ProverRunTime.TotalSeconds, outcome);
         }
 
+        if (CommandLineOptions.Clo.XmlSink != null) {
```
rakamaric, line 1415: How come you don't need ` && splitNum >= 0` like in the if condition above?

rakamaric (changes requested):

Source/VCGeneration/Split.cs
robin-aws: The `WriteEndSplit` method doesn't actually refer to `splitNum` (since that's already recorded in the parent `<split ...>` element), unlike the tracing output. I have a slight preference for still recording the data even if something weird happens and `splitNum < 0`, but I don't feel too strongly and could add the condition if you'd prefer.

rakamaric: Got it. Makes sense.

rakamaric (approved):
Looks good to me.

PR #464, opened 2021-12-02
Remove automatic prune mode

keyboardDrummer:
Please review the first commit separately from the rest, since the first commit moves code into different files.
First commit: https://github.com/boogie-org/boogie/pull/464/commits/1ad9fd29cb97ce711c411407c34d95689f69d6c5
Rest: https://github.com/boogie-org/boogie/pull/464/files/1ad9fd29cb97ce711c411407c34d95689f69d6c5..420bba4545abf2fe9d505d2cd33a167045bc2e48

### Changes

- Remove Automatic prune mode. `{:exclude_dep}`, which was useful with automatic prune mode, is replaced by `{:include_dep}` which is useful is the remaining prune mode.
- For debugging purposes, add the option to print a Boogie program after it is pruned
- Add a missing check for triggers in exist clauses when computing incoming axiom dependencies
- Remove where clause pruning. This was causing issues with a test in Dafny and it's not something Dafny currently needs, or is tested in Boogie. We may add it back later.


Source/Core/CommandLineOptions.cs
```diff
@@ -590,80 +591,58 @@ public bool EmitDebugInformation
...
+     * 
+     * When verifying FMultipliesByTwo, pruning will remove G and its axiom, but not F and its axiom.
+     *
+     * Axioms defined in a uses clause have an incoming edge from the clause's declaration.
+     * Uses clauses can be placed on functions and constants.
+     * 
+     * Adding the {:include_dep} attribute to an axiom will give it an incoming edge from each declaration that it references.
+     * The {:include_dep} attribute is useful in a migration scenario.
+     * When turning on pruning in a Boogie program with many axioms,
+     * one many add {:include_dep} to all of them to prevent pruning too much,
```
RustanLeino, line 634:
```suggestion
     * one may add {:include_dep} to all of them to prevent pruning too much,
```

Source/Core/CommandLineOptions.cs
```diff
@@ -590,80 +591,58 @@ public bool EmitDebugInformation
...
+     *
+     * ```
+     * function F(int) returns (int) uses {
+     *   axiom forall x: int :: F(x) == x * 2
+     * }
+     * function G(int) returns (int) uses {
+     *   axiom forall x: int :: G(x) == F(x) + 1
+     * }
+     *
+     * procedure FMultipliesByTwo(x: int)
```
RustanLeino, line 621:
```suggestion
     * procedure FMultipliedByTwo(x: int)
```

Source/Core/CommandLineOptions.cs
```diff
@@ -590,80 +591,58 @@ public bool EmitDebugInformation
...
+     *   ensures F(x) - x == x
+     * { }
+     * ```
+     * 
+     * When verifying FMultipliesByTwo, pruning will remove G and its axiom, but not F and its axiom.
```
RustanLeino, line 626:
```suggestion
     * When verifying FMultipliedByTwo, pruning will remove G and its axiom, but not F and its axiom.
```

Source/Core/CommandLineOptions.cs
```diff
@@ -590,80 +591,58 @@ public bool EmitDebugInformation
...
+     * When turning on pruning in a Boogie program with many axioms,
+     * one many add {:include_dep} to all of them to prevent pruning too much,
+     * and then iteratively remove {:include_dep} attributes and add uses clauses to enable pruning more.
+     *
+     * Apart from uses clauses and {:include_dep}, incoming edges are also created for axioms that contain triggers.
+     * If a quantifier with a trigger occurs in an axiom, then no incoming edges are determined from the body of the quantifier,
+     * regardless of {:include_dep} being present. However, for each trigger of the quantifier,
+     * each declaration referenced in the trigger has an outgoing edge to a merge node,
+     * that in turn has an outgoing edge to the axiom.
+     * The merge node is only traversable in the reachable analysis if each of its incoming edges has been reached.
```
RustanLeino, line 642:
```suggestion
     * The merge node is traversable in the reachable analysis only if each of its incoming edges has been reached.
```

Source/Core/CommandLineOptions.cs
```diff
@@ -1197,6 +1176,14 @@ protected override bool ParseOption(string name, CommandLineOptionEngine.Command
 
           return true;
 
+        case "printPruned":
```
RustanLeino, line 1179: Describe this option in the `/help` message.

Also, describe `{:include_dep}` (and delete the description for `{:exclude_dep}`) in the `/attrHelp` message in this file.

RustanLeino (changes requested):
Looks great! My only comments (other than some typos) are to write the help text for `{:include_dep}` and `/printPruned`.

Source/Core/CommandLineOptions.cs
keyboardDrummer: The description in the comment is ~50 lines long, that doesn't seem to fit well in help. Would a summary of the description be useful?

robin-aws (dismissed):
Acknowledging that Rustan's minor suggestions were addressed on his behalf.

Source/Core/CommandLineOptions.cs
```diff
@@ -2416,10 +2398,8 @@ be n (default 500)
                 only for monomorphic programs.
   /reflectAdd   In the VC, generate an auxiliary symbol, elsewhere defined
                 to be +, instead of +.
-  /prune:<n>
-                0 (default) - none
-                1 - automatic pruning
-                2 - aggressive pruning. Requires binding axioms to functions and constants using 'uses'
+  /prune
```
rakamaric, line 2406: We should probably summarize here what pruning does.

rakamaric: Also, it seems that you did not add the print pruned options to help.

rakamaric (changes requested):

Source/Core/CommandLineOptions.cs
keyboardDrummer: I'm not sure how to add a useful summary. I can describe the use-case in a few sentences but the full explanation is 50 lines and I think that's too much for the generic help.

`/printPruned` is described in the help right?

```
/printPruned:<file>
                After pruning, print the Boogie program to the specified file.
```

rakamaric: How about simply:
```
/prune
     Remove any top-level Boogie declarations that are not accessible by the implementation that is about to be verified.
```
Even just this I think gives us a good hint for what is going on here.

keyboardDrummer: That's in there now

rakamaric (approved):
Looks good to me.

PR #468, opened 2021-12-06
Feature: Splitting on every assert

### LINKED ISSUE #465: Feature request: /vcsSplitOnEveryAssert
As part of the epic on verification stability (https://github.com/dafny-lang/dafny/issues/1582), I've found that the implementation of the `{:split_here}` and `{:focus}` directives can easily be reused to isolate every assertion in its own split, just as if `{:split_here}` was present on every assertion. I have a POC implementation that has proven useful as a debugging tool, to identify which asserts in a procedure were by far the most expensive to verify. Focusing on improving those assertions then greatly improved the stability of the procedure.

The equivalent `{:vcs_split_on_every_assert}` attribute would also be useful as a blunt instrument for improving the stability of a procedure: although it would increase the total verification time quite a bit because of the overhead, it would also greatly improve the stability of the verification.

robin-aws:
Resolves #465

See linked issue for details.

Note the known issue that this simple implementation (behaving as if every assertion had `{:split_here}`) results in the first split VC always having no assertions. I'm opting not to address that in the name of avoiding additional complexity, since it doesn't affect the usefulness of the feature.


keyboardDrummer (dismissed):

Test/test0/SplitOnEveryAssert.bpl
```diff
@@ -0,0 +1,42 @@
...
+// CHECK:      checking split 6/12, .*
+// CHECK:      checking split 7/12, .*
+// CHECK:      checking split 8/12, .*
+// CHECK:      checking split 9/12, .*
+// CHECK:      checking split 10/12, .*
+// CHECK:      checking split 11/12, .*
+// CHECK:      checking split 12/12, .*
+// CHECK-L: SplitOnEveryAssert.bpl(30,5): Error: This assertion might not hold.
+
+procedure {:vcs_split_on_every_assert} Ex() returns (y: int)
```
rakamaric, line 19: Maybe adding another procedure to this regression test that does not have this attribute would be useful. I guess it would also be nice to have a regression that uses `/vcsSplitOnEveryAssert`.

robin-aws: Yeah I was on the fence about that too - added!

keyboardDrummer (approved):

rakamaric (approved):
Looks good to me.

PR #472, opened 2021-12-20
Add noop prover that always returns unknown.

keyboardDrummer:
### Changes

Add a 'noop' prover that always returns unknown. Useful for getting prover logs quickly without having to wait for a solver

Can be turned on by passing `-proverOpt:SOLVER=noop`. Updated documentation to include this.

### Testing

Added a test that uses the 'noop' solver.

### DIFF
--- a/Source/Provers/SMTLib/FunctionDependencyCollector.cs
+++ b/Source/Provers/SMTLib/FunctionDependencyCollector.cs
@@ -1,0 +1,33 @@
+using System.Collections.Generic;
+using Microsoft.Boogie.VCExprAST;
+
+namespace Microsoft.Boogie.SMTLib;
+
+public class FunctionDependencyCollector : BoundVarTraversingVCExprVisitor<bool, bool>
+{
+  private List<Function> functionList;
+
+  // not used but required by interface
+  protected override bool StandardResult(VCExpr node, bool arg)
+  {
+    return true;
+  }
+
+  public List<Function> Collect(VCExpr expr)
+  {
+    functionList = new List<Function>();
+    Traverse(expr, true);
+    return functionList;
+  }
+
+  public override bool Visit(VCExprNAry node, bool arg)
+  {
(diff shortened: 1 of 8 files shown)


Source/Provers/SMTLib/FunctionDependencyCollector.cs
```diff
@@ -0,0 +1,33 @@
+using System.Collections.Generic;
+using Microsoft.Boogie.VCExprAST;
+
+namespace Microsoft.Boogie.SMTLib;
+
+public class FunctionDependencyCollector : BoundVarTraversingVCExprVisitor<bool, bool>
```
keyboardDrummer, line 6: Class moved into its own file, but is unchanged.

Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
```diff
@@ -11,35 +11,6 @@
 
 namespace Microsoft.Boogie.SMTLib
 {
-  public class FunctionDependencyCollector : BoundVarTraversingVCExprVisitor<bool, bool>
```
keyboardDrummer, line 14: Class was moved to another file

Source/Provers/SMTLib/SMTLibProverOptions.cs
```diff
@@ -102,7 +106,7 @@ protected override bool Parse(string opt)
             Solver = SolverKind.YICES2;
             break;
           default:
-            ReportError("Invalid SOLVER value; must be 'Z3' or 'CVC5' or 'Yices2'");
+            ReportError("Invalid SOLVER value; must be 'Z3' or 'CVC5', 'Yices2' or 'noop'");
```
RustanLeino, line 109:
```suggestion
            ReportError("Invalid SOLVER value; must be 'Z3' or 'CVC5' or 'Yices2' or 'noop'");
```

RustanLeino (approved):
I like it! Good option to have.

RustanLeino (approved):

PR #480, opened 2022-01-10
Include split locations in XML output

atomb:
This includes a location in addition to an outcome for each `split` entry in the XML result report.


Source/VCGeneration/Split.cs
```diff
@@ -69,6 +69,7 @@ void ObjectInvariant()
 
 
       private readonly List<Block> blocks;
+      public IEnumerable<AssertCmd> Asserts => blocks.SelectMany(block => block.cmds.OfType<AssertCmd>());
```
keyboardDrummer, line 72: This only seems to be used by the class its defined in and only once. If that's the case, consider moving this to a local variable.

Source/VCGeneration/Split.cs
```diff
@@ -139,6 +140,18 @@ void ObjectInvariant()
         }
       }
 
+      public IToken AssertToken
```
keyboardDrummer, line 143: This operation does some work: it traverses the list of commands in the block. I would reflect that by making this a method instead of a property: `GetAssertToken()`

keyboardDrummer (dismissed):

Source/Core/Xml.cs
```diff
@@ -121,6 +121,14 @@ public void WriteStartSplit(int splitNum, DateTime startTime)
         wr.WriteStartElement("split");
         wr.WriteAttributeString("number", splitNum.ToString());
         wr.WriteAttributeString("startTime", startTime.ToString(DateTimeFormatString));
+
+        if (token != null) {
+          wr.WriteStartElement("location");
```
robin-aws, line 126: I'm a bit worried about attaching a single "location" to splits: it's obviously useful if you happened to use `/vcsSplitOnEveryAssert`, but it could be very confusing if you're using `{split_here}` and/or `{:focus}` and happen to end up with some splits with only one assert in them. It would be more general to log the location of EVERY assertion in a split IMO, but that would increase the XML size substantially if it was on all the time.

Let's discuss the options a bit more before merging this.

robin-aws (changes requested):

Source/Core/Xml.cs
atomb: Good call. I was thinking primarily of the `/vcsSplitOnEveryAssert` case, and this could perhaps be a little strange in other cases.

atomb: Just for experimentation, I pushed a commit that includes locations for every assert within a split, which seems, on further consideration, like the most informative thing for a user. It _doesn't_ list anything when there are no splits (since it wouldn't add any information; you'll know the proof includes all assertions at once).

Source/Core/Xml.cs
```diff
@@ -126,16 +127,30 @@ public void WriteSplit(int splitNum, DateTime startTime, string outcome, TimeSpa
 
       cce.BeginExpose(this);
       {
-        wr.WriteStartElement("split");
+        wr.WriteStartElement("assertionBatch");
```
robin-aws, line 130: Just for the record, this will be a breaking change for any consumers processing this XML, but I'm fine with that. It's such an easy change to account for, and I'm not aware of any consumers besides Dafny processing this output.

robin-aws (approved):

Source/Core/Xml.cs
atomb: Yeah, it's true. Dafny will need to change, but it'll be easy. I take it that you're happy with the terminology?

PR #481, opened 2022-01-12
Add randomSeed option to Boogie

keyboardDrummer:
### Changes

- Add a `randomSeed` option to Boogie that takes an integer argument which is used to randomize arbitrary decisions in the Boogie pipeline and solving process. When a randomSeed is passed, Boogie will: randomise all names in the generated SMTLib commands; randomize the order in which commands are generated; and set randomisation options in the solver if Z3 is used.
- Adding the `{:random_seed <n>}` attribute to an implementation has the same behavior as setting the `/randomSeed:<n>` option, but only for that implementation.
- Fix a bug in the normaliseDeclarationOrder command line option

### Testing
- Add a test that verifies the /randomSeed option will reorder declarations, rename variables, and set Z3 randomisation options.
- Add a test for the normaliseDeclarationOrder command line option


robin-aws:
Nice! A couple of high level comments right off the bat before I review the implementation.

I would mention the motivation for this in the PR description (and the eventually squashed commit), and possibly in the option documentation - I totally get that it's connected to measuring instability but good to make that clearer for future historians. :)

I don't like having just a single option to both enable the randomness and set the seed, especially since folks will be used to the Z3 behavior (always being random, and setting a seed of 0 being equivalent to the default behavior). I think you have a few options:

1. Have the randomization logic always on, but explicitly set a random seed of 0 by default. That still makes Boogie "deterministic", but could easily lead to a lot of verification regressions for users.
2. Special-case `/randomSeed:0` to maintain the current behavior. Think of it as "what a weird coincidence, my random shuffle put everything in the same order" :)
3. Provide two options, something like `/randomize` and `/randomSeed`.

I think my preference is 2 but could be convinced otherwise.

Source/VCGeneration/Checker.cs
```diff
@@ -190,45 +190,61 @@ private void SetRandomSeed(int? randomSeed)
...
             ctx.DeclareGlobalVariable(glVarDecl, null);
           }
         }
       }
     }
 
+    private static IEnumerable<Declaration> GetReorderedDeclarations(IEnumerable<Declaration> declarations, Random random)
+    {
+      if (random == null) {
+        // By ordering the declarations based on content and naming them based on order, the solver input stays content under reordering and renaming.
```
cpitclaudel, line 237: `content` or `constant`?

Source/VCGeneration/Split.cs
```diff
@@ -117,21 +116,21 @@ void ObjectInvariant()
...
         this.parent = par;
-        this.impl = impl;
+        this.Implementation = implementation;
         Interlocked.Increment(ref currentId);
 
-        PrintTopLevelDeclarationsForPruning(impl, "before");        
+        PrintTopLevelDeclarationsForPruning(implementation, "before");        
         TopLevelDeclarations = Prune.GetLiveDeclarations(par.program, blocks).ToList();
-        PrintTopLevelDeclarationsForPruning(impl, "after");
+        PrintTopLevelDeclarationsForPruning(implementation, "after");
```
cpitclaudel, line 133: I'm usually not a fan of this kind of renaming, since it pollutes the history of the codebase without much benefit.

cpitclaudel (approved):
Looks good to me.  Some thoughts:

- I don't fully understand how names are generated by boogie currently, and reading the diff isn't enough to get a good sense of that.  I could dive deeper, but I don't think it's necessary.

- With the current implementation you can't look at the SMT input and recover the original names; this will make more painful to look at generated SMT files, but it's not a huge issue since this setting defaults to off.  However, you could have randomized names in a different way (for example using a random substitution cipher) to get an invertible transform.

Source/VCExpr/RandomiseNamer.cs
```diff
@@ -0,0 +1,39 @@
...
+  }
+    
+  public override UniqueNamer Clone()
+  {
+    return new RandomiseNamer(this);
+  }
+
+  protected override string GetModifiedName(string uniqueInherentName)
+  {
+    return $"random{random.Next()}";
```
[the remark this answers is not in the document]
robin-aws: I think collisions are handled elsewhere, in `ScopedNamer.GetName`, but I admit I'm still trying to follow it. The hierarchy could possibly use a few more comments.

robin-aws (commented):
Still reviewing

Source/VCExpr/RandomiseNamer.cs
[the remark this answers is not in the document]
keyboardDrummer: Yes they're handled in the ScopedNamer.

Source/VCGeneration/Split.cs
keyboardDrummer: The field had to capitalise its first letter because it's public now, and then I also made the name longer.

Source/Core/CommandLineOptions.cs
```diff
@@ -2452,8 +2453,17 @@ be n (default 500)
                 aggressive and it may create an exponential number of splits.
 
   /randomSeed:<n>
-                Randomise arbitrary decisions during verification condition generation, 
-                and during solving   
+                Note: Boogie output is deterministic in the input, whether this
```
cpitclaudel, line 2456: *Boogie's?

Source/Core/CommandLineOptions.cs
```diff
@@ -2452,8 +2453,17 @@ be n (default 500)
                 aggressive and it may create an exponential number of splits.
 
   /randomSeed:<n>
-                Randomise arbitrary decisions during verification condition generation, 
-                and during solving
```
cpitclaudel: I would keep a high-level explanation first, rather than starting with a note.  Maybe something like this:

> Turn on randomization in the input that Boogie passes to the SMT solver and in the SMT solver itself.  

… and then most of the text that you had below

Source/Core/CommandLineOptions.cs
```diff
@@ -2443,6 +2452,19 @@ be n (default 500)
                 a linear number of splits. The default way (top-down) is more
                 aggressive and it may create an exponential number of splits.
 
+  /randomSeed:<n>
+                Note: Boogie output is deterministic in the input, whether this
+                option is used or not.
+ 
+                Certain Boogie inputs are unstable in the sense that changes to 
+                the input that preserve its meaning may cause the output to change.
+                The /randomSeed option simulates meaning preserving changes to
```
robin-aws, line 2461:
```suggestion
                The /randomSeed option simulates meaning-preserving changes to 
```

Source/Core/CommandLineOptions.cs
```diff
@@ -2443,6 +2452,19 @@ be n (default 500)
...
+                option is used or not.
+ 
+                Certain Boogie inputs are unstable in the sense that changes to 
+                the input that preserve its meaning may cause the output to change.
+                The /randomSeed option simulates meaning preserving changes to 
+                the input without requiring the user to actually make those changes.
+
+                The /randomSeed option is implemented by renaming variables and 
+                reordering declarations in the input, and by setting 
+                solver options that have similar affects.
```
robin-aws, line 2466:
```suggestion
                solver options that have similar effects.
```

Source/Core/CommandLineOptions.cs
```diff
@@ -2041,6 +2049,7 @@ public int GetArgumentSeparatorIndex(string argList, int startIndex)
 
      {:random_seed N}
```
robin-aws, line 2050: This is a bit of an API gap, since this sets the Z3 random seed options as well. If you have a method that happens to verify for a non-default random seed, and you want to temporarily set it in source, you can't do that without also enabling the randomization in Boogie.

Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
```diff
@@ -48,9 +48,15 @@ void ObjectInvariant()
       InitializeGlobalInformation();
       SetupAxiomBuilder(gen);
 
-      Namer = libOptions.NormalizeNames ? new NormalizeNamer() : new KeepOriginalNamer();
+      Namer = (options.RandomSeed, libOptions.NormalizeNames) switch
+      {
+        (null, true) => new NormalizeNamer(),
+        (null, false) => new KeepOriginalNamer(),
+        _ => new RandomiseNamer(new Random(options.RandomSeed.Value))
+      };
```
robin-aws, line 56: MUCH better, thank you :)

Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
```diff
@@ -65,6 +71,11 @@ void ObjectInvariant()
         PrepareCommon();
       }
     }
+    
+    private ScopedNamer ResetNamer(ScopedNamer namer) {
+      return options.RandomSeed == null ? libOptions.NormalizeNames ? new NormalizeNamer(namer) : new KeepOriginalNamer(namer) :
```
robin-aws, line 76: The `switch` alternative would work nicely here too.

robin-aws (commented):
Finished reviewing - I mainly focussed on the UX rather than the implementation, but Clement has already covered the latter with comments well.

Source/Core/CommandLineOptions.cs
keyboardDrummer: The verification result

Source/Core/CommandLineOptions.cs
keyboardDrummer: > If you have a method that happens to verify for a non-default random seed, and you want to temporarily set it in source, you can't do that without also enabling the randomization in Boogie.

You can, you just set this attribute.

The `{:random_seed N}` attribute, whose affect has been modified in this PR, will have the same affect for a single implementation as setting the /randomSeed option has. That's the whole point of it! I've made the documentation more explicit.

cpitclaudel (dismissed):

Test/pruning/RandomSeed.bpl
```diff
@@ -0,0 +1,16 @@
+// RUN: %parallel-boogie -infer:j /proverLog:"%t.log" /normalizeDeclarationOrder:0 /randomSeed:10000 "%s" > "%t"
```
robin-aws, line 1: Can we add a copy of this test that uses `{:randomSeed}` as well?

Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
```diff
@@ -540,14 +555,14 @@ public override void BeginCheck(string descriptiveName, VCExpr vc, ErrorHandler
...
-
       if (options.Solver == SolverKind.Z3)
       {
         SendThisVC("(set-option :" + Z3.TimeoutOption + " " + options.TimeLimit + ")");
         SendThisVC("(set-option :" + Z3.RlimitOption + " " + options.ResourceLimit + ")");
         if (options.RandomSeed.HasValue)
         {
-          SendThisVC("(set-option :" + Z3.RandomSeedOption + " " + options.RandomSeed.Value + ")");
+          SendThisVC("(set-option :" + Z3.SmtRandomSeed + " " + options.RandomSeed.Value + ")");
+          SendThisVC("(set-option :" + Z3.SatRandomSeed + " " + options.RandomSeed.Value + ")");
```
robin-aws, line 565: This change could cause verification regressions as well, right? I can deal with that personally since this isn't a frequently used feature, and we're not committing to zero "regressions" (yet :), but if someone's code is broken by this is there an easy workaround?

Source/Core/CommandLineOptions.cs
robin-aws: I meant if you find that a command like this succeeds:

```
dafny /compile:0 MyFile.dfy /proc:"*MyMethod" /proverOpt:O:smt.random_seed=12345
```

Then adding `{:randomSeed 12345}` won't necessarily verify, since it also perturbs the Boogie translation (and sets `sat.random_seed` as well now). Is there an attribute you can use to set arbitrary solver options per-procedure?

robin-aws (changes requested):
Just a couple lingering API concerns, although I might just be misunderstanding.

Source/Core/CommandLineOptions.cs
keyboardDrummer: I see. 

> Is there an attribute you can use to set arbitrary solver options per-procedure?

I don't think so.

Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
keyboardDrummer: The workaround would be to find a new random seed that works :-)

Source/VCGeneration/Checker.cs
```diff
@@ -387,297 +381,6 @@ public ProverInterface.Outcome ReadOutcome()
   // -----------------------------------------------------------------------------------------------
   // -----------------------------------------------------------------------------------------------
 
-  public abstract class ProverInterface
```
keyboardDrummer, line 374: Apart from being moved this class only has had the method `SetRandomSeed` removed.

Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
```diff
@@ -73,7 +73,7 @@ void ObjectInvariant()
     }
     
     private ScopedNamer ResetNamer(ScopedNamer namer) {
-      return (options.RandomSeed, libOptions.NormalizeNames) switch
+      return (RandomSeed: options.RandomSeed, libOptions.NormalizeNames) switch
```
cpitclaudel, line 76: What does that label do?

Source/Provers/SMTLib/SMTLibProverOptions.cs
```diff
@@ -30,7 +30,7 @@ public enum SolverKind
     Z3,
     CVC5,
     YICES2,
-    NoOp
+    NoOpWithZ3Options
```
cpitclaudel, line 33: This is a bit odd.  Why the name change?

Source/VCGeneration/Checker.cs
```diff
@@ -184,26 +185,20 @@ private void SetRlimit(uint rlimit)
...
     
     /// <summary>
     /// Set up the context.
     /// </summary>
-    private void Setup(Program prog, ProverContext ctx, Split s = null)
+    private void Setup(Program prog, ProverContext ctx, Split split = null)
     {
-      var implementationRandomSeed = s?.Implementation.RandomSeed;
-      var random = implementationRandomSeed == null ? CommandLineOptions.Clo.Random : new Random(implementationRandomSeed.Value);
+      solverOptions.RandomSeed = split?.Implementation.RandomSeed ?? CommandLineOptions.Clo.RandomSeed;
```
cpitclaudel, line 194: Here we are overwriting the class-level `solverOptions`; is that to allow setting per-procedure random seeds?  AFAICT this method doesn't otherwise mutate `this` (or does it?)

cpitclaudel (approved):

Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
keyboardDrummer: Nothing, removed it.

Source/Provers/SMTLib/SMTLibProverOptions.cs
keyboardDrummer: To indicate that Z3 specific options should be set when using this solver.

Source/VCGeneration/Checker.cs
keyboardDrummer: > is that to allow setting per-procedure random seeds?

Yes, a checker is only used for one split at a time, and its settings get updated if a new split requires that.

Source/Core/CommandLineOptions.cs
robin-aws: Okay, I'm happy moving forward, thanks for clarifying.

robin-aws (approved):

PR #482, opened 2022-01-14
First cut at putting Z3 resource counts in XML output

atomb:
Adopted directly from previous work by @robin-aws.


Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -1269,8 +1262,12 @@ private static void CleanupCheckers(string requestId)
 
       if (CommandLineOptions.Clo.XmlSink != null)
       {
-        CommandLineOptions.Clo.XmlSink.WriteEndMethod(verificationResult.Outcome.ToString().ToLowerInvariant(),
-          verificationResult.End, verificationResult.End - verificationResult.Start);
+        lock (CommandLineOptions.Clo.XmlSink) {
```
robin-aws, line 1265: I'm pretty sure this was the start of an attempt to make `/xml` thread safe, but I would revert it and keep this PR focussed on just adding the resource count, and punt on #460 for now

Source/Houdini/Houdini.cs
```diff
@@ -1439,7 +1439,7 @@ protected void DebugRefutedCandidates(Implementation curFunc, List<Counterexampl
         }
 
         DateTime end = DateTime.UtcNow;
-        xmlRefuted.WriteEndMethod("errors", end, end.Subtract(start));
+        xmlRefuted.WriteEndMethod("errors", end, end.Subtract(start), 0);
```
robin-aws, line 1442: Note I'm NOT confident in this since I don't yet understand how Houdini works.

Source/ExecutionEngine/ExecutionEngine.cs
atomb: Done.

Source/Houdini/Houdini.cs
atomb: The `resourceCount` argument is now nullable, and no `resourceCount` attribute gets added to the XML if it's null. This should lead to no changes in Houdini XML output.

keyboardDrummer:
Can we have a test where we check for a concrete value of resources? This value is supposed to constant across runs right?

atomb:
> Can we have a test where we check for a concrete value of resources? This value is supposed to constant across runs right?

It seems to be constant at least when running on the same platform with the same random seed. I added concrete values to the test. Let's see if CI confirms that it's stable between platforms.

atomb:
The test with the concrete resource count included is now failing, and I suspect it's because the CI uses a different version of Z3 than I have locally. I could update the test to match the output of Z3 4.8.8, but then it would be pretty fragile and would pin us more strongly to a specific Z3 version.

keyboardDrummer:
> The test with the concrete resource count included is now failing, and I suspect it's because the CI uses a different version of Z3 than I have locally. I could update the test to match the output of Z3 4.8.8, but then it would be pretty fragile and would pin us more strongly to a specific Z3 version.

Ok. How about a test that verifies it's the same when you do two runs?

Test/commandline/xml.bpl
```diff
@@ -1,7 +1,11 @@
 // Can't use %parallel-boogie here yet - see https://github.com/boogie-org/boogie/issues/460
-// RUN: %boogie -xml:"%t.xml" "%s"
+// RUN: %boogie -randomSeed:0 -xml:"%t-1.xml" "%s"
+// RUN: %boogie -randomSeed:0 -xml:"%t-2.xml" "%s"
+// RUN: grep -Eo "resourceCount=\"[0-9]+\"" "%t-1.xml" > "%t-res1"
+// RUN: grep -Eo "resourceCount=\"[0-9]+\"" "%t-2.xml" > "%t-res2"
+// RUN: diff "%t-res1" "%t-res2"
```
robin-aws, line 6: Nice! :)

robin-aws (approved):

robin-aws:
Thought of something immediately after approving. :)

In my original branch I was also including the resource count in the `/trace` output. We could fast-follow with that if it's uncontroversial?

rakamaric (approved):
Looks good to me.

atomb:
> Thought of something immediately after approving. :)
> 
> In my original branch I was also including the resource count in the `/trace` output. We could fast-follow with that if it's uncontroversial?

Sounds good to me.

atomb:
There's a question of terminology in the text. How does "Z3 ticks" sound? For example

````
[0.224 s, 553604 Z3 ticks, 24 proof obligations]
````

keyboardDrummer:
> 553604 Z3 ticks

Sounds good to me :-)

PR #486, opened 2022-01-21
Include resource counts in /trace output

atomb:


Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -1446,8 +1446,10 @@ private static string TimeIndication(VerificationResult verificationResult)
       var result = "";
       if (CommandLineOptions.Clo.Trace)
       {
-        result = string.Format("  [{0:F3} s, {1} proof obligation{2}]  ",
-          (verificationResult.End - verificationResult.Start).TotalSeconds, verificationResult.ProofObligationCount,
+        result = string.Format("  [{0:F3} s, {1} Z3 ticks, {2} proof obligation{3}]  ",
```
robin-aws, line 1449: This might not be Z3 though. As far as I could tell earlier "rlimit" is used as a solver-agnostic term, but I couldn't find any definitive definition of what "r" expands to.

I do like your term of "ticks", but we should do our due diligence to see if there's a commonly-used term out there in the field already. And whatever we do pick, we should probably rename the various uses of `ResourceCount` in the code at the same time to be less confusing.

robin-aws: I notice CVC4 also accepts `--rlimit` and even uses the term "resource count": https://cvc4.github.io/getting-started.html

The only downside is using that term in this form: "42 resource count" reads weird, and "42 resources" is right out. :)

atomb: Yeah, I think I would have used "resource count" or "resources", but they do read weird. :) And "rlimit" sounds weird as a result value as opposed to a limit.

Making sure the code uses the same terminology as the messages sounds like a good plan, too.

robin-aws: I think since this is just console output and we can improve it later, I'd be happy with something like `resource count = 42` for now to avoid the awkward grammar. We can revisit if we want to define a different concept in the future.

atomb: I ultimately fell back on "resource count", leaving the naming discussion for the future.

robin-aws (approved):

Issue #487, opened 2022-01-21
Exception in solver interaction leads to false report of validity


atomb:
If exceptions arise during interaction with a solver, `VCGen.VerifyImplementation` can sometimes silently and erroneously return `Correct`. In particular, I've seen this arise twice:

* When incorrectly implementing a new feature in `NoopSolver` that led to an exception in `SMTLibProcessTheoremProver.GetRCount`.
* When attempting to use Dafny with Yices and configuring it incorrectly.

You can replicate this by putting the following in an executable file named `badz3`:
````
#!/bin/bash
echo "dfksal"
````

And then running Boogie with the prover set to that script, something like this:
````
dotnet Source/BoogieDriver/bin/Debug/net6.0/BoogieDriver.dll /trace /proverOpt:PROVER_PATH=`pwd`/badz3 Test/test2/Arrays.bpl
````

This leads to the following output:

````
Parsing Test/test2/Arrays.bpl
Coalescing blocks...
Inlining...
[TRACE] Using prover: /Users/aarotomb/Repositories/boogie/badz3

Verifying P0 ...
  [0.101 s, 2 proof obligations]  verified

Verifying P1 ...
  [0.001 s, 1 proof obligation]  verified

Verifying P2 ...
  [0.000 s, 1 proof obligation]  verified

Verifying Q0 ...
  [0.001 s, 2 proof obligations]  verified

Verifying Q1 ...
  [0.000 s, 1 proof obligation]  verified

Verifying Q2 ...
  [0.000 s, 1 proof obligation]  verified

Verifying Q3 ...
  [0.000 s, 1 proof obligation]  verified

Verifying Q4 ...
  [0.000 s, 1 proof obligation]  verified

Verifying Skip0 ...
  [0.000 s, 2 proof obligations]  verified

Verifying Skip1 ...
  [0.001 s, 2 proof obligations]  verified

Boogie program verifier finished with 10 verified, 0 errors
````

When running with a working Z3, it reports 8 verified, 2 errors.

I believe that the issue arises at least in part from the following line:
https://github.com/boogie-org/boogie/blob/da30caa77454a296878056ae5ed7dd4414a9f2ea/Source/VCGeneration/VCGen.cs#L839

[...]

PR #488, opened 2022-01-21
Don't silently swallow exceptions when verifying

atomb:
(Edited to address my own comment below).

Re-throw exceptions that occur during the split and verify worker, rather than silently swallowing them. Avoids printing "valid" when exceptions occur during solver interaction.

Addresses #487.

### DIFF
--- a/Source/ExecutionEngine/ExecutionEngine.cs
+++ b/Source/ExecutionEngine/ExecutionEngine.cs
@@ -1251,12 +1251,21 @@ namespace Microsoft.Boogie
             verificationResult.Errors = null;
             verificationResult.Outcome = VCGen.Outcome.Inconclusive;
           }
-          catch(AggregateException e)
+          catch(AggregateException ae)
           {
-            printer.AdvisoryWriteLine("Advisory: {0} SKIPPED because: {1}",
-              impl.Name, e.Message);
-            verificationResult.Errors = null;
-            verificationResult.Outcome = VCGen.Outcome.SolverException;
+            ae.Flatten().Handle(e =>
+            {
+              if (e is IOException)
+              {
+                printer.AdvisoryWriteLine("Advisory: {0} SKIPPED due to I/O exception: {1}",
+                  impl.Name, e.Message);
+                verificationResult.Errors = null;
+                verificationResult.Outcome = VCGen.Outcome.SolverException;
+                return true;
+              }
+
+              return false;
+            });
           }
 
           verificationResult.ProofObligationCountAfter = vcgen.CumulativeAssertionCount;
--- a/Source/ExecutionEngine/ExecutionEngine.cs
+++ b/Source/ExecutionEngine/ExecutionEngine.cs
@@ -1589,6 +1589,14 @@ namespace Microsoft.Boogie
           }
 
           break;
+        case VCGen.Outcome.Uninitialized:
+          if (implName != null && implTok != null)
+          {
+            errorInfo = errorInformationFactory.CreateErrorInformation(implTok,
+              "Verification result not initialized.");
+          }
+
+          break;
       }
 
       if (errorInfo != null)
@@ -1634,6 +1642,9 @@ namespace Microsoft.Boogie
           break;
         case VCGen.Outcome.Inconclusive:
           traceOutput = "inconclusive";
+          break;
+        case VCGen.Outcome.Uninitialized:
+          traceOutput = "not yet initialized (internal error)";
           break;
         case VCGen.Outcome.Errors:
           Contract.Assert(errors != null);
@@ -1711,6 +1722,9 @@ namespace Microsoft.Boogie
             Interlocked.Add(ref stats.CachedErrorCount, cnt);
           }
 
+          break;
+        case VCGen.Outcome.Uninitialized:
+          // There's currently no counter for internal errors.
           break;
       }
     }
--- a/Source/ExecutionEngine/ExecutionEngine.cs
+++ b/Source/ExecutionEngine/ExecutionEngine.cs
@@ -1244,7 +1244,7 @@ namespace Microsoft.Boogie
             verificationResult.Errors = null;
             verificationResult.Outcome = VCGen.Outcome.Inconclusive;
           }
-          catch(Exception e)
+          catch(AggregateException e)
           {
             printer.AdvisoryWriteLine("Advisory: {0} SKIPPED because: {1}",
               impl.Name, e.Message);
--- a/Source/ExecutionEngine/ExecutionEngine.cs
+++ b/Source/ExecutionEngine/ExecutionEngine.cs
@@ -133,6 +133,11 @@ namespace Microsoft.Boogie
         Console.Write(", {0} out of resource", stats.OutOfResourceCount);
       }
 
+      if (stats.SolverExceptionCount != 0)
+      {
+        Console.Write(", {0} solver exceptions", stats.SolverExceptionCount);
+      }
+
       Console.WriteLine();
       Console.Out.Flush();
     }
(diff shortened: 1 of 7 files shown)


atomb:
An alternative (or additional?) approach could be to address what I think is perhaps a separate part of the problem:

https://github.com/boogie-org/boogie/blob/da30caa77454a296878056ae5ed7dd4414a9f2ea/Source/VCGeneration/SplitAndVerifyWorker.cs#L97-L109

This releases the checker silently if any exception occurs during checking.

atomb:
I'm not sure this is ready to go yet, but I'm adding reviewers in the hope of identifying the correct solution.

atomb:
The current fix now seems pretty reasonable to me, and it's very small. It does cause two current tests to fail because the expected error gets caught in a different place, leading to different output.

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -1244,6 +1244,13 @@ private static void CleanupCheckers(string requestId)
             verificationResult.Errors = null;
             verificationResult.Outcome = VCGen.Outcome.Inconclusive;
           }
+          catch(AggregateException e)
+          {
+            printer.AdvisoryWriteLine("Advisory: {0} SKIPPED because: {1}",
+              impl.Name, e.Message);
+            verificationResult.Errors = null;
+            verificationResult.Outcome = VCGen.Outcome.Inconclusive;
```
keyboardDrummer, line 1252: Could we set the default of verificationResult.Outcome to Inconclusive? Right now the default is Outcome.Correct

Source/VCGeneration/SplitAndVerifyWorker.cs
```diff
@@ -105,6 +105,7 @@ async Task DoWork(Split split)
       }
       catch {
         split.ReleaseChecker();
+        throw; // This can be better handled farther up the call stack.
```
keyboardDrummer, line 108: Nitpick: I don't think the comment is necessary. The catch being here only to release the checker makes sense to me.

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -1244,6 +1244,13 @@ private static void CleanupCheckers(string requestId)
             verificationResult.Errors = null;
             verificationResult.Outcome = VCGen.Outcome.Inconclusive;
           }
+          catch(AggregateException e)
```
keyboardDrummer, line 1247: Can we catch something more specific? I wouldn't want to swallow any bugs here but catching a bad solver argument seems good.

keyboardDrummer, line 1247: Can we catch something more specific? I wouldn't want to swallow any bugs here but catching a bad solver argument seems good.

Source/ExecutionEngine/ExecutionEngine.cs
keyboardDrummer, line 1252: Could we set the default of verificationResult.Outcome to Inconclusive? Right now the default is Outcome.Correct, which seems like asking for soundness issues.

keyboardDrummer, line 1252: Inconclusive means that the solver ran but couldn't prove unsat or sat right? For the scenario of invalid solver arguments I wouldn't mind an additional enum value, like `InvalidConfiguration`.

Source/VCGeneration/SplitAndVerifyWorker.cs
keyboardDrummer, line 108: Nitpick: I don't think the comment is necessary.

Source/ExecutionEngine/ExecutionEngine.cs
atomb: Oh, yeah, definitely! My original fix did set it to something else (a new constructor) and I was somewhat dissatisfied with losing that. But I've just checked and the test suite still passes if it's initialized to `Inconclusive`. 🎉

Source/ExecutionEngine/ExecutionEngine.cs
atomb: There's nothing below `AggregateException` in the class hierarchy, AFAIK, though we could pull out the wrapped exceptions and behave differently depending on what they are. This code will always lead to a message and an inconclusive result, so I wouldn't expect it to hide bugs. It could be that, after some experience with it, we realize that it would be best to handle different exceptions differently, but at the moment I'm not sure what the different behavior would be.

At the moment, this does catch bad solver arguments and paths, but also things like a broken pipe if a solver crashes, or exceptions thrown by internal code that relates to solver interaction (like failing to parse a solver result).

Source/ExecutionEngine/ExecutionEngine.cs
atomb: Ah, but the Dafny test suite completely fails with this change. I'll need to dig deeper to figure out what's going on there.

atomb: And so does the Boogie test suite. I was just running the wrong one previously. The bottom line is that the following code doesn't actually change the outcome when a proof succeeds:

https://github.com/boogie-org/boogie/blob/6251cd8ae3a87866f31d705fe8c30ed57c053dab/Source/VCGeneration/Split.cs#L1306-L1309

I think that solving this correctly would require a substantially different approach to tracking the outcome (which I'd like to do, but I'm not yet sure the best way to do). There's no single value that could be assigned at that would be correct in this location.

My current best guess at a solution would be to change it so that Boogie constructs a list of (proof obligation, result) pairs, and builds the overall result out of a combination of those. That'll require some substantial work, though.

So, my inclination would be to merge this, along with your other two suggestions (a new outcome alternative and removed comment) but to leave that refactoring to a future PR (which I will do, since it'll be really nice to have for other future work).

keyboardDrummer: Sounds good, thanks :-)

Source/ExecutionEngine/ExecutionEngine.cs
keyboardDrummer: Inspecting the internals of `AggregateException` and rethrowing if it doesn't contain something we want to handle, seems fine to me.

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -1784,6 +1818,10 @@ private static ErrorInformation CreateErrorInformation(Counterexample error, VC.
       {
         cause = "Out of memory on";
       }
+      else if (outcome == VCGen.Outcome.SolverException)
+      {
+        cause = "Solver exception on";
```
keyboardDrummer, line 1823: From a user perspective, I think I like to know "is it my fault?" and "solver exception" seems a bit ambiguous is that regard. It sounds like it's the solver's fault, but this problem also occurs when you give incorrect solver arguments right?

Source/ExecutionEngine/ExecutionEngine.cs
atomb: That sounds good to me. I'll update it to look inside the `AggregateException`.

atomb: One possibility is to _only_ handle `IOException` here. Solver configuration issues result in a `ProverException` which is then handled higher up, meaning the output expected by the current solver configuration error tests doesn't change.

Source/ExecutionEngine/ExecutionEngine.cs
atomb: With the most recent change, this is better behaved. An invalid solver path or option will lead to the old behavior: quitting immediately with a fatal error. A solver that starts up and then crashes will lead to the new behavior: a warning and a result of "inconclusive", allowing the user to still get results for any proof obligation on which the solver succeeds.

Source/ExecutionEngine/ExecutionEngine.cs
atomb: The latest commit does just this, and I think I like the way it behaves now (as detailed in the response below).

keyboardDrummer (approved):

PR #491, opened 2022-01-28
feat: Ability to kill verification

MikaelMayer:
This PR enables Boogie's execution engine to kill verification on demand.

**Changes**
This PR adds cancellation token checks inside the loop that listens to the prover's output.
It enables `ExecutionEngine.CancelRequest(requestId), which triggers the `CancellationTokenSource` to cancel, to abruptly kills the while loop where the SMTLibProcess waits for the Z3 process's output. (because Z3 can actually loop forever or for very long in some cases, especially when there is no timeout)

This alone should be able to fix the following bugs:
https://github.com/dafny-lang/ide-vscode/issues/119
https://github.com/dafny-lang/ide-vscode/issues/114
https://github.com/dafny-lang/ide-vscode/issues/96

### DIFF
--- a/Source/Core/ProverUtil.cs
+++ b/Source/Core/ProverUtil.cs
@@ -4,6 +4,7 @@ using System.Collections.Generic;
 using System.Reflection;
 using System.Text;
 using System.Diagnostics.Contracts;
+using System.Threading;
 
 namespace Microsoft.Boogie
 {
@@ -301,7 +302,8 @@ The generic options may or may not be used by the prover plugin.
   {
     // Really returns ProverInterface.
     //public abstract object! SpawnProver(ProverOptions! options, object! ctxt);
-    public abstract object SpawnProver(SMTLibOptions libOptions, ProverOptions options, object ctxt);
+    public abstract object SpawnProver(SMTLibOptions libOptions, ProverOptions options, object ctxt,
+      CancellationToken cancellationToken);
 
     // Really returns ProverContext
     public abstract object /*!*/ NewProverContext(ProverOptions /*!*/ options);
--- a/Source/ExecutionEngine/ExecutionEngine.cs
+++ b/Source/ExecutionEngine/ExecutionEngine.cs
@@ -1007,7 +1007,7 @@ namespace Microsoft.Boogie
               {
                 cts.Token.ThrowIfCancellationRequested();
               }
-
+              // Here we guarantee that requestId is bound to a CancellationTokenSource inside RequestIdToCancellationTokenSource
               VerifyImplementation(program, stats, er, requestId, extractLoopMappingInfo, stablePrioritizedImpls,
                 taskIndex, outputCollector, checkerPool, programId);
               ImplIdToCancellationTokenSource.TryRemove(id, out old);
@@ -1042,6 +1042,7 @@ namespace Microsoft.Boogie
       }
       catch (AggregateException ae)
       {
+        ThrowIfCanceled(ae);
         ae.Flatten().Handle(e =>
         {
           if (e is ProverException)
@@ -1124,6 +1125,14 @@ namespace Microsoft.Boogie
       return outcome;
     }
 
+    private static void ThrowIfCanceled(AggregateException ae) {
+      var canceledException =
+        ae.Flatten().InnerExceptions.FirstOrDefault(exception => exception is OperationCanceledException, null);
+      if (canceledException != null) {
+        throw canceledException;
+      }
+    }
+
     public static void CancelRequest(string requestId)
     {
       Contract.Requires(requestId != null);
@@ -1148,13 +1157,13 @@ namespace Microsoft.Boogie
       {
         if (RequestIdToCancellationTokenSource.IsEmpty)
         {
-          checkerPool.Dispose();
+          checkerPool?.Dispose();
           checkerPool = null;
         }
       }
     }
 
-
+    // Requires requestId to be bound to a CancellationTokenSource in RequestIdToCancellationTokenSource
     private static void VerifyImplementation(Program program, PipelineStatistics stats, ErrorReporterDelegate er,
       string requestId, Dictionary<string, Dictionary<string, Block>> extractLoopMappingInfo,
       Implementation[] stablePrioritizedImpls, int index, OutputCollector outputCollector, CheckerPool checkerPool,
@@ -1196,7 +1205,9 @@ namespace Microsoft.Boogie
 
         verificationResult = new VerificationResult(requestId, impl, programId);
 
-        using (var vcgen = CreateVCGen(program, checkerPool))
+        var cancellationToken = RequestIdToCancellationTokenSource[requestId].Token;
+
+        using (var vcgen = CreateVCGen(program, checkerPool, cancellationToken))
         {
           vcgen.CachingActionCounts = stats.CachingActionCounts;
           verificationResult.ProofObligationCountBefore = vcgen.CumulativeAssertionCount;
@@ -1251,8 +1262,8 @@ namespace Microsoft.Boogie
             verificationResult.Errors = null;
             verificationResult.Outcome = VCGen.Outcome.Inconclusive;
           }
-          catch(AggregateException ae)
-          {
+          catch(AggregateException ae) {
+            ThrowIfCanceled(ae);
             ae.Flatten().Handle(e =>
             {
               if (e is IOException)
@@ -1349,9 +1360,9 @@ namespace Microsoft.Boogie
       }
     }
 
-    private static ConditionGeneration CreateVCGen(Program program, CheckerPool checkerPool)
+    private static ConditionGeneration CreateVCGen(Program program, CheckerPool checkerPool, CancellationToken ct)
     {
-      return new VCGen(program, checkerPool);
+      return new VCGen(program, checkerPool, ct);
     }
 
     #region Houdini
--- a/Source/Houdini/Houdini.cs
+++ b/Source/Houdini/Houdini.cs
@@ -5,6 +5,7 @@ using VC;
 using System.IO;
 using Microsoft.Boogie.GraphUtil;
 using System.Linq;
+using System.Threading;
 
 namespace Microsoft.Boogie.Houdini
 {
@@ -408,19 +409,22 @@ namespace Microsoft.Boogie.Houdini
     }
 
     public static TextWriter explainHoudiniDottyFile;
+    private CancellationToken cancellationToken;
 
     protected Houdini()
     {
+      cancellationToken = CancellationToken.None;
     }
 
     public Houdini(Program program, HoudiniSession.HoudiniStatistics stats, string cexTraceFile = "houdiniCexTrace.txt")
     {
       this.program = program;
       this.cexTraceFile = cexTraceFile;
-      Initialize(program, stats);
+      Initialize(program, stats, new CancellationToken());
     }
 
-    protected void Initialize(Program program, HoudiniSession.HoudiniStatistics stats)
+    protected void Initialize(Program program, HoudiniSession.HoudiniStatistics stats,
+      CancellationToken cancellationToken)
     {
       if (CommandLineOptions.Clo.Trace)
       {
@@ -465,9 +469,9 @@ namespace Microsoft.Boogie.Houdini
       */
 
       var checkerPool = new CheckerPool(CommandLineOptions.Clo);
-      this.vcgen = new VCGen(program, checkerPool);
+      this.vcgen = new VCGen(program, checkerPool, cancellationToken);
       this.proverInterface = ProverInterface.CreateProver(program, CommandLineOptions.Clo.ProverLogFilePath,
-        CommandLineOptions.Clo.ProverLogFileAppend, CommandLineOptions.Clo.TimeLimit, taskID: GetTaskID());
+        CommandLineOptions.Clo.ProverLogFileAppend, CommandLineOptions.Clo.TimeLimit, cancellationToken, taskID: GetTaskID());
 
       vcgenFailures = new HashSet<Implementation>();
       Dictionary<Implementation, HoudiniSession> houdiniSessions = new Dictionary<Implementation, HoudiniSession>();
@@ -891,7 +895,7 @@ namespace Microsoft.Boogie.Houdini
         this.NotifyImplementation(currentHoudiniState.Implementation);
 
         houdiniSessions.TryGetValue(currentHoudiniState.Implementation, out var session);
-        ProverInterface.Outcome outcome = TryCatchVerify(session, stage, completedStages, out var errors);
+        ProverInterface.Outcome outcome = TryCatchVerify(session, stage, completedStages, out var errors, cancellationToken);
         UpdateHoudiniOutcome(currentHoudiniState.Outcome, currentHoudiniState.Implementation, outcome, errors);
         this.NotifyOutcome(outcome);
 
@@ -1485,11 +1489,11 @@ namespace Microsoft.Boogie.Houdini
     }
 
     private ProverInterface.Outcome TryCatchVerify(HoudiniSession session, int stage, IEnumerable<int> completedStages,
-      out List<Counterexample> errors)
+      out List<Counterexample> errors, CancellationToken cancellationToken)
     {
       ProverInterface.Outcome outcome;
       try {
-        outcome = session.Verify(proverInterface, GetAssignmentWithStages(stage, completedStages), out errors, GetErrorLimit());
+        outcome = session.Verify(proverInterface, GetAssignmentWithStages(stage, completedStages), out errors, GetErrorLimit(), cancellationToken);
       }
       catch (UnexpectedProverOutputException upo)
       {
@@ -1543,7 +1547,7 @@ namespace Microsoft.Boogie.Houdini
         this.NotifyAssignment(currentHoudiniState.Assignment);
 
         //check the VC with the current assignment
-        ProverInterface.Outcome outcome = TryCatchVerify(session, stage, completedStages, out var errors);
+        ProverInterface.Outcome outcome = TryCatchVerify(session, stage, completedStages, out var errors, cancellationToken);
         this.NotifyOutcome(outcome);
 
         DebugRefutedCandidates(currentHoudiniState.Implementation, errors);
--- a/Source/Provers/SMTLib/SMTLibProcess.cs
+++ b/Source/Provers/SMTLib/SMTLibProcess.cs
@@ -14,6 +14,7 @@ namespace Microsoft.Boogie.SMTLib
     readonly Process prover;
     readonly Inspector inspector;
     readonly SMTLibProverOptions options;
+    private readonly CancellationToken cancellationToken;
     readonly Queue<string> proverOutput = new();
     readonly Queue<string> proverErrors = new();
     readonly TextWriter toProver;
@@ -21,9 +22,10 @@ namespace Microsoft.Boogie.SMTLib
     static int smtProcessIdSeq = 0;
     ConsoleCancelEventHandler cancelEvent;
 
-    public SMTLibProcess(SMTLibOptions libOptions, SMTLibProverOptions options)
+    public SMTLibProcess(SMTLibOptions libOptions, SMTLibProverOptions options, CancellationToken cancellationToken)
     {
       this.options = options;
+      this.cancellationToken = cancellationToken;
       smtProcessId = smtProcessIdSeq++;
 
       var psi = new ProcessStartInfo(options.ExecutablePath(), options.SolverArguments.Concat(" "))
@@ -432,6 +434,7 @@ namespace Microsoft.Boogie.SMTLib
         {
           while (proverOutput.Count == 0 && proverErrors.Count == 0 && !prover.HasExited)
           {
+            cancellationToken.ThrowIfCancellationRequested();
             Monitor.Wait(this, 100);
           }
 
--- a/Source/VCGeneration/Checker.cs
+++ b/Source/VCGeneration/Checker.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics.Contracts;
 using System.Linq;
+using System.Threading;
 using Microsoft.Boogie.VCExprAST;
 using System.Threading.Tasks;
 using VC;
@@ -133,7 +134,7 @@ namespace Microsoft.Boogie
       {
         ctx = (ProverContext) cce.NonNull(ctx).Clone();
         prover = (ProverInterface)
-          CommandLineOptions.Clo.TheProverFactory.SpawnProver(CommandLineOptions.Clo, SolverOptions, ctx);
+          CommandLineOptions.Clo.TheProverFactory.SpawnProver(CommandLineOptions.Clo, SolverOptions, ctx, vcgen.CancellationToken);
       }
       else
       {
@@ -148,7 +149,7 @@ namespace Microsoft.Boogie
         // context in the cache, so that the prover can setup stuff in
         // the context to be cached
         prover = (ProverInterface)
-          CommandLineOptions.Clo.TheProverFactory.SpawnProver(CommandLineOptions.Clo, SolverOptions, ctx);
+          CommandLineOptions.Clo.TheProverFactory.SpawnProver(CommandLineOptions.Clo, SolverOptions, ctx, vcgen.CancellationToken);
         cachedContexts.Add(key, cce.NonNull((ProverContext) ctx.Clone()));
       }
 
@@ -299,13 +300,14 @@ namespace Microsoft.Boogie
     {
       lock (this)
       {
-        try
-        {
+        try {
           outcome = thmProver.CheckOutcome(cce.NonNull(handler), CommandLineOptions.Clo.ErrorLimit);
-        }
-        catch (UnexpectedProverOutputException e)
-        {
+        } catch (UnexpectedProverOutputException e) {
           outputExn = e;
+        } catch (OperationCanceledException) {
+          outcome = ProverInterface.Outcome.Canceled;
+          // We don't want to log the comment as the process has been killed
+          throw;
         }
         catch (Exception e)
         {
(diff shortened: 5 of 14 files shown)


RustanLeino (commented):
This looks good to me, but I'm not an expert on how these cancelations are done, so I'm hoping someone else will take a look and approve the PR.

Source/Core/ProverUtil.cs
```diff
@@ -301,7 +302,8 @@ public abstract class ProverFactory
   {
     // Really returns ProverInterface.
     //public abstract object! SpawnProver(ProverOptions! options, object! ctxt);
-    public abstract object SpawnProver(SMTLibOptions libOptions, ProverOptions options, object ctxt);
+    public abstract object SpawnProver(SMTLibOptions libOptions, ProverOptions options, object ctxt,
+      CancellationToken cancellationToken);
```
cpitclaudel, line 306: Is it a problem to break this API?  Existing implementations of this interface will have to gain an extra argument.
(Alternatively, I don't know what `ctxt` is, but maybe it could contain the cancellationToken)

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -1124,6 +1125,14 @@ public static void Inline(Program program)
       return outcome;
     }
 
+    private static void ThrowIfCanceled(AggregateException ae) {
+      var canceledException =
+        ae.Flatten().InnerExceptions.FirstOrDefault(exception => exception is OperationCanceledException, null);
+      if (canceledException != null) {
+        throw canceledException;
+      }
```
cpitclaudel, line 1133: When I have to do something with side-effects on a collection I find that a plain "foreach" works better than Linq:

```
  foreach (var e in ae.Flatten().InnerExceptions) {
    if (e is OperationCanceledException) {
      throw e;
    }
  }
```

Or, if we're trying to keep things as short as possible:

```
  foreach (var e in ae.Flatten().InnerExceptions.OfType<OperationCanceledException>()) {
    throw e;
  }
```

Source/Houdini/Houdini.cs
```diff
@@ -408,19 +409,22 @@ public HoudiniState CurrentHoudiniState
...
     {
+      cancellationToken = CancellationToken.None;
     }
 
     public Houdini(Program program, HoudiniSession.HoudiniStatistics stats, string cexTraceFile = "houdiniCexTrace.txt")
     {
       this.program = program;
       this.cexTraceFile = cexTraceFile;
-      Initialize(program, stats);
+      Initialize(program, stats, new CancellationToken());
```
cpitclaudel, line 423: How do you decide between `new CancellationToken` and `CancellationToken.None`?

Source/VCGeneration/Checker.cs
```diff
@@ -299,13 +300,14 @@ private void WaitForOutput(object dummy)
...
           outcome = thmProver.CheckOutcome(cce.NonNull(handler), CommandLineOptions.Clo.ErrorLimit);
-        }
-        catch (UnexpectedProverOutputException e)
-        {
+        } catch (UnexpectedProverOutputException e) {
           outputExn = e;
+        } catch (OperationCanceledException) {
+          outcome = ProverInterface.Outcome.Canceled;
+          // We don't want to log the comment as the process has been killed
+          throw;
```
cpitclaudel, line 310: Would the code still work if you remove the `throw`? If not I wouldn't set `outcome`.

Source/VCGeneration/Checker.cs
```diff
@@ -417,7 +423,7 @@ public override void BeginCheck(string descriptiveName, VCExpr vc, ErrorHandler
     }
 
     [NoDefaultContract]
-    public override Outcome CheckOutcome(ErrorHandler handler, int taskID = -1)
+    public override Outcome CheckOutcome(ErrorHandler handler, int errorLimit)
```
cpitclaudel, line 426: ?

Source/VCGeneration/StratifiedVC.cs
```diff
@@ -356,6 +357,7 @@ public class StratifiedInliningInfo
 
     // boolControlVC (block -> its Bool variable)
     public Dictionary<Block, VCExprVar> blockToControlVar;
+    private CancellationToken cancellationToken;
```
cpitclaudel, line 360: The warnings for this line seem relevant

Source/VCGeneration/VCGen.cs
```diff
@@ -17,13 +17,14 @@ namespace VC
 
   public class VCGen : ConditionGeneration
   {
+    public CancellationToken CancellationToken { get; }
```
cpitclaudel, line 20: Is this needed?  It would be inherited, no?

Source/VCGeneration/VCGen.cs
```diff
@@ -351,7 +352,7 @@ bool CheckUnreachable(Block cur, List<Cmd> seq)
             }
 
             ch.BeginCheck(cce.NonNull(impl.Name + "_smoke" + id++), vc, new ErrorHandler(absyIds, this.callback),
-              CommandLineOptions.Clo.SmokeTimeout, CommandLineOptions.Clo.ResourceLimit);
+              CommandLineOptions.Clo.SmokeTimeout, CommandLineOptions.Clo.ResourceLimit, cancellationToken);
```
cpitclaudel, line 355: Why not `this.cancellationToken`?

Source/VCGeneration/VCGen.cs
```diff
@@ -489,13 +490,13 @@ void DFS(Block cur)
 
           if (needToCheck)
           {
-            CheckUnreachable(cur, seq);
+            CheckUnreachable(cur, seq, cancellationToken);
```
cpitclaudel, line 493: Same here, isn't there `this.cancellationToken`?

cpitclaudel (changes requested):
Looks good, I just have a few questions

Source/Houdini/Houdini.cs
MikaelMayer: I should use `CancellationToken.None` here, I discovered it later.

Source/VCGeneration/Checker.cs
MikaelMayer: As the comment says, if we don't throw, it fails because it will call `thmProver.LogComment("canceled");`, which will fail because it will try to send a comment to the Z3 process and the process is not valid anymore since it was killed.

The default value of "outcome" is "valid". If I do not set it to something else, the method `Checker.ReadOutcome` will return `valid`, which will makes its two callers potentially take bad decisions. I don't think these two are in the same thread that is being killed.
Your remark however gave me the indication I also need to update the outcome for the split, as well as the API for the callback.

Source/VCGeneration/Checker.cs
MikaelMayer: In both places where this function is called, and in its parent class original definition, the second argument is named "errorLimit" and not a "taskID", so this seems like is a discrepancy I fixed here.

Source/VCGeneration/VCGen.cs
MikaelMayer: Correct. I forgot to remove it.

Source/VCGeneration/VCGen.cs
MikaelMayer: Actually, the cancellationToken was no longer needed here. I removed it.

Source/VCGeneration/VCGen.cs
MikaelMayer: It was no longer needed anyway, so I removed it.

Source/VCGeneration/Split.cs
```diff
@@ -1340,6 +1340,13 @@ public void ReadOutcome(ref ConditionGeneration.Outcome curOutcome, out bool pro
               curOutcome = ConditionGeneration.Outcome.Inconclusive;
             }
 
+            return;
+          case ProverInterface.Outcome.Canceled:
+            if (curOutcome != ConditionGeneration.Outcome.Canceled)
+            {
+              curOutcome = ConditionGeneration.Outcome.Canceled;
+            }
```
cpitclaudel, line 1348: Why not just set it unconditionally?  Is it a setter?

cpitclaudel (approved):
:+1:

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -1042,6 +1042,7 @@ public static void Inline(Program program)
       }
       catch (AggregateException ae)
       {
+        ThrowIfCanceled(ae);
```
atomb, line 1045: This could probably be handled more cleanly inside the `ae.Flatten().Handle` call below.

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -1251,8 +1262,8 @@ private static void CleanupCheckers(string requestId)
             verificationResult.Errors = null;
             verificationResult.Outcome = VCGen.Outcome.Inconclusive;
           }
-          catch(AggregateException ae)
-          {
+          catch(AggregateException ae) {
+            ThrowIfCanceled(ae);
```
atomb, line 1266: Similar to my previous comment, though perhaps it makes sense to have a separate method since this logic needs to happen in multiple places.

Source/VCGeneration/Split.cs
MikaelMayer: Good catch. I tried to mimick what I saw before, but apparently I did not understand what they were doing. Now I do: Let's assign it unconditionally.

Source/ExecutionEngine/ExecutionEngine.cs
MikaelMayer: I don't think so. Calling code will rely on this OperationCanceledException to be thrown instead of an all-purpose AggregateException
Handle will throw the first exception that wasn't handled. Here we don't care about other exceptions if one is an OeprationCanceledException.
Moreover, the existing code to which you point just "cancels out" the exception "OperationCanceledException" because it returns true there, meaning it pretends it was handled. I'm not sure why someone wrote that though.

atomb: Ah, right, and you want `OperatingCanceledException` to take priority over anything else. (Otherwise you could just change that `true` to a `false`, I think?). That makes sense.

cpitclaudel (dismissed):

cpitclaudel:
(I don't have merge rights but I think this is good to go)

Source/Provers/SMTLib/SMTLibProcess.cs
```diff
@@ -14,16 +14,18 @@ public class SMTLibProcess : SMTLibSolver
...
+    private readonly CancellationToken cancellationToken;
     readonly Queue<string> proverOutput = new();
     readonly Queue<string> proverErrors = new();
     readonly TextWriter toProver;
     readonly int smtProcessId;
     static int smtProcessIdSeq = 0;
     ConsoleCancelEventHandler cancelEvent;
 
-    public SMTLibProcess(SMTLibOptions libOptions, SMTLibProverOptions options)
+    public SMTLibProcess(SMTLibOptions libOptions, SMTLibProverOptions options, CancellationToken cancellationToken)
```
keyboardDrummer, line 25: I'm surprised to see a cancellationToken in a constructor. I'd expect a cancellationToken to be passed to a method that return a `Task`, or otherwise is a blocking method, where the semantics of the token would be "if you cancel it, this asynchronous method will return immediately". For a constructor that already returns immediately, it's less clear what the meaning of the token is.

Have you considered adding the token as an argument to `GetProverResponse`? That way you can also cancel even if the solver process is not stuck, but for some other reason you don't want to wait for the current response to be returned, and then you might re-use that solver later on and pass in a new cancellation token when calling `GetProverResponse`.

Similar, could you pass the CancellationToken down through from ExecutionEngine down through VerifyImplementation towards `SplitAndVerifyWorker.StartCheck`
split.BeginCheck, instead of through constructors?

Source/ExecutionEngine/ExecutionEngine.cs
MikaelMayer: Here are some places that catch this exception, and they don't catch `AggregatedException`:

https://github.com/dafny-lang/dafny/blob/master/Source/DafnyLanguageServer/Workspace/TextDocumentLoader.cs#L98
https://github.com/dafny-lang/dafny/blob/cf0427b87715fd358cbbffed21d51defbeb58b17/Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs#L86
https://github.com/dafny-lang/dafny/blob/cf0427b87715fd358cbbffed21d51defbeb58b17/Source/DafnyLanguageServer/Workspace/DocumentDatabase.cs#L159

Source/Houdini/Houdini.cs
```diff
@@ -408,19 +409,22 @@ public HoudiniState CurrentHoudiniState
...
     public Houdini(Program program, HoudiniSession.HoudiniStatistics stats, string cexTraceFile = "houdiniCexTrace.txt")
     {
       this.program = program;
       this.cexTraceFile = cexTraceFile;
-      Initialize(program, stats);
+      Initialize(program, stats, CancellationToken.None);
     }
 
-    protected void Initialize(Program program, HoudiniSession.HoudiniStatistics stats)
+    protected void Initialize(Program program, HoudiniSession.HoudiniStatistics stats,
```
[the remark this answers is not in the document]
MikaelMayer: Good catch. I was using parameters in many places until I realized I should just put it in the constructor,. Since I have no idea what this "Houdini" class is about, It's a conservative change that someone knowledgeable about it would know what to do with (perhaps they also want to rely on CancellationToken). Hence, it's always defined as CancellationToken.None for now.

Source/Houdini/Houdini.cs
```diff
@@ -408,19 +409,22 @@ public HoudiniState CurrentHoudiniState
     }
 
     public static TextWriter explainHoudiniDottyFile;
+    private CancellationToken cancellationToken;
 
     protected Houdini()
     {
+      cancellationToken = CancellationToken.None;
```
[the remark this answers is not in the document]
MikaelMayer: Right.

Source/Provers/SMTLib/SMTLibProcess.cs
MikaelMayer: I was not aware of this habit to put CancellationTokens only in methods that return Task.

I did consider to put the token as an argument to `GetProverResponse()` but I did not chose that option, because I do want the process to be killed if the task is canceled, not reused. If the `CancellationToken.ThrowIfRequested()` is called, it means that we were in the middle of a computation that took longer to execute than the patience of the user (or, in the case of the Dafny IDE plugin, the user changed the file by 1 character and verification is on every change, which would completely invalidate the verification results).
Since we have no way to know how longer the process needs time to execute, I don't think any library should rely on a process that just got canceled. In the case of Dafny, reusing a canceled process would simply be like starting a new process, except that you have to wait an extra time between 0 and infinite until the previous verification finishes.

To conclude, if the CancellationToken is not in the calling method, it has to be a field, hence why it appears in the constructor, but you already figured that out. Let me know if you have more insights to share that might shake the reasoning above.

keyboardDrummer: I'm fine with keeping the current behavior, but I would like to only pass CancellationTokens as part of asynchronous methods. I think you can do both.

Source/ExecutionEngine/ExecutionEngine.cs
keyboardDrummer: Is it is possible for a Task to throw anything other than an `AggregateException`? ([documentation](https://docs.microsoft.com/en-us/dotnet/standard/parallel-programming/exception-handling-task-parallel-library)) It seems to me like those places are incorrectly handling something other than `AggregateException`. I've placed breakpoints in the handlers and ran the entire test-suite but didn't hit the breakpoints.

Also, sometimes rethrowing an `AggregateException` through a call to `ae.Flatten().Handle` and sometimes throwing a `OperationCanceledException` seems like an API problem to me, since then callers of `InferAndVerify` have to handle both exceptions regularly and again if they are nested inside an `AggregateException`.

MikaelMayer: Wow this is fascinating. It looks like you've uncovered a latent issue there.

> Is it is possible for a Task to throw anything other than an AggregateException? I think it's not.

I think it is, quote from https://docs.microsoft.com/en-us/archive/msdn-magazine/2013/march/async-await-best-practices-in-asynchronous-programming:
>Every Task will store a list of exceptions. When you await a Task, the first exception is re-thrown, so you can catch the specific exception type (such as InvalidOperationException). However, when you synchronously block on a Task using Task.Wait or Task.Result, all of the exceptions are wrapped in an AggregateException and thrown. Refer again to Figure 4. The try/catch in MainAsync will catch a specific exception type, but if you put the try/catch in Main, then it will always catch an AggregateException. Error handling is much easier to deal with when you don’t have an AggregateException, so I put the “global” try/catch in MainAsync.

keyboardDrummer: Ah, thanks for the info :-) 

Do you have a comment on this?
> sometimes rethrowing an AggregateException through a call to ae.Flatten().Handle and sometimes throwing a OperationCanceledException seems like an API problem to me, since then callers of InferAndVerify have to handle each exception type both as a type level exception and as one nested inside an AggregateException. I think the API for any method should be to only throw AggregateException or to never throw it.

keyboardDrummer: How do you feel about the second paragraph in my previous comment?

MikaelMayer: I wanted to make sure I found relevant information to you before commenting on your second paragraph, because I was unsure about the first one.
>Also, sometimes rethrowing an AggregateException through a call to ae.Flatten().Handle and sometimes throwing a OperationCanceledException seems like an API problem to me, since then callers of InferAndVerify have to handle each exception type both as a type level exception and as one nested inside an `AggregateException`.

I agree on that, it would be better to have only one type of exception for an API.
> I think the API for any method should be to only throw `AggregateException` or to never throw it.

I think I agree for now only on the cases when we do rethrow an `AggregateException` like above. Otherwise, calling threads should be able to throw the exceptions they want after receiving an `AggregateException`, they will still be gathered into an `AggregateException` depending on how they had been called.
I changed the code to remove entirely `ThrowIfCanceled` and reimplemented the same flow by just removing the line doing the test on `OperationCanceledException` as @atomb  suggested.
The Dafny test pass, however when debugging, it still pauses execution and displays the `AggregateException` as a "user-unhandled exception", which it did not do with the custom `ThrowIfCanceled`. I'm not sure whether it's a smell or it's just an IDE opinionated about when to stop the debugger or not on exception.

Source/Provers/SMTLib/SMTLibProcess.cs
MikaelMayer: > I would like to only pass CancellationTokens as part of asynchronous or blocking methods to make the meaning of the tokens clear.
A constructor is a blocking method; can you please let me know precisely what you mean?

keyboardDrummer: With blocking method I meant a method that does some time consuming operation such as IO and only returns after it's done with that operation. An example is `SMTLibProcess.ReadProver` since it calls `Monitor.Wait(this, 100)`. I think the idiomatic signature of this method would be `Task<string> ReadProver(CancellationToken cancellationToken)` which I think would also allow it to return immediately when `OutputDataReceived` fires without having to use `Monitor.Wait`. Changing the return type to `Task<string`> is a hassle but we can already add the CancellationToken argument in this PR.

MikaelMayer: I agree it's idiomatic, but look at the changes it triggers. in the case of the prover (the last commit I just pushed)
https://github.com/boogie-org/boogie/pull/491/commits/528025538f0258be8cff7bbf6116b9ad15df51af
I would then add that, when the same unchanged argument is given over and over to several methods of the same class, it is also idiomatic to put it in a constructor so that the signatures of methods stays as simple as possible.

keyboardDrummer: The changes in that commit look good to me. I think with a different design a .NET `Task` might have been equivalent to `Func<CancellationToken, WorkTask>`, similar to a [State monad](https://wiki.haskell.org/State_Monad), and then you wouldn't need to propagate cancellationTokens, but with the current design, lots of propagation of cancellationTokens is the way it is.

But em.. thinking more about this. I'm confused you even need to do `cancellationToken.ThrowIfCancellationRequested();` there. Shouldn't the `!prover.HasExited` in `while (proverOutput.Count == 0 && proverErrors.Count == 0 && !prover.HasExited)` already allow it to return when the process gets killed, which should happen when `ExecutionEngine.CancelRequest` is called. What's going on?

MikaelMayer: That's a brilliant finding. HIstorically, I first threw the error there so that i would stop the loop, and later I added `cancellationToken.Register(KillProcess);` in the constructor of `SMTLibProcessTheoremProver` to fix another error.
But it looks like when I comment `cancellationToken.ThrowIfCancellationRequested()` because `cancellationToken.Register(KillProcess);` does the work.

I pushed a commit that removes the cancellationTokens inside the process alltogether

keyboardDrummer: But, why do we need the changes in this PR at all, since `SMTLibProcess.Close` should already be called as a result of `ExecutionEngine.CancelRequest` being called?

Also, I don't think we should merge this PR without it having a test.

MikaelMayer: After a lot of further investigation, here is what happens:

`ExecutionEngine.CancelRequest` calls `CleanupCheckers(requestId)`
`CleanupCheckers(requestId)` calls `checkerPool?.Dispose();` **only if there is no other pending request**
`checkerPool.Dispose();` calls `checker.Close();` **only for checkers that are available (idle)**

Here we see why it cannot work without `cancellationToken.Register(KillProcess);`. First, if there are other pending requests, the `checkerPool` is not disposed. Second, disposing the `checkerPool` only calls `Close` on the checkers that were available. However, one of the two checkers that were created in my test did not finish, so it was never registered to be available and does not close.

This PR has a test but it is in Dafny:
https://github.com/dafny-lang/dafny/pull/1771
I can convert it to Boogie, but can you please let me know how you would proceed for that? There is no asynchronous testing anywhere, so I'm not sure where to start.

MikaelMayer: The use of `Task.WaitASync(CancellationToken)` is very interesting, one, thanks for your suggestion, and let me have a deeper look at it.
I also tested your branch, and right now the Dafny Language Server test succeeds, but as expected the process that started the test does not finish because Z3 is not killed. See the attached screenshot.
image

I suggest we do both what you are doing and what I am doing, so that we can end even faster for the client (with your change, no need to wait the extra 2000ms if the prover is unresponsive)

>You might say that starting Z3 processes is cheap so we don't need to recycle them, but then why do we have a CheckerPool class in the first place?

I do not have a definitive answer, but I think this is an excellent question to ask now.
Here are some insights I gathered while looking at CheckerPool.cs and Checker.cs just now:
* CheckerPool ensures that there are at most <number of VCSCores> checkers in the pool. This is very clever, because we don't want more than that processes running at the same time.
  However, CheckerPool **does not keep a pointer on checkers**, only an abstract representation `int notCreatedCheckers ` of how many checkers it can still create.
* When something needs a checker, `CheckerPool.FindCheckerFor` will return a promise for a checker (a Task<Checker>). This promise is fullfilled by either
  * Detecting if an existing checker is idle, meaning there is a checker in `availableCheckers`, meaning something called the checker's `GoBackToIdle()` method (in two places, one being the end of `SplitAndVerifyWorker.DoWork` in the `finally` you refactored when it calls `split.ReleaseChecker();`) which normally ensures that the z3 process answered and can now listen to standard input
  * If the number of maximum checkers has not been reached, creating a new checker and returning it immediately as a `Task.FromResult(checker)`
  * Else, it **assumes that a checker will eventually get idle**, and enqueue the request for a checker in `checkerWaiters`, and assigns it the continuation to `PrepareChecker` on the obtained checker when it will be available, and returning the checker itself. This request is fullfilled in `Checker.GoBackToIdle`, which in turns call `CheckerPool.AddChecker`, which, if there is a waiter in `checkerWaiters`, immediately provide the checker, and if nothing is waiting for a checker, it will enqueue the idle checker in `availableCheckers`

So here is a updated comment now with this updated information:
When a cancellation is requested, there is no way to query the process to say "Hey, please cancel this request" nicely. Moreover, it's even possible that the process will never stop and that the checker will never get back to idle, and the CheckerPool won't be aware of it, it just knows that it can still create `notCreatedCheckers` more checkers for new requests.
Which means, after <number of cores> requests that create failing checkers, the checkerpool won't create any new checkers and that's how even a request for a simple file fails afterwards, because of all these blocking checkers.

Now that I understand this, I think I nailed down a way to make things even more consistent
- When a checker's cancellation token is triggered, it should just kill the process, period. Maybe give it 2 seconds to recover resources, using Process.Close like it is currently.
- A naïve idea would be that, when the checker's cancellation token is triggered (meaning the checker won't be recovered, and the process is closed), it should also somehow increment `CheckerPool.notCreatedCheckers` so that the CheckerPool is no longer limited. However, this can lead to a deadlock, because if `CheckerPool.notCreatedCheckers` was at 0 right before, it means that any request since the last time it was set to 0 is actually waiting for a checker, and they won't ever recover.
- The robust idea is that, if there is a `checkerWaiter`, it will be immediately fed with a fresh Checker and we don't touch `CheckerPool.notCreatedCheckers`. In the contrary, we increase `CheckerPool.notCreatedCheckers`

RustanLeino: > > You might say that starting Z3 processes is cheap so we don't need to recycle them, but then why do we have a CheckerPool class in the first place?
> 
> I do not have a definitive answer, but I think this is an excellent question to ask now.

Historically, I think the `CheckerPool` was added for two reasons. One reason was to limit the maximum number of checkers in use. The other reason was to reuse checkers, rather than starting a new Z3 process, because starting a new Z3 process took something like 0.1-0.3 seconds, which was noticeable for files with lots of tiny methods.

Even if we spin up new Z3 processes on demand now, it may be worthwhile to start up some Z3 processes (and put them into the `CheckerPool`) before they are needed. This may reduce the latency in getting a new Z3 process.

shazqadeer:
Abandon this PR?

PR #495, opened 2022-02-04
Allow continuing if rlimit response is invalid

atomb:
Previously, the code to get the resource count from Z3 could crash if Z3's response was badly-formed. Now it will continue with a dummy value of -1.

### DIFF
--- a/Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
+++ b/Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
@@ -1951,7 +1951,20 @@ namespace Microsoft.Boogie.SMTLib
     {
       SendThisVC("(get-info :rlimit)");
       var resp = Process.GetProverResponse();
-      return int.Parse(resp[0].Name);
+      try
+      {
+        return int.Parse(resp[0].Name);
+      }
+      catch
+      {
+        // If anything goes wrong with parsing the response from the solver,
+        // it's better to be able to continue, even with uninformative data.
+        lock (proverWarnings)
+        {
+          proverWarnings.Add("Failed to parse resource count from solver.");
+        }
+        return -1;
+      }
     }
 
     object ParseValueFromProver(SExpr expr)
--- a/Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
+++ b/Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
@@ -1961,7 +1961,7 @@ namespace Microsoft.Boogie.SMTLib
         // it's better to be able to continue, even with uninformative data.
         lock (proverWarnings)
         {
-          proverWarnings.Add("Failed to parse resource count from solver.");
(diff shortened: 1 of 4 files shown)


Test/prover/mocksolver.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+while true ; do
+    read cmd
+    case $cmd in
+      *get-info*name*   ) echo "(:name \"Z3\")" ;;
+      *get-info*rlimit* ) echo "(:rlimit bad)" ;;
```
robin-aws, line 6: Cute way to reproduce the bug, nice :) 

I'm assuming the real-world behaviour is a random error message of some kind? Would it be smarter for the system to check for that earlier? I think making `GetRCount` more robust is smart anyway, but are we swallowing a potentially useful error message?

Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
```diff
@@ -1951,7 +1951,20 @@ public override int GetRCount()
...
+      {
+        return int.Parse(resp[0].Name);
+      }
+      catch
+      {
+        // If anything goes wrong with parsing the response from the solver,
+        // it's better to be able to continue, even with uninformative data.
+        lock (proverWarnings)
+        {
+          proverWarnings.Add("Failed to parse resource count from solver.");
```
robin-aws, line 1964: As per below, let's include the response in the warning

robin-aws (changes requested):

Test/prover/mocksolver.sh
atomb: Yes, I expect so. I haven't been able to reproduce the situation well enough to know what sort of response it's actually getting from the solver. It may be possible to detect it earlier once we know what it is. And your next comment gets at that. :)

Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
atomb: Agreed.

Test/prover/rlimit-parse.bpl
```diff
@@ -0,0 +1,7 @@
+// RUN: %boogie /trace /proverOpt:PROVER_PATH=mocksolver.sh "%s" > "%t"
+// RUN: %OutputCheck --file-to-check "%t" "%s"
+// CHECK: resource count: -1
```
robin-aws, line 3: Non-blocking nice to have: can we CHECK for the prover warning? Does that appear in stdout?

robin-aws (approved):

Test/prover/rlimit-parse.bpl
atomb: It's supposed to get printed if you supply `/proverWarnings:1`, but that doesn't seem to be happening. I think that may be a deeper issue, now that I look at it, but perhaps one to investigate later.

PR #496, opened 2022-02-08
feat: Adding optional trivia to tokens.

MikaelMayer:
This PR adds optional trivia to tokens. Since Boogie is used as a back-end for other programming languages which might need tree-to-tree transformation (such as Dafny), it is handy to have a way to store the leading and trailing trivia in each token.

Just added tests for a functionality that makes it possible to "borrow trivia" from one token to another token.


Source/Core/Token.cs
```diff
@@ -83,7 +88,48 @@ public int line
...
+    {
+      get { return this._leadingTrivia; }
+      set { this._leadingTrivia = value; }
+    }
+    
+    public string /*!*/ trailingTrivia
+    {
+      get { return this._trailingTrivia; }
+      set { this._trailingTrivia = value; }
+    }
```
cpitclaudel, line 102: Why do you have these backing fields instead of using an auto-implemented property?

MikaelMayer: just to be consistent with the previous implementation.
If it was just me, I'd had public fields.

MikaelMayer:
I don't need this anymore.

PR #497, opened 2022-02-10
fix: Cancellation tokens to cancel and restart checkers

keyboardDrummer:
This PR enables Boogie's execution engine to effectively kill verification on demand, and recycle killed checkers/SMTLib processes

## Changes

- Enable cancellation while waiting for input from the solver
- Only recycle a `Checker` if its associated prover is responsive
- Return a `PipelineOutcome.Cancelled` outcome if a request was cancelled.

## Issues it fixes
This PR should be able to fix the following bugs:
https://github.com/dafny-lang/ide-vscode/issues/119
https://github.com/dafny-lang/ide-vscode/issues/114
https://github.com/dafny-lang/ide-vscode/issues/96

## How it has been tested

1. Add a unit test `InferAndVerifyCanBeCancelledWhileWaitingForProver`

2. Dafny tests
https://github.com/dafny-lang/dafny/pull/1771
It takes a program which is hard to verify, and two changes that makes it easy to verify and then a parse error.
Without this PR, the test fails and loops indefinitely on the first program.
With this PR, the tests succeeds and all the changes are taken into account.

### DIFF
--- a/Source/Provers/SMTLib/SMTLibProcess.cs
+++ b/Source/Provers/SMTLib/SMTLibProcess.cs
@@ -4,7 +4,6 @@ using System.Linq;
 using System.Text;
 using System.Diagnostics;
 using System.IO;
-using System.Threading;
 using System.Diagnostics.Contracts;
 using System.Threading.Tasks;
 
@@ -58,6 +57,7 @@ namespace Microsoft.Boogie.SMTLib
         prover.StartInfo = psi;
         prover.ErrorDataReceived += prover_ErrorDataReceived;
         prover.OutputDataReceived += prover_OutputDataReceived;
+        prover.Exited += prover_Exited;
         prover.Start();
         toProver = prover.StandardInput;
         prover.BeginErrorReadLine();
@@ -67,6 +67,17 @@ namespace Microsoft.Boogie.SMTLib
       {
         throw new ProverException(string.Format("Unable to start the process {0}: {1}", psi.FileName, e.Message));
       }
+    }
+
+    private void prover_Exited(object sender, EventArgs e)
+    {
+      lock (this) {
+        if (outputReceivers.TryDequeue(out var source)) {
+          source.SetResult(null);
+        }
+      }
+
+      DisposeProver();
     }
 
     [NoDefaultContract] // important, since we have no idea what state the object might be in when this handler is invoked
@@ -114,71 +125,61 @@ namespace Microsoft.Boogie.SMTLib
       toProver.WriteLine(cmd);
     }
 
-    internal Inspector Inspector
+    internal Inspector Inspector => inspector;
+
+    public override async Task<SExpr> GetProverResponse()
     {
-      get { return inspector; }
-    }
+      await toProver.FlushAsync();
 
-    public override Task<SExpr> GetProverResponse(CancellationToken cancellationToken)
-    {
-      return Task.Factory.StartNew(GetProverResponseSync, cancellationToken);
-    }
+      while (true) {
+        var exprs = await ParseSExprs(true).ToListAsync();
+        Contract.Assert(exprs.Count <= 1);
+        if (exprs.Count == 0) {
+          return null;
+        }
 
-    private SExpr GetProverResponseSync()
-    {
-      toProver.Flush();
-
-      lock (this) {
-        while (true) {
-          var exprs = ParseSExprs(true).ToArray();
-          Contract.Assert(exprs.Length <= 1);
-          if (exprs.Length == 0) {
-            return null;
-          }
-
-          var resp = exprs[0];
-          if (resp.Name == "error") {
-            if (resp.Arguments.Length == 1 && resp.Arguments[0].IsId) {
-              if (resp.Arguments[0].Name.Contains("max. resource limit exceeded")) {
-                return resp;
-              } else {
-                HandleError(resp.Arguments[0].Name);
-                return null;
-              }
+        var resp = exprs[0];
+        if (resp.Name == "error") {
+          if (resp.Arguments.Length == 1 && resp.Arguments[0].IsId) {
+            if (resp.Arguments[0].Name.Contains("max. resource limit exceeded")) {
+              return resp;
             } else {
-              HandleError(resp.ToString());
+              HandleError(resp.Arguments[0].Name);
               return null;
             }
-          } else if (resp.Name == "progress") {
-            if (inspector != null) {
-              var sb = new StringBuilder();
-              foreach (var a in resp.Arguments) {
-                if (a.Name == "labels") {
-                  sb.Append("STATS LABELS");
-                  foreach (var x in a.Arguments) {
-                    sb.Append(" ").Append(x.Name);
-                  }
-                } else if (a.Name.StartsWith(":")) {
-                  sb.Append("STATS NAMED_VALUES ").Append(a.Name);
-                  foreach (var x in a.Arguments) {
-                    sb.Append(" ").Append(x.Name);
-                  }
-                } else {
-                  continue;
-                }
-
-                inspector.StatsLine(sb.ToString());
-                sb.Clear();
-              }
-            }
-          } else if (resp.Name == "unsupported") {
-            // Skip -- this may be a benign "unsupported" from a previous command.
-            // Of course, this is suboptimal.  We should really be using
-            // print-success to identify the errant command and determine whether
-            // the response is benign.
           } else {
-            return resp;
+            HandleError(resp.ToString());
+            return null;
           }
+        } else if (resp.Name == "progress") {
+          if (inspector != null) {
+            var sb = new StringBuilder();
+            foreach (var a in resp.Arguments) {
+              if (a.Name == "labels") {
+                sb.Append("STATS LABELS");
+                foreach (var x in a.Arguments) {
+                  sb.Append(" ").Append(x.Name);
+                }
+              } else if (a.Name.StartsWith(":")) {
+                sb.Append("STATS NAMED_VALUES ").Append(a.Name);
+                foreach (var x in a.Arguments) {
+                  sb.Append(" ").Append(x.Name);
+                }
+              } else {
+                continue;
+              }
+
+              inspector.StatsLine(sb.ToString());
+              sb.Clear();
+            }
+          }
+        } else if (resp.Name == "unsupported") {
+          // Skip -- this may be a benign "unsupported" from a previous command.
+          // Of course, this is suboptimal.  We should really be using
+          // print-success to identify the errant command and determine whether
+          // the response is benign.
+        } else {
+          return resp;
         }
       }
     }
@@ -212,7 +213,6 @@ namespace Microsoft.Boogie.SMTLib
     }
 
     public override event Action<string> ErrorHandler;
-    int errorCnt;
 
     protected override void HandleError(string msg)
     {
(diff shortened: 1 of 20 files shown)


Source/VCGeneration/CheckerPool.cs
```diff
@@ -112,5 +112,10 @@ public void AddChecker(Checker checker)
         availableCheckers.Push(checker);
       }
     }
+
+    public void CheckerDied()
```
keyboardDrummer, line 116: If someone is waiting for a checker, they will never stop waiting.

Source/UnitTests/ExecutionEngineTests/CancellationTests.cs
```diff
@@ -0,0 +1,109 @@
...
+      Assert.AreEqual(PipelineOutcome.VerificationCompleted, outcome2);
+    }
+
+    private string fast = @"
+procedure easy() ensures 1 + 1 == 0; {
+}
+";
+
+    string slow = @"
+  type LayerType;
```
keyboardDrummer, line 54: I think we can still remove the canCall logic here.

cpitclaudel (commented):
I don't have enough context to do a sufficient review, but this looks very nice.  Can you say why you can ignore the CancelledException early on?

Source/Provers/SMTLib/SMTLibProcess.cs
```diff
@@ -68,6 +69,17 @@ public SMTLibProcess(SMTLibOptions libOptions, SMTLibProverOptions options)
       }
     }
 
+    private void prover_Exited(object sender, EventArgs e)
+    {
+      lock (this) {
+        if (outputReceivers.TryDequeue(out var source)) {
+          source.SetResult(null);
+        }
```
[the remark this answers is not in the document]
keyboardDrummer: There's a while so it should remove all of them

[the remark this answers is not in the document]
MikaelMayer: There cannot be two things waiting for it in our current code be because we always prefix the calls to `getProverResponse` with `await`. However, as a safeguard for future problems that are unlikely to happend, we added the comment that this class is not thread-safe, meaning it should be also used only with one thread (which is what happens, it is owned by a Checker that runs the prover on only one thread.)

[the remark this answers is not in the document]
cpitclaudel: > I guess you reviewed an older version with just an if :)

!!  Precisely right :)

Source/VCGeneration/ConditionGeneration.cs
```diff
@@ -39,8 +39,6 @@ public ConditionGenerationContracts(Program p, CheckerPool checkerPool)
   [ContractClass(typeof(ConditionGenerationContracts))]
   public abstract class ConditionGeneration : IDisposable
   {
-    protected internal object CheckerCommonState;
-
```
atomb, line 43: 🎉

Source/VCGeneration/ProverInterface.cs
```diff
@@ -156,12 +158,7 @@ public virtual Absy Label2Absy(string label)
   public abstract void BeginCheck(string descriptiveName, VCExpr vc, ErrorHandler handler);
 
   [NoDefaultContract]
-  public abstract Outcome CheckOutcome(ErrorHandler handler, int errorLimit);
-
-  public virtual string[] CalculatePath(int controlFlowConstant)
-  {
-    throw new System.NotImplementedException();
-  }
```
atomb, line 164: Oh, nice! This doesn't need to be here?

Source/VCGeneration/SplitAndVerifyWorker.cs
```diff
@@ -94,22 +94,22 @@ private void TrackSplitsCost(List<Split> splits)
...
+        cancellationToken.ThrowIfCancellationRequested();
+        StartCheck(split, checker, cancellationToken);
         await split.ProverTask;
-        await ProcessResult(split);
+        await ProcessResult(split, cancellationToken);
       }
-      catch {
+      finally {
         split.ReleaseChecker();
-        throw;
```
atomb, line 108: Have you tested this against a badly-behaved prover? I'm worried that this could undo the fix from #488.

atomb (changes requested):
I think this looks great, and seems to simplify code and add functionality at the same time. I'd just like to confirm that issue #487 (fixed in PR #488) doesn't reappear.

Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
```diff
@@ -909,8 +922,21 @@ public override Outcome CheckOutcomeCore(ErrorHandler handler, int errorLimit)
...
-    private Outcome RunTimeoutDiagnostics(ErrorHandler handler, Outcome result, ref bool popLater)
+    T WrapInPushPop<T>(ref bool popLater, Func<T> action)
+    {
+      if (popLater)
+      {
+        SendThisVC("(pop 1)");
+      }
+      SendThisVC("(push 1)");
+      var result = action();
+      popLater = true;
```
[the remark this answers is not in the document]
keyboardDrummer: Yes

Source/VCGeneration/Checker.cs
```diff
@@ -290,56 +323,56 @@ public TimeSpan ProverRunTime
...
+      }
+      catch (Exception e)
+      {
+        outputExn = new UnexpectedProverOutputException(e.Message);
+      }
 
-        hasOutput = true;
-        proverRunTime = DateTime.UtcNow - proverStart;
+      switch (outcome)
+      {
```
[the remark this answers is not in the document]
keyboardDrummer: I can't find the code on which you made this comment, so I don't know how to reply.

Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
```diff
@@ -71,7 +68,17 @@ void ObjectInvariant()
...
+    private static ScopedNamer GetNamer(SMTLibOptions libOptions, ProverOptions options)
+    {
+      return (RandomSeed: options.RandomSeed, libOptions.NormalizeNames) switch
+      {
+        (null, true) => new NormalizeNamer(),
+        (null, false) => new KeepOriginalNamer(),
+        _ => new RandomiseNamer(new Random(options.RandomSeed.Value))
+      };
+    }
+
```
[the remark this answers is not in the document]
keyboardDrummer: Some change was needed for this PR and I think slightly more but not needed changes followed from that. Sorry.

The needed namer related change was `Namer.Reset();` => `HasReset = true;`

Source/VCGeneration/ProverInterface.cs
keyboardDrummer: Not for `https://github.com/boogie-org/boogie`, but possibly for a consumer of `https://github.com/boogie-org/boogie`, although I don't see why you would want to override this. For APIs that are [unexpectedly] used outside of Boogie, we should document the use-case and mark the API with an annotation, like `[PublicAPI]`, otherwise it'll be impossible to make changes to Boogie without fear of breaking a consumer.

Source/VCGeneration/SplitAndVerifyWorker.cs
keyboardDrummer: Since we don't catch the exception, we also don't have to rethrow it since it's thrown up automatically. Where does your concern come from?

atomb: Ah, right, the change from `catch` to `finally` should avoid it. And there is a test that would be failing if this change did undo that fix. 👍

atomb (approved):

Source/VCGeneration/Checker.cs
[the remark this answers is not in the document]
MikaelMayer: I think it's the switch that converts the outcome to a string..

Source/Provers/SMTLib/SMTLibProcess.cs
```diff
@@ -113,85 +128,60 @@ public override void Send(string cmd)
...
-        if (exprs.Length == 0)
-        {
+      while (true) {
+        var exprs = await ParseSExprs(true).ToListAsync();
+        Contract.Assert(exprs.Count <= 1);
+        if (exprs.Count == 0) {
           return null;
         }
 
         var resp = exprs[0];
```
cpitclaudel, line 144: The existing code ignores possible malformed output from the underlying prover, right?  I guess it's OK since it was already like that.

Source/Provers/SMTLib/SMTLibProcess.cs
```diff
@@ -113,85 +128,60 @@ public override void Send(string cmd)
...
                 sb.Append("STATS NAMED_VALUES ").Append(a.Name);
-                foreach (var x in a.Arguments)
-                {
+                foreach (var x in a.Arguments) {
                   sb.Append(" ").Append(x.Name);
                 }
-              }
-              else
-              {
+              } else {
```
cpitclaudel, line 171: I would avoid such refactorings in general, since they pollute the Git history.  Incidentally, this whole method  would probably look a lot better with pattern matching.

Source/Provers/SMTLib/SMTLibProcess.cs
```diff
@@ -416,42 +405,18 @@ IEnumerable<SExpr> ParseSExprs(bool top)
...
     {
-      string error = null;
-      while (true)
-      {
-        if (error != null)
-        {
-          HandleError(error);
-          errorCnt++;
-          error = null;
+      lock (this) {
```
cpitclaudel, line 412: Do ou want external clients to be able take this lock?  If not I'd use a private object.

cpitclaudel: Oh but I see below that the code was already like this, sorry

Source/Provers/SMTLib/SMTLibProcess.cs
```diff
@@ -466,35 +431,39 @@ void DisposeProver()
...
-      }
     }
 
     void prover_ErrorDataReceived(object sender, DataReceivedEventArgs e)
     {
-      lock (this)
-      {
-        if (e.Data != null)
+      lock (this) {
+        if (e.Data == null)
```
cpitclaudel, line 456: In the method above you moved the e.data test out of the lock, but not here.  Could you do the same here?  It would avoid taking the lock when `e.Data` is `null`

Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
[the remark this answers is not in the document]
cpitclaudel: I'm not a fan of this API and its `out popLater` parameter, but I'm fine with sticking with what the code used to do.

Source/VCGeneration/Checker.cs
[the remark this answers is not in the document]
cpitclaudel: The code is the switch on `outcome` in WaitForOutput.  But since all that this PR does to it is reindent it, it's probably not worth changing it.

cpitclaudel (approved):
LGTM, none of my comments are blocking

Source/Provers/SMTLib/SMTLibProcess.cs
keyboardDrummer: Agreed, I'm not sure how this happened. The IDE reformatted this method and I accidentally committed it.

Source/Provers/SMTLib/SMTLibProcess.cs
keyboardDrummer: Ah, good point. Never thought of that. Is it worth the extra code though? External clients taking the lock seems like only a theoretical issue.

PR #501, opened 2022-02-14
Move the implementation of the CommandLineOptions class from Core to ExecutionEngine

keyboardDrummer:
### Changes

- Move the implementation of CommandLineOptions class from `Core` to `ExecutionEngine`. This enables having project specific options be defined inside that project instead of in `Core`.
- Move SMTLibOptions from `Core` to `SMTLib`
- Introduce `VCGenOptions`, `HoudiniOptions` and `ConcurrencyOptions`, host them in their respective projects and use their values through a local `options` object instead of through `CommandLineOptions.Cli`.
- Make VCGeneration depend on SMTLib instead of the other way around, which wasn't possible before, so `SMTLibOptions`, `ProverFactory` and `ProverInterface` can live in SMTLib and methods such as `SpawnProver` and `NewProverContext` can return proper types instead of `object`.

### Breaking changes
Consumers who were using `CommandLineOptions.Clo` must now instead use their own static variable to store the `ExecutionEngineOptions` in. In Dafny for example, references to `CommandLineOptions.Clo` should be replaced with references to `DafnyOptions.Clo`. Of course, these customers may also stop referring to options in static variables but that requires more changes.

### Testing
The changes are covered by automated tests


Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -1925,7 +1758,7 @@ private static ErrorInformation CreateErrorInformation(Counterexample error, VC.
       return errorInfo;
     }
 
-    private static void WriteErrorInformationToXmlSink(ErrorInformation errorInfo, List<Block> trace)
+    private static void WriteErrorInformationToXmlSink(XmlSink sink, ErrorInformation errorInfo, List<Block> trace)
```
atomb, line 1761: I like that this is now a parameter, too!

Source/Provers/SMTLib/SMTLib.csproj
```diff
@@ -12,7 +12,6 @@
     <ProjectReference Include="..\..\Graph\Graph.csproj" />
     <ProjectReference Include="..\..\Model\Model.csproj" />
     <ProjectReference Include="..\..\VCExpr\VCExpr.csproj" />
-    <ProjectReference Include="..\..\VCGeneration\VCGeneration.csproj" />
```
atomb, line 15: Nice reduction in coupling.

Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
```diff
@@ -2488,7 +2488,7 @@ public DatatypeTypeCtorDecl LookupDatatype(string name)
 
   public class Factory : ProverFactory
   {
-    public override object SpawnProver(SMTLibOptions libOptions, ProverOptions options, object ctxt)
+    public override ProverInterface SpawnProver(SMTLibOptions libOptions, ProverOptions options, object ctxt)
```
atomb, line 2491: Much better!

Source/UnitTests/TestUtil/BoogieTestBase.cs
```diff
@@ -10,7 +10,9 @@ public BoogieTestBase()
 
       // FIXME: THIS IS A HACK. Boogie's methods
       // depend on its command line parser being set!
-      CommandLineOptions.Install(new Microsoft.Boogie.CommandLineOptions());
+
+      // TODO FIX
+      //CommandLineOptions.Install(new Microsoft.Boogie.CommandLineOptions());
```
atomb, line 15: Is there work still needed here? I imagine that, since this is in the tests, they'd be breaking if this was still necessary.

Source/ExecutionEngine/CommandLineOptions.cs
```diff
@@ -201,6 +213,20 @@ public bool CheckBooleanFlag(string flagName, ref bool flag)
         return CheckBooleanFlag(flagName, ref flag, true);
       }
 
+      /// <summary>
+      /// If there is one argument and it is a non-negative integer, then set "arg" to that number and return "true".
+      /// Otherwise, emit error message, leave "arg" unchanged, and return "false".
```
atomb, line 218: This comment doesn't seem to match the functionality.

Source/ExecutionEngine/CommandLineOptions.cs
```diff
@@ -214,7 +240,46 @@ public bool GetNumericArgument(ref bool arg)
         }
         return result;
       }
-      
+
+      /// <summary>
+      /// If there is one argument and it is a non-negative integer, then set "arg" to that number and return "true".
+      /// Otherwise, emit error message, leave "arg" unchanged, and return "false".
```
atomb, line 246: This comment also seems mismatched with the functionality.

Source/ExecutionEngine/CommandLineOptions.cs
```diff
@@ -293,6 +389,29 @@ public bool GetNumericArgument(ref int arg, Predicate<int> filter)
         return false;
       }
 
+      /// <summary>
+      /// If there is one argument and it is a non-negative integer less than "limit",
+      /// then set "arg" to that number and return "true".
+      /// Otherwise, emit error message, leave "arg" unchanged, and return "false".
```
atomb, line 395: Ditto.

atomb (changes requested):
This is a nice improvement! There are just a few comments on methods that seem not to match their actual functionality.

Source/UnitTests/TestUtil/BoogieTestBase.cs
keyboardDrummer: Good point. Will remove this class.

atomb (changes requested):
There's one tiny comment inconsistency left. Otherwise it looks great.

atomb (approved):
Looks good!

Source/ExecutionEngine/CommandLineOptions.cs
```diff
@@ -396,11 +516,11 @@ public void Error(string message, params string[] args)
...
 
-  ---- General options -------------------------------------------------------
+    ---- General options -------------------------------------------------------
 
-  /version      print the " + ToolName + @" version number
-  /help         print this message
-  /attrHelp     print a message about supported declaration attributes";
+    /version      print the " + ToolName + @" version number
+    /help         print this message
+    /attrHelp     print a message about supported declaration attributes";
```
cpitclaudel, line 523: Was this reindentation on purpose?

keyboardDrummer: It wasn't. Will correct.

PR #502, opened 2022-02-15
Stop taking lock before its needed

keyboardDrummer:


Source/Provers/SMTLib/SMTLibProcess.cs
```diff
@@ -452,11 +452,12 @@ void prover_OutputDataReceived(object sender, DataReceivedEventArgs e)
 
     void prover_ErrorDataReceived(object sender, DataReceivedEventArgs e)
     {
+      if (e.Data == null)
```
MikaelMayer, line 455: I'm actually surprised that the Data could be null, but better safe than sorry !

I don't think the previous version would have caused trouble, perhaps a few microseconds of delay due to the locking and unlocking, but I think it's always better to improve the code.

MikaelMayer (approved):

cpitclaudel (approved):
Neat, thanks!

PR #504, opened 2022-02-17
Fixed an assertion failed

MikaelMayer:
I suddenly get tons of exceptions in my IDE because the release checker is done many times, and the second time it crashes.
This PR fixes it so that we avoid alltogether this kind of error. Furthermore, it enables throwing the correct exception if one was thrown.

### DIFF
--- a/Source/VCGeneration/Split.cs
+++ b/Source/VCGeneration/Split.cs
@@ -1481,8 +1481,10 @@ namespace VC
 
       public void ReleaseChecker()
       {
-        Checker.GoBackToIdle();
-        checker = null;
+        if (checker != null) {
+          Checker.GoBackToIdle();
+          checker = null;
+        }
       }
     }
 }


Source/VCGeneration/Split.cs
```diff
@@ -1481,8 +1481,10 @@ private static Cmd AssertIntoAssume(Cmd c)
 
       public void ReleaseChecker()
       {
-        Checker.GoBackToIdle();
-        checker = null;
+        if (checker != null) {
```
keyboardDrummer, line 1484: Can you use `checker?.GoBackToIdle()` ?

keyboardDrummer: It would just do `this.checker = null;` which I think isn't strictly necessary but I'd keep it in for now. I'm guessing `Split` shouldn't have a checker field but that's how it is for now.

MikaelMayer: `this.checker` is never assigned in the above snippet, would it be somewhere else?

keyboardDrummer: I'm not sure I understand your question. `Split.checker` is currently assigned a value in `Split.BeginCheck` and let's keep it that way. I've added a definition of `ResetChecker` in the above snippet to be more clear.

MikaelMayer: Oh ok, that should work given this definition of `ResetChecker`

shazqadeer:
Abandon this PR?

PR #505, opened 2022-02-17
Add more structured descriptions of proof obligations

atomb:
Adds a structured type for describing proof obligations with potentially different text for success and failure forms. As a result, describing successful proofs becomes less awkward. This is more valuable for Boogie clients (such as Dafny) than it is for Boogie itself, though it has the benefit of grouping error messages in one place.

At the moment, Boogie only uses the failure text, but I could imagine it using the success text at some point. Dafny plans to use the success text right away, and to provide a large number of additional subclasses to attach to the instances of `AssertCmd` it generates.

I also _hope_ that it could serve as a more general replacement for `ErrorData`, though I don't have a very thorough idea of how the various Boogie clients currently use that property.

I've included quite a few reviewers in the hope of identifying any incompatibilities this introduces as early as possible.


cpitclaudel (approved):
This looks reasonable to me, but: will clients like Dafny also have to provide two strings everywhere instead of one?  Or will Dafny use a different mechanism altogether to get richer errors?

atomb:
> This looks reasonable to me, but: will clients like Dafny also have to provide two strings everywhere instead of one? Or will Dafny use a different mechanism altogether to get richer errors?

I think Dafny will likely have another file like `ProofObligationDescription.cs` that'll be pretty long, but we can certainly set it up so that we just provide one description per type of proof obligation (the positive one) and set the other to "Unable to prove: blah" by default to make it a little less onerous.

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -1382,23 +1382,24 @@ private static string TimeIndication(ExecutionEngineOptions options, Verificatio
...
                 else if (returnError != null)
                 {
                   tok = returnError.FailingReturn.tok;
-                  msg = "A postcondition might not hold on this return path.";
+                  msg = returnError.FailingReturn.Description.FailureDescription;
                 }
                 else
                 {
                   tok = assertError.FailingAssert.tok;
                   if (assertError.FailingAssert is LoopInitAssertCmd)
```
keyboardDrummer, line 1396: Can we simplify this big `if` into something like?
```csharp
if (assertError.FailingAssert.ErrorMessage == null || options.ForceBplErrors)
{
  msg = assertError.FailingAssert.ErrorData as string;
}
else
{
  msg = assertError.FailingAssert.ErrorMessage;
}
msg ??= assertError.FailingAssert.Description.FailureDescription;
```

Source/Core/AbsyCmd.cs
```diff
@@ -3775,6 +3781,15 @@ public MiningStrategy ErrorDataEnhanced
       set { errorDataEnhanced = value; }
     }
 
+    public AssertCmd(IToken /*!*/ tok, Expr /*!*/ expr, ProofObligationDescription description, QKeyValue kv)
```
keyboardDrummer, line 3784: Can we reduce duplication by letting another constructor call this one with a `new AssertionDescription()` description?

Source/Core/AbsyCmd.cs
```diff
@@ -3762,6 +3766,8 @@ public object ErrorData
       set { errorData = value; }
     }
 
+    public ProofObligationDescription Description { get; set; } = new AssertionDescription();
```
keyboardDrummer, line 3769: Shouldn't this be `static`, there's no interface being implemented right?

keyboardDrummer (changes requested):

Source/ExecutionEngine/ExecutionEngine.cs
atomb: Yeah, definitely.

Source/Core/AbsyCmd.cs
atomb: There's no interface being implemented, but the idea is that some clients may pass in something other than `AssertionDescription`, so not all `AssertCmd` instances will have the same `Description`.

Source/Core/ProofObligationDescription.cs
```diff
@@ -0,0 +1,104 @@
...
+  /// A description of what this proof obligation means (or might mean)
+  /// when the prover has failed to establish its validity.
+  /// </summary>
+  public virtual string FailureDescription =>
+    $"Failed to prove: {SuccessDescription}";
+
+  /// <summary>
+  /// A short description of the general type of this proof obligation.
+  /// </summary>
+  public abstract string ShortDescription { get; }
```
robin-aws, line 28: Is there a commitment that this can be used as a stable identifier for this class of PO? I'm thinking about cases where languages built on top of Boogie may want to replace these more generic, Boogie-specific descriptions with more specific versions. Or is the intention that all client languages will override any such descriptions when creating the Boogie command objects?

Source/Core/ProofObligationDescription.cs
```diff
@@ -0,0 +1,104 @@
...
+
+  public override string FailureDescription => "This assertion might not hold.";
+
+  public override string ShortDescription => "assert";
+}
+
+public class PreconditionDescription : ProofObligationDescription
+{
+  public override string SuccessDescription =>
+    "All preconditions hold for call.";
```
robin-aws, line 43:
```suggestion
    "All preconditions hold for this call.";
```

Source/Core/ProofObligationDescription.cs
```diff
@@ -0,0 +1,104 @@
...
+
+  public override string FailureDescription =>
+    "A precondition for this call might not hold.";
+
+  public override string ShortDescription => "precondition";
+}
+
+public class RequiresDescription : ProofObligationDescription
```
robin-aws, line 51: What is the difference between "preconditions" and "requires"? I'm wondering if we're using mostly the same messages for both because Dafny doesn't distinguish, but there IS a distinction at the Boogie level?

Source/Core/ProofObligationDescription.cs
```diff
@@ -0,0 +1,104 @@
...
+
+/// <summary>
+/// A multi-faceted description of a proof obligation. This class is intended
+/// to provide several forms of human-readable text summarizing the meaning of
+/// a proof obligation: one for the case where the proof has succeeded, one
+/// for the case where the proof has failed, and one for the case of quickly
+/// identifying the type of proof obligation (for inclusion in long lists of
+/// proof obligations, for example).
+/// </summary>
+public abstract class ProofObligationDescription {
```
robin-aws, line 11: Could this be further sub-classed in the future to add more metadata? That could be a good thing (and would make the new field a more future-proof replacement for `ErrorData`). But that might influence the class naming and documentation.

robin-aws (changes requested):

Source/Core/ProofObligationDescription.cs
atomb: Using it as a stable identifier is the idea. And, yes, my intention is that languages that build on Boogie could add to or override any of these. So, when using Boogie alone, you'll have the set of stable identifiers already in this file. Dafny, for example, could keep any that it still finds useful, or could override all of them when creating command objects (as I expect Dafny would do).

Source/Core/ProofObligationDescription.cs
atomb: The way these messages are used, "precondition" refers to the set of all things that must be proved to call a procedure whereas "requires" refers to a specific requires clause. So a precondition is basically a set of requires clauses. Boogie's error messages use both descriptions. Dafny's might not. But to allow this PR to leave all of Boogie's output unchanged it needs to have classes that distinguish the concepts.

Source/Core/ProofObligationDescription.cs
atomb: That's exactly the idea I have in mind. Currently, Dafny passes in strings for the `ErrorData` field. As things are right now, we could begin creating subclasses for _some_ kinds of Dafny assertions, and leave some as strings, allowing us to incrementally migrate error types without, at first, changing any of the messages. Once we've finished that (which might take a while, because there are a lot of them), all of the messages will be in one place and we can review them for consistency, clarity, etc. Ultimately, any subclasses could have any extra metadata that would be useful (and I expect it will be necessary in some cases, because some messages include things references to source constructs in Dafny).

Source/Core/AbsyCmd.cs
MikaelMayer: Can you reduce it further by setting the default value kv = null, since this is what you do in most places?

MikaelMayer (changes requested):
Just a simple change that would remove some of the ", null" but otherwise this is a very good path towards having custom error messages reported !

Source/Core/ProofObligationDescription.cs
robin-aws: I would add a comment to that effect then. I would imagine other clients otherwise being tempted to give short sentences that are not ideal as programmatically-compared identifiers.

Source/Core/ProofObligationDescription.cs
robin-aws: Cool. This does mean that, since the new field is typed to be a `ProofObligationDescription` rather than just an `object`, we are asserting that any tooling built on Boogie should aim to provide such descriptive phrases, and further subclass if they want to add more arbitrary metadata.

Should we deprecate `ErrorData`?

atomb: Yeah, my intent is that new tooling would do exactly that.

I'd personally like to deprecate `ErrorData`, but it could be useful to get feedback from other folks developing Boogie clients before doing so.

robin-aws (approved):

Source/Core/AbsyCmd.cs
keyboardDrummer: How would clients pass in something different? I see that this field has a public setter but at what point would you assign using that? Could you add a unit test that shows how to use the new API ?

To me it seems counterintuitive to have metadata of a type be held in an instance field of said type, as opposed to it being a virtual property/method or a static one.

atomb: Although the Boogie codebase currently has only one type for descriptions of an `AssertCmd`, other language clients will likely have many. So different instances of `AssertCmd` (for instance), will have different `Description` properties, instead of the current case where they have different `ErrorData` values.

Said another way, it's not metadata about a type (even though there's currently only one description appropriate for a given type in the Boogie codebase). It's metadata about an instance of that type, which could be instantiated any way a client chooses. Clients will customize `AssertCmd`s by subclassing `AssertionDescription` rather than by subclassing `AssertCmd`.

atomb: You can see [this commit](https://github.com/atomb/dafny/commit/ad105bbbe3516cfc9aba90472e993dcfc475f069), that I haven't yet turned into a PR, for an example of how that'd work.

MikaelMayer (dismissed):
Looks good !

Source/Core/AbsyCmd.cs
keyboardDrummer: To me it looks like the current API is:

```csharp
// library
record AssertCmd(ProofObligationDescription description) : PredicateCmd {

}

abstract class ProofObligationDescription {
  public abstract string SuccessDescription { get; }
  ...
}

class DefaultAssertDescription {
  public override string SuccessDescription => "foo";
}

// consumer
class SpecificAssertDescription {
  public override string SuccessDescription => "defaultSuccessDescription";
}
class User {
  void Use() {
    var defaultAssertCmd = new AssertCmd(new DefaultAssertDescription());
    var specificAssertCmd = new AssertCmd(new SpecificAssertDescription());
  }
}
```

and my suggestion is to change that to:

```csharp
// library
abstract class AssertCmd : PredicateCmd, ProofObligation {
  public string SuccessDescription => "defaultSuccessDescription";
}

// this interface is optional. 
interface ProofObligation {
  string SuccessDescription { get; }
  ...
}

// consumer
class SpecificAssertCmd : AssertCmd {
  public override string SuccessDescription => "foo";
}

class User {
  void Use() {
    var defaultAssertCmd = new AssertCmd();
    var specificAssertCmd = new SpecificAssertCmd();
  }
}
```

The reason I prefer the latter version is that in this PR, there's only 1 specific value of `ProofObligationDescription` passed to AssertCmd, so it seems strange to have an instance field for that. Yes you want to be able to customise this in consumers, but don't need a field for that since you can also use subclassing.

atomb: That is another reasonable design approach. There are a few reasons I prefer an additional constructor parameter over subclassing in this case.

The reason I most want to emphasize is that, to me, subclassing Boogie commands feels like extending the language rather than just adding metadata (even if the effect in this particular case is the same). I feel like Boogie is more coherent as a compilation target language than as a library that can be extended in arbitrary ways. Attaching complex, structured metadata to compilation targets, on the other hand, is a common thing (see LLVM), and frequently quite useful (as in this case).

Another reason is that the approach you're suggesting widens the interface of commands (at least client-specific subclasses). It would add, say, 3 properties to each derived command, rather than one property with a value that has several properties of its own. Grouping properties like that tends to lead to more general-purpose code and easier-to-learn interfaces, in my experience.

Source/Core/ProofObligationDescription.cs
```diff
@@ -0,0 +1,106 @@
...
+  public override string SuccessDescription =>
+    "This postcondition holds.";
+
+  public override string FailureDescription =>
+    "This is the postcondition that might not hold.";
+
+  public override string ShortDescription => "ensures";
+}
+
+public class InvariantEstablishedDescription : AssertionDescription
```
keyboardDrummer, line 86: Why not place all these description classes where they are used, instead of in this file?

atomb: I see having these messages all in one file as a distinct advantage of this approach. It makes any effort to review all messages for consistency and usability much easier.

Source/Core/AbsyCmd.cs
keyboardDrummer: > I feel like Boogie is more coherent as a compilation target language than as a library that can be extended in arbitrary ways.

My impression is that the API this PR introduces cannot be used through a .bpl program, so it requires consuming Boogie as a library anyways.

> The reason I most want to emphasize is that, to me, subclassing Boogie commands feels like extending the language rather than just adding metadata (even if the effect in this particular case is the same). ... 
Attaching complex, structured metadata to compilation targets, on the other hand, is a common thing (see LLVM), and frequently quite useful (as in this case).

I can see how you feel that. Let's try avoiding the subclasses then. Also, do you happen to have a link to an LLVM example of this?

One more thing, it seems that Boogie uses subclasses of AssertCmd to configure error messages. Would it make sense to stop doing that and use the instance field this PR adds instead? After this PR there's some duplication of classes, for example there's `InvariantEstablishedDescription` and `LoopInitAssertCmd`. Do we still need `LoopInitAssertCmd` now that we have `InvariantEstablishedDescription` ?

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -1750,7 +1735,8 @@ private ErrorInformation CreateErrorInformation(Counterexample error, VC.VCGen.O
         {
           if (assertError.FailingAssert.ErrorMessage == null || Options.ForceBplErrors)
           {
-            string msg = assertError.FailingAssert.ErrorData as string ?? "This assertion might not hold.";
+            string msg = assertError.FailingAssert.ErrorData as string ??
```
keyboardDrummer, line 1727: ErrorData and Description are both ways to customise the error message. Shouldn't there be only one?

keyboardDrummer (approved):

Source/ExecutionEngine/ExecutionEngine.cs
atomb: Marking it as `[Obsolete]` leads to a slew of compilation warnings, and probably will for quite a while, since I think it'll take a while to migrate all of the clients toward `Description`. So I think I want to leave it off for now and discuss what other Boogie-based tool developers would find most workable.

robin-aws: How about just a code comment steering folks away from `ErrorData` then?

atomb: Yep, sounds good!

robin-aws (approved):

PR #508, opened 2022-02-18
Reduce references to `CoreOptions.Clo` from 204 to 92

keyboardDrummer:
- CheckerPool is now a field of ExecutionEngine and ExecutionEngine inherits from `IDisposable`, to give more control over the lifetime of the `CheckerPool` to Boogie consumers and to ensure `CheckerPool` does not hold a different `ExecutionEngineOptions` object then is used by the rest of the code.
  - Snapshots turn out not to work well when reusing the checkerPool, so when snapshots are used the checker pool is not reused for each snapshot.
- Introduce interfaces `ExecutionEngineOptions`, `HoudiniOptions`, `VCGenOptions` and `ConcurrencyOptions` for holding options related to their respective assemblies.
- Reduce the references to `CoreOptions.Clo` from 204 to 92.

### DIFF
--- a/Source/VCGeneration/Split.cs
+++ b/Source/VCGeneration/Split.cs
@@ -17,7 +17,7 @@ namespace VC
 
   public class Split
   {
-      public int? RandomSeed => Implementation.RandomSeed ?? CommandLineOptions.Clo.RandomSeed;
+      public int? RandomSeed => Implementation.RandomSeed ?? Options.RandomSeed;
     
       class BlockStats
       {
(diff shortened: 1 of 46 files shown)


Source/VCGeneration/Split.cs
```diff
@@ -949,13 +953,13 @@ bool ContainsAssert(Block b)
                || c is AssertCmd && splitOnEveryAssert;
       }
       
-      public static List<Split /*!*/> FindManualSplits(Split s, bool splitOnEveryAssert)
+      public static List<Split /*!*/> FindManualSplits(Split initialSplit, bool splitOnEveryAssert)
```
atomb, line 956: This and the rest of the changes to this method are somewhat unrelated to the command-line option changes, but they're all semantics-preserving cleanups, so I'm not really opposed to including them in one PR.

atomb (approved):
This is a welcome cleanup. Thanks for doing it!

atomb (approved):
It'd be worth figuring out why it doesn't work, if it should, but it seems reasonable to use this workaround for now. Is there an issue for that bug?

PR #509, opened 2022-02-21
[Civl] Action inlining

shazqadeer:
This PR adds the following features:
- Allow asserts to be used anywhere in an atomic action
- Allow an action to call another action

To add this feature, the creation of atomic actions inside CivlTypeChecker is split into two parts.  In the first part, basic information about each action is computed but the implementations of the actions are not processed.  In the second part, after linear type checking has been done, the rest of the initialization of the atomic actions is done in two phases.  First, all calls to atomic actions from other atomic actions are inlined. Second, asserts inside the inlined bodies of atomic actions are hoisted to the beginning using a variation of weakest preconditions.


shazqadeer:
A test if failing:

******************** TEST 'Boogie :: civl/inductive-sequentialization/NBuyer.bpl' FAILED ********************
Script:
--
: 'RUN: at line 1';   dotnet /home/runner/work/boogie/boogie/Source/BoogieDriver/bin/Debug/net6.0/BoogieDriver.dll -useBaseNameForFileName /vcsCores:2 "/home/runner/work/boogie/boogie/Test/civl/inductive-sequentialization/NBuyer.bpl" > "/home/runner/work/boogie/boogie/Test/civl/inductive-sequentialization/Output/NBuyer.bpl.tmp"
: 'RUN: at line 2';   /usr/bin/diff --unified=3 --strip-trailing-cr --ignore-all-space "/home/runner/work/boogie/boogie/Test/civl/inductive-sequentialization/NBuyer.bpl.expect" "/home/runner/work/boogie/boogie/Test/civl/inductive-sequentialization/Output/NBuyer.bpl.tmp"
--
Exit Code: 1

In this situation, what is the usual way to debug the problem?  Is there any way to retrieve the file "/home/runner/work/boogie/boogie/Test/civl/inductive-sequentialization/Output/NBuyer.bpl.tmp"?

@keyboardDrummer , @bkragl

bkragl:
There is a diff in the output that shows the failing check.

```
$ "/usr/bin/diff" "--unified=3" "--strip-trailing-cr" "--ignore-all-space" "/home/runner/work/boogie/boogie/Test/civl/inductive-sequentialization/NBuyer.bpl.expect" "/home/runner/work/boogie/boogie/Test/civl/inductive-sequentialization/Output/NBuyer.bpl.tmp"
# command output:
--- /home/runner/work/boogie/boogie/Test/civl/inductive-sequentialization/NBuyer.bpl.expect	2022-02-21 07:15:18.361435598 +0000
+++ /home/runner/work/boogie/boogie/Test/civl/inductive-sequentialization/Output/NBuyer.bpl.tmp	2022-02-21 07:18:23.083527510 +0000
@@ -1,2 +1,8 @@
+NBuyer.bpl(228,3): Error: Gate of LastBuyer' fails in IS step of MAIN3
+Execution trace:
+    NBuyer.bpl(124,1): inline$INV3$0$Entry
+    NBuyer.bpl(130,3): inline$INV3$0$anon0
+    NBuyer.bpl(143,5): inline$INV3$0$anon6_Then
+    NBuyer.bpl(124,1): inline$INV3$0$Return
 
-Boogie program verifier finished with 130 verified, 0 errors
+Boogie program verifier finished with 129 verified, 1 error
```

shazqadeer:
@bkragl :  I would like to land this PR even though not all issues with inlining have been sorted out.  For example, there is proof instability when using action inlining with various examples under inductive-sequentialization.  I will debug these problems gradually. Please provide feedback.

bkragl:
If I remember correctly, the remaining issue is that the heuristic that injects triggers for the initial values of local variables in the transition relation for a commutativity check does not work properly. The reason is that we only put a trigger if the existentially quantified variable is the "first copy" (because that's what we have a corresponding triggering expression for), but this breaks with the havocs coming from inlined calls.

I had a proposal to skip the introduction of a variable copy on a havoc if we are still at the first copy for that variable. There was a concern that this is not sound if the variable was already used, but I think we can fix it and basically eliminate all unnecessary havocs. The idea is a follows. Initialize a set `usedVars` with the first copy of all local variables used in the gate. Then along a path, for assume/assign statements add the current copy of all used variables to `usedVars`. And for a havoc statement, skip the introduction of a new copy for any havoced variable where the current copy is not in `usedVars`. What do you think?

I think this solution would be easy to implement, but if you want we can also punt on it in this PR.

shazqadeer:
As I was digging more into why Paxos is not working, I got puzzled by why triggers work and the pool-based quantifier instantiation does not. I want to try and understand that some more.  Your proposal seems interesting to try on its own.  

I suggest we try to land this PR as it is and leave subsequent experimentation for other PRs.

Source/Concurrency/CivlTypeChecker.cs
```diff
@@ -144,21 +144,29 @@ public void TypeCheck()
       TypeCheckGlobalVariables();
       TypeCheckLemmaProcedures();
       TypeCheckYieldInvariants();
-
-      TypeCheckActionDecls();
+      
+      TypeCheckActions();
```
bkragl, line 148: Probably worth adding a comment about the two-step process of checking/initializing actions and the dependencies between the type checking phases. For example, I'm not sure why the "finish initialization" work cannot be done at the end of `TypeCheckActions`. I guess previously we tried to do all declaration checking first, and then do implementation checking. But now `TypeCheckActions` anyway already looks at action implementations.

Source/Concurrency/CivlTypeChecker.cs
```diff
@@ -342,15 +359,32 @@ private void TypeCheckActionDecls()
...
+      foreach (var action in allActions)
+      {
+        Inliner.ProcessImplementation(Options, program, action.impl);
+      }
+      foreach (var action in allActions)
+      {
+        var impl = action.impl;
+        impl.OriginalBlocks = null;
+        impl.OriginalLocVars = null;
+      }
```
bkragl, line 381: Maybe there should be a helper method for inlining (in `Inliner`?).

Test/civl/assert-not-first-4.bpl
```diff
@@ -0,0 +1,17 @@
+// RUN: %parallel-boogie "%s" > "%t"
+// RUN: %diff "%s.expect" "%t"
+
+procedure {:yields} {:layer 1} P(arr: [int]int, i: int)
+requires {:layer 1} arr[i] == 1;
+{
+    call foo(arr, i);
```
bkragl, line 7: Adding a comment that says that this is a failing example would be helpful. I spent a minute trying to understand it, before I realized that the `.expect` files says that it is supposed to fail :)

Source/Concurrency/CivlTypeChecker.cs
```diff
@@ -1402,17 +1436,32 @@ public override Expr VisitIdentifierExpr(IdentifierExpr node)
...
+        else if (!actionProcToLayerRange[proc].Subset(actionProcToLayerRange[node.Proc]))
+        {
+          civlTypeChecker.Error(node, "Caller layer range must be subset of callee layer range");
+        }
+        else
+        {
+          callGraph.AddEdge(proc, node.Proc);
+        }
         return base.VisitCallCmd(node);
       }
```
bkragl, line 1457: There is a soundness issue here where an atomic action can call an introduction action, thereby getting access to a global variable a layer to early. Imagine atomic action `A`, introduction action `I`, and global variable `g`, all with singleton layer range `[1]`. Then `I` is allowed to access `g`, but `A` is not. However, if `A` calls `I` then the access to `g` will get inlined into `A`.

bkragl (commented):
Nice feature for the usability of Civl! Most of my comments are minor, with the exception of one soundness issue.

Source/Concurrency/CivlTypeChecker.cs
shazqadeer: Added another case in the type checking above.  Also added a test case.

Source/Concurrency/CivlCoreTypes.cs
```diff
@@ -120,9 +120,92 @@ protected List<Variable> AssignedVariables()
...
+      }
+      block.Cmds = block.Cmds.Where(cmd => !(cmd is AssertCmd)).ToList();
+      return postconditions;
+    }
+    
+    private static AssertCmd Forall(IEnumerable<Variable> vars, AssertCmd assertCmd)
+    {
+      var freeObjects = new GSet<object>();
+      assertCmd.Expr.ComputeFreeVariables(freeObjects);
+      var quantifiedVars = freeObjects.OfType<Variable>().Intersect(vars);
```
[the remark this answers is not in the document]
shazqadeer: I don't know.
I don't know.
Yes.

Source/Concurrency/CivlTypeChecker.cs
shazqadeer: I will do this in a separate PR also.

Test/civl/inductive-sequentialization/NBuyer.bpl
```diff
@@ -154,6 +154,7 @@ modifies QuoteCH, RemCH, DecCH, contribution;
     RemCH := (lambda i:int :: (lambda r:int :: if i == n && r == price - sum(contribution, 1, n-1) then 1 else 0));
     PAs := MapAddPA(SingletonPA(SellerFinish(0)), SingletonPA(LastBuyer(n)));
     choice := LastBuyer(n);
+    assume RemCH[n][price - sum(contribution, 1, n-1)] > 0;
```
bkragl, line 157: The two new assumes are not necessary, if `n-1` is added to the `INV3` pool.

Source/Concurrency/CivlCoreTypes.cs
```diff
@@ -434,6 +529,11 @@ public override BinderExpr VisitBinderExpr(BinderExpr node)
         }
       }
 
+      if (node.Attributes != null)
+      {
+        expr.Attributes = VisitQKeyValue(node.Attributes);
+      }
+
```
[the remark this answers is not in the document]
shazqadeer: But we are overriding StandardVisitor.VisitBinderExpr, so it won't be called, right?

Test/civl/inductive-sequentialization/NBuyer.bpl
shazqadeer: I added n-1 to INV3 and also attached INV3 to the lambda bound variables.   The proof succeeds on my machine.

Source/Concurrency/CivlCoreTypes.cs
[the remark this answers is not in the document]
bkragl: It should, because we are calling `base.VisitBinderExpr(node)`.

bkragl (approved):

PR #510, opened 2022-02-22
Remove all references to CoreOptions.Clo

keyboardDrummer:
- Add a fix to `Source/VCGeneration/Checker.GoBackToIdle` for when a model view file is used.
- ExecutionEngine takes a `VerificationResultCache` as an argument instead of storing it in a static variable, so consumers have control over the lifetime of the cache. Consumers who want to migrate cheaply should create their own static `VerificationResultCache`.
- Remove all references to `CoreOptions.Clo`

### DIFF
--- a/Source/ExecutionEngine/CommandLineOptions.cs
+++ b/Source/ExecutionEngine/CommandLineOptions.cs
@@ -2030,7 +2030,7 @@ namespace Microsoft.Boogie
       return semicolonIndex;
     }
 
-    public string ProverHelp => TheProverFactory.BlankProverOptions().Help;
+    public string ProverHelp => TheProverFactory.BlankProverOptions(this).Help;
 
     public override string AttributeHelp =>
 @"Boogie: The following attributes are supported by this version.
--- a/Source/ExecutionEngine/CommandLineOptions.cs
+++ b/Source/ExecutionEngine/CommandLineOptions.cs
@@ -651,7 +651,6 @@ namespace Microsoft.Boogie
     public static void Install(CoreOptions options)
     {
       Contract.Requires(options != null);
-      CoreOptions.Clo = options;
     }
 
     // Flags and arguments
--- a/Source/ExecutionEngine/CommandLineOptions.cs
+++ b/Source/ExecutionEngine/CommandLineOptions.cs
@@ -108,17 +108,6 @@ namespace Microsoft.Boogie
       }
     }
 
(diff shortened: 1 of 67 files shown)


Source/VCGeneration/Checker.cs
```diff
@@ -61,6 +61,12 @@ public void GetReady()
     public void GoBackToIdle()
     {
       Contract.Requires(IsBusy);
+      if (Options.ModelViewFile != null) {
+        // Don't re-use theorem provers whose ProverContext still needs to be queried to extract model data.
+        Pool.CheckerDied();
+        Close();
+        return;
+      }
```
atomb, line 69: It seems slightly unfortunate to have what seems like a functional change among what is mostly refactoring, though it seems like a good change.

atomb (approved):
Looks good. Large but straightforward.

atomb (dismissed):
Looks good!

Source/ExecutionEngine/CommandLineOptions.cs
```diff
@@ -108,6 +108,17 @@ public string FileTimestamp
...
+    {
+      if (pattern != null)
+      {
+        pattern = pattern.Replace("@PREFIX@", logPrefix).Replace("@TIME@", fileTimestamp);
+        string fn = Files.Count == 0 ? "" : Files[Files.Count - 1];
+        fn = Util.EscapeFilename(fn);
+        pattern = pattern.Replace("@FILE@", fn);
+      }
+    }
+
```
atomb, line 121: Is this version of `ExpandFilename` used anywhere?

atomb (approved):
It seems that you might have added a method that's not called anywhere? And I see that there's an easily-fixable compilation error. But I'll leave it up to CI to make the decision about that. :)

Source/ExecutionEngine/CommandLineOptions.cs
keyboardDrummer: Only in Boogie consumers

Source/VCGeneration/Checker.cs
keyboardDrummer: I've added this bugfix in this PR because I'm testing this Boogie PR against Dafny master and the current Boogie master does not work with Dafny without this fix.

I will add a test in Boogie for the bug this fix is for as well but I didn't get to that yet.

atomb (approved):

PR #512, opened 2022-02-22
Refactor solver process handling to allow non-interactive mode

### LINKED ISSUE #511: Support invoking solvers in batch mode
# Purpose

Currently, Boogie always invokes solvers in interactive mode, sending commands and reading responses incrementally and re-using a specific running solver for multiple queries. This is typically the most efficient mode for solvers that support it, but some solvers expect an entire input file to be ready on startup. Even for solvers that _can_ work in interactive mode, it can occasionally be useful to run them in batch mode.

# Implementation Proposal

The `SMTLibProcessTheoremProver` class is the primary source of interactivity with solvers. I propose turning it into an abstract class that can have multiple implementations. Two subclasses, let's call them `SMTLibInteractiveTheoremProver` and `SMTLibBatchTheoremProver`, could implement certain methods differently. I have an experimental local implementation of `SMTLibInteractiveTheoremProver` that overrides the following methods from `ProverInterface`:

* `GoBackToIdle`
* `FlushAxiomsToTheoremProver`
* `Close`
* `BeginCheck`
* `Reset`
* `FullReset`
* `CheckOutcome`
* `CheckOutcomeCore`
* `Evaluate`
* `UnsatCore`
* `Check`
* `Send`
* `GetRCount`
* `CheckAssumptions`

Some of these methods (and the overall design of `ProverInterface`) clearly assume the possibility of solver interaction. However, my current belief is that no-op versions of many of those methods should work out for a batch mode solver.  The following methods probably can't be implemented for a batch solver (or at least not easily):

* `Evaluate`
* `CheckAssumptions`

It seems as though `Evaluate` is never used, and `CheckAssumptions` is used only by Houdini. So, at least for now, I think batch mode would be incompatible with Houdini.

atomb:
This PR moves portions of the `SMTLibProcessTheoremProver` class into a subclass called `SMTLibInteractiveTheoremProver`. The goal is that everything in `SMTLibProcessTheoremProver` should be useful for either interactive or batch mode solvers. It then adds another subclass, `SMTLibBatchTheoremProver`, to address issue #511.

Apologies for the large diff. It consists almost entirely of moving code, with some refactoring to separate process interaction from result parsing (to ensure that the latter can be used in batch mode, too).

Fixes #511.


cpitclaudel:
Could you set up the branch so that the diff is smaller?  Save a copy of the two files, copy the original file to create the new one (https://stackoverflow.com/a/46484848/695591), commit, then restore your changes by copying them back over the two files, and commit again; this should give diffs only against the original file with many removed lines but very few added ones.

atomb:
> Could you set up the branch so that the diff is smaller? Save a copy of the two files, copy the original file to create the new one (https://stackoverflow.com/a/46484848/695591), commit, then restore your changes by copying them back over the two files, and commit again; this should give diffs only against the original file with many removed lines but very few added ones.

I've done that, and the result is somewhat strange. It leads to better diffs locally, with `git diff`, but the diff displayed on GitHub looks the same.

atomb:
For some more specific guidance on reviewing, I recommend focusing on [this commit](https://github.com/boogie-org/boogie/pull/512/commits/78959a52663bf9983d2eb8c6d6cd2aacf73b8fa3). The two files it lists started out identical, and then I removed large swaths of each so that one now focuses on generic functionality and one focuses on interactive-specific functionality. So it should be reasonable to largely skim over the deletions and focus on the additions, where something actually needed to change to make the separation work (or, in some cases, to do minor cleanups).

keyboardDrummer:
Shouldn't the batch solver also be in this PR to enable getting the base class code right?

atomb:
> Shouldn't the batch solver also be in this PR to enable getting the base class code right?

I have written the batch solver, but left it out of this PR to make it slightly less enormous. I would be happy to include it, though.

keyboardDrummer:
> > Shouldn't the batch solver also be in this PR to enable getting the base class code right?
> 
> I have written the batch solver, but left it out of this PR to make it slightly less enormous. I would be happy to include it, though.

I think that would help, then I'll review the resulting code as a whole, looking more at the end result than the diff.

Source/Provers/SMTLib/NoopSolver.cs
```diff
@@ -14,33 +15,35 @@ public override void Close()
   {
   }
 
-  private SExpr response;
+  private Queue<SExpr> responses = new Queue<SExpr>();
```
cpitclaudel, line 18: I think you can say `new ()`

Source/Provers/SMTLib/SMTLibProcess.cs
```diff
@@ -92,12 +92,18 @@ void ControlCHandler(object o, ConsoleCancelEventArgs a)
       }
     }
 
+    public override void IndicateEndOfInput()
+    {
+      prover.StandardInput.Close();
+      toProver = null;
```
cpitclaudel, line 98: Instead of setting this to null here and checking for null below, could we check whether toProver has been closed?

Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
```diff
@@ -32,7 +32,6 @@ void ObjectInvariant()
       Contract.Invariant(_backgroundPredicates != null);
     }
 
-    [NotDelayed]
```
atomb, line 35: Didn't mean to remove this.

Source/Provers/SMTLib/SMTLibBatchTheoremProver.cs
```diff
@@ -0,0 +1,318 @@
...
+    {
+      return Task.CompletedTask;
+    }
+
+    private void SetupProcess()
+    {
+      Process?.Close();
+      Process = options.Solver == SolverKind.NoOpWithZ3Options ? new NoopSolver() : new SMTLibProcess(libOptions, options);
+      Process.ErrorHandler += HandleProverError;
+      CheckSatSent = false;
```
atomb, line 37: Could be more sharable with superclass.

Source/Provers/SMTLib/SMTLibBatchTheoremProver.cs
```diff
@@ -0,0 +1,318 @@
...
+    {
+      // we feed the axioms when BeginCheck is called.
+      return 0;
+    }
+
+    public override void Close()
+    {
+      base.Close();
+      Process?.Close();
+      Process = null;
```
atomb, line 50: Could be shared?

Source/Provers/SMTLib/SMTLibBatchTheoremProver.cs
```diff
@@ -0,0 +1,318 @@
...
+      FlushLogFile();
+    }
+
+    public override void Reset(VCExpressionGenerator generator)
+    {
+    }
+
+    public override void FullReset(VCExpressionGenerator generator)
+    {
+      if (options.Solver == SolverKind.Z3 || options.Solver == SolverKind.NoOpWithZ3Options)
```
cpitclaudel, line 91: Odd that this is conditional on Z3

Source/Provers/SMTLib/SMTLibBatchTheoremProver.cs
```diff
@@ -0,0 +1,318 @@
...
+      return result;
+    }
+
+    [NoDefaultContract]
+    public override async Task<Outcome> CheckOutcomeCore(ErrorHandler handler, CancellationToken cancellationToken,
+      int errorLimit)
+    {
+      if (Process == null || proverErrors.Count > 0) {
+        return Outcome.Undetermined;
+      }
```
atomb, line 122: May not be necessary anymore?

Source/Provers/SMTLib/SMTLibBatchTheoremProver.cs
```diff
@@ -0,0 +1,318 @@
...
+      var rlimitSExp = await Process.GetProverResponse().WaitAsync(cancellationToken);
+      resourceCount = ParseRCount(rlimitSExp);
+
+      var modelSExp = await Process.GetProverResponse().WaitAsync(cancellationToken);
+      errorModel = ParseErrorModel(modelSExp);
+
+      return result;
+    }
+
+    private string[] CalculatePath(int controlFlowConstant, Model model)
```
atomb, line 172: Put in a note that we could do `(get-value ControlFlow)` instead of `(get-model)`.

Source/Provers/SMTLib/SMTLibBatchTheoremProver.cs
```diff
@@ -0,0 +1,318 @@
...
+          break;
+        }
+
+        v = result;
+      }
+
+      return path.ToArray();
+    }
+
+    private Model ParseErrorModel(SExpr resp)
```
atomb, line 208: Could this be in the superclass and used by both batch and interactive?

Source/Provers/SMTLib/SMTLibProcess.cs
atomb: It doesn't look like there's a way to check that. :(

Source/Provers/SMTLib/SMTLibBatchTheoremProver.cs
atomb: That's inherited from the original code, so I'm inclined to leave it that way.

cpitclaudel (dismissed):
Thanks for the fixes!

keyboardDrummer:
Could you document what types of solvers the interactive and the batch handlers work with? I have the impression the batch solver only sends a single sequence of inputs and then receives a single sequence of output. What's the disadvantage of this approach? I see that some features are not supported in batch mode, but for the features that are supported, do we need to handle interactive and batch mode differently?

atomb:
> Could you document what types of solvers the interactive and the batch handlers work with? I have the impression the batch solver only sends a single sequence of inputs and then receives a single sequence of output. What's the disadvantage of this approach? I see that some features are not supported in batch mode, but for the features that are supported, do we need to handle interactive and batch mode differently?

Both interactive and batch modes work with Z3, and I believe with CVC5, as well. Batch mode is closer to what we'd need to work with existing first-order solvers like Vampire (though there are other changes that would likely be needed for it).

The disadvantage of batch mode is that you can't send a command based on information derived from a solver response. For supported features, I don't think the rest of the Boogie code base, or Boogie clients, should need to even know whether the solver is operating in interactive or batch mode.

I'll add a comment on the `SMTLibBatchTheoremProver` class documenting all this.

keyboardDrummer (approved):
Approving, although I stil wonder whether we could more clearly specify in C# what interface batch vs interactive solvers support.

Something I'm not sure of is what the difference in responsibility is between `SMTLibProcessTheoremProver` and `SMTLibProcess`. I wonder whether those two should be merged. Because right now they're not merged, one might think that the lowest level solver interface is defined by `SMTLibProcess`, which then makes it surprising that the batch and interactive theorem prover both use the same `SMTLibProcess` class.

For a batch solver, I'd expect the two methods:

```
  public abstract void Send(string cmd);
  public abstract Task<SExpr> GetProverResponse();
```

to instead look like:

```
  public abstract IAsyncEnumerable<SExpr> Request(IEnumerable<string> cmds);
```

Source/Provers/SMTLib/ProverUtil.cs
```diff
@@ -82,6 +84,7 @@ public virtual string Help
 MEMORY_LIMIT=<int>        Memory limit of the prover in megabytes.
 VERBOSITY=<int>           The higher, the more verbose.
 TIME_LIMIT=<uint>          Time limit per verification condition in milliseconds.
+BATCH_MODE=<bool>         If true, run in batch mode rather than interactive mode.
```
robin-aws, line 87: Consider adding a "See here for details" link of some kind, since "batch mode" is not a standard solver concept AFAICT.

robin-aws (changes requested):
This looks like a great change. I agree with @keyboardDrummer's points about being clearer about what "batch mode" supports, but I don't consider it a blocker either.

I do think we need more testing coverage though. I feel like running more than just a single input program through this separate, very low-level interface is going to bump up the coverage quite a lot. I would almost suggest finding a way to run the majority of the test quite through both modes, with the exception of those programs that wouldn't support batch mode verification (if the input program can even include anything not supported that is!).

Is there a way we could convince lit to run a test case through both prover modes and assert the behavior matches? We could always define additional `%` substitutions...

keyboardDrummer:
> I think that'd be a nicer API, as well. The downside would be that those methods are currently part of the top-level interface that both types of solvers implement, and therefore clients of that interface would need to work differently based on whether they're using a batch or interactive solver. 

I don't think so, those are methods exposed by SMTLibSolver and consumed by SmtLibProcessTheoremProver, so they're not part of the API of `SmtLibProcessTheoremProver`.

Consider introducing a `BatchSolver` class that wraps around a `SMTLibSolver` and exposes
```
public abstract IAsyncEnumerable<SExpr> Request(IEnumerable<string> cmds);
```
which is then used by `SMTLibBatchTheoremProver`

I think this would force `SMTLibBatchTheoremProver` to correctly first send all commands and then retrieve all responses, instead of having that class rely on good intentions as it does now.

Test/prover/batch-solver.bpl
```diff
@@ -0,0 +1,8 @@
+// RUN: %boogie /trace /proverOpt:BATCH_MODE=true "%s" > "%t"
```
keyboardDrummer, line 1: I might be good to convert some number of existing tests to use batch mode, say 20 of them.

atomb:
> > I think that'd be a nicer API, as well. The downside would be that those methods are currently part of the top-level interface that both types of solvers implement, and therefore clients of that interface would need to work differently based on whether they're using a batch or interactive solver.
> 
> I don't think so, those are methods exposed by SMTLibSolver and consumed by SmtLibProcessTheoremProver, so they're not part of the API of `SmtLibProcessTheoremProver`.
> 
> Consider introducing a `BatchSolver` class that wraps around a `SMTLibSolver` and exposes
> 
> ```
> public abstract IAsyncEnumerable<SExpr> Request(IEnumerable<string> cmds);
> ```
> 
> and using that in `SMTLibBatchTheoremProver` instead of a `SMTLibSolver`.
> 
> I think this would force `SMTLibBatchTheoremProver` to correctly first send all commands and then retrieve all responses, instead of having that class rely on the programmer not making any mistakes.

I'll give it a try. It would be nice not to depend on the client calling methods in the right order.

Test/prover/batch-solver.bpl
atomb: I'm currently working on getting all except for a select subset running in batch mode, as it turns out. Hence the failing test runs at the moment while I debug the script. :)

atomb:
Unfortunately, I think that setting things up with a single `Request` method as you suggest, @keyboardDrummer, would require more refactoring of the `SMTLibProcess` class than I'd want to include in this PR (given its complexity already). The reason is that the current implementation makes it very hard to know when there's no more output from the solver. It essentially assumes that if the client asks for more output then it should exist, and blocks until such output is available. So creating an `IAsyncEnumerable` out of that output would be tricky. I think the code could be refactored to make that no longer true, but it would be a significant change.

Source/Provers/SMTLib/ProverUtil.cs
atomb: I reworded it in a way that's hopefully more self-explanatory.

This is perhaps an argument for Boogie to have some documentation, though. :) There's currently none in the repository.

.github/workflows/test.yml
```diff
@@ -53,7 +54,7 @@ jobs:
           # Run unit tests
           dotnet test --no-build -c ${{ matrix.configuration }} ${SOLUTION}
           # Run lit test suite
-          lit -v --timeout=120 -D configuration=${{ matrix.configuration }} Test
+          ./Test/test-with-param.sh ${{ matrix.boogie_param }} -v --timeout=120 -D configuration=${{ matrix.configuration }}
```
robin-aws, line 57: I'm pretty sure you can achieve the same thing by doing the following, instead of adding the extra scripting layer on top:

1. Define an extra parameter for lit that can be passed on the command line, something like `prover-batch-mode`. See https://github.com/dafny-lang/dafny/blob/master/Test/lit.site.cfg#L134 for how the Dafny config adds a `dafny_params` parameter (although in this case a boolean switch probably makes more sense).
2. Add a custom lit "feature" if and only if batch mode is turned on. Dafny uses this to communicate what platform the tests are running on: https://github.com/dafny-lang/dafny/blob/master/Test/lit.site.cfg#L154-L161
3. Use the `UNSUPPORTED` lit command prefix to skip tests that don't support the "feature" above.

atomb: Oh, nice! I'll give that a try.

Test/lit.site.cfg
```diff
@@ -100,6 +100,13 @@ boogieParams = lit_config.params.get('boogie_params', '')
 if len(boogieParams) > 0:
     boogieExecutable = boogieExecutable + ' ' + boogieParams
 
+# Allow user to turn on batch mode in a way that allows skipping
+# tests that don't support it.
+batchMode = lit_config.params.get('batch_mode', '')
+if batchMode:
```
robin-aws, line 106: Careful! `batchMode` is still just a string here, and Python considers any non-empty string to be truthy. That's why in the CI for this latest commit you can see by the number of unsupported tests that it's running in batch mode all the time. :)

You probably want to make use of `str2bool` here: https://docs.python.org/3/distutils/apiref.html#distutils.util.strtobool

robin-aws (changes requested):
Much better! Aside from one rather unfortunate bug :)

Test/lit.site.cfg
atomb: Or it will be once I get the right Python packages installed in CI.

robin-aws (approved):

keyboardDrummer:
> Unfortunately, I think that setting things up with a single `Request` method as you suggest, @keyboardDrummer, would require more refactoring of the `SMTLibProcess` class than I'd want to include in this PR (given its complexity already). The reason is that the current implementation makes it very hard to know when there's no more output from the solver. It essentially assumes that if the client asks for more output then it should exist, and blocks until such output is available. So creating an `IAsyncEnumerable` out of that output would be tricky. I think the code could be refactored to make that no longer true, but it would be a significant change.

Would it help to pass in a `int responseCount` in the initial call?

keyboardDrummer:
> It looks like it's too late to include any changes like that in this PR, but I'd be interested in exploring it more in future PRs that touch this portion of the code (of which I expect there will be a few).

Okay, I think I might be rewarding to make this change, but feel free to leave it for later :-) Maybe I'll try to make it.

PR #515, opened 2022-02-25
Support XML generation with /vcsCores

### LINKED ISSUE #460: /xml feature breaks when verifying multiple VCs in parallel

atomb:
This PR updates the process of collecting verification results and generating XML from them so that it should be safe to use `/xml` and `/vcsCores` together without getting mangled, interleaved output.

In the process, it makes the results of checking individual splits more readily available at the higher levels of the verification call tree. This should be a step toward allowing clients such as Dafny to access fine-grained verification results without needing to read the XML output. The `VerificationResult` type now includes a list of `SplitResult` objects. If there's any more information a client is likely to need from verification, it should probably wind up in either `VerificationResult` or `SplitResult`.

Fixes #460

### DIFF
--- a/Source/Core/Xml.cs
+++ b/Source/Core/Xml.cs
@@ -43,7 +43,7 @@ namespace Microsoft.Boogie
       }
 
       cce.BeginExpose(this);
-      lock (this) {
+      {
         XmlWriterSettings settings = new XmlWriterSettings();
         settings.Indent = true;
         wr = XmlWriter.Create(filename, settings);
@@ -62,7 +62,7 @@ namespace Microsoft.Boogie
       if (wr != null)
       {
         cce.BeginExpose(this);
-        lock (this) {
+        {
           wr.WriteEndDocument();
           wr.Close();
           wr = null;
@@ -81,7 +81,7 @@ namespace Microsoft.Boogie
       Contract.Ensures(IsOpen);
       Contract.Assert(wr != null);
       cce.BeginExpose(this);
-      lock(this) {
+      {
         wr.WriteStartElement("method");
         wr.WriteAttributeString("name", methodName);
         wr.WriteAttributeString("startTime", startTime.ToString(DateTimeFormatString));
@@ -97,7 +97,7 @@ namespace Microsoft.Boogie
       Contract.Ensures(IsOpen);
       Contract.Assert(wr != null);
       cce.BeginExpose(this);
-      lock (this) {
+      {
         wr.WriteStartElement("conclusion");
         wr.WriteAttributeString("endTime", endTime.ToString(DateTimeFormatString));
         wr.WriteAttributeString("duration", elapsed.TotalSeconds.ToString());
@@ -124,7 +124,7 @@ namespace Microsoft.Boogie
       Contract.Assert(wr != null);
 
       cce.BeginExpose(this);
-      lock (this) {
+      {
         wr.WriteStartElement("split");
         wr.WriteAttributeString("number", splitNum.ToString());
         wr.WriteAttributeString("startTime", startTime.ToString(DateTimeFormatString));
@@ -148,7 +148,7 @@ namespace Microsoft.Boogie
       Contract.Ensures(IsOpen);
       Contract.Assert(wr != null);
       cce.BeginExpose(this);
-      lock (this) {
+      {
         wr.WriteStartElement("error");
         wr.WriteAttributeString("message", message);
         WriteTokenAttributes(errorToken);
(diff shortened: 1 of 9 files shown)


Source/VCGeneration/Split.cs
```diff
@@ -15,6 +15,14 @@ namespace VC
   using Bpl = Microsoft.Boogie;
   using System.Threading.Tasks;
 
+  public record SplitResult
```
robin-aws, line 18: I don't love spreading the type "SplitResult" around, since it perpetuates the confusing overloading of "Split" as both the act of splitting larger VCs into smaller ones and the label used to refer to those smaller VCs themselves. Could we use "VerificationConditionResult" or "VCResult" instead? We could replace `int splitNum` with a more general "string vcIdentifier".

Source/Core/Xml.cs
```diff
@@ -114,36 +114,26 @@ public void WriteEndMethod(string outcome, DateTime endTime, TimeSpan elapsed, i
...
       Contract.Requires(splitNum > 0);
+      Contract.Requires(outcome != null);
       Contract.Requires(IsOpen);
       //modifies this.*;
       Contract.Ensures(IsOpen);
       Contract.Assert(wr != null);
+
       cce.BeginExpose(this);
-      {
+      lock (this) {
```
robin-aws, line 127: Is there any point in adding locking in the `XmlSink` class itself, given you have to lock around the sequence of multiple calls on the caller side anyway?

robin-aws (commented):
🥇

Source/VCGeneration/Split.cs
atomb: I think that's a very good point. I'll rename both of those things.

Source/Core/Xml.cs
atomb: Perhaps not. I added them before realizing that there would still need to be a top-level lock. I'll experiment.

Source/VCGeneration/Split.cs
atomb: As for the identifier, I suspect it's going to make sense to add things to this record in the future, likely including a list of `ProofObligationDescription` from #505.

One thing to consider is that the XML file still uses the term `split`. We'll want to change that before long, too.

keyboardDrummer (approved):

keyboardDrummer (approved):

Source/VCGeneration/Split.cs
```diff
@@ -15,6 +15,14 @@ namespace VC
   using Bpl = Microsoft.Boogie;
   using System.Threading.Tasks;
 
+  public record VCResult
```
robin-aws, line 18: This change overall LGTM, but it made me realize that we're not yet recording the resource count for splits, only for methods. :( Can we fix that in the next PR?

robin-aws (approved):

Source/VCGeneration/Split.cs
atomb: That just occurred to me independently, too. And, yes, I'd very much like to add that ASAP.

PR #520, opened 2022-03-01
Return verification result tasks

keyboardDrummer:
- Let `InferAndVerify` return a `Task` that does implementation specific work asynchronously, making it easier for Boogie client to concurrently verify implementations from multiple Boogie programs, using code like `Task.WhenAll(programs.Select(program => engine.InferAndVerify(program,...))`.
- Let multiple calls to `InferAndVerify` on the same `ExecutionEngine` instance share the allowed amount of parallelism that's configured by the VcsCores option.
- XML output from implementations is kept in order when using parallel Boogie

Refactoring:
- Move the static field `ExecutionEngine.printer` to an instance field of `CommandLineOptions`.
- Move the static field `ExecutionEngine.ModelWriter` to an instance field of `CommandLineOptions`.


Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -918,26 +790,86 @@ public void Inline(Program program)
...
+      // {
+      //   OtherDefinitionAxiomsCollector.Collect(Options, program.Axioms);
+      //   DependencyCollector.Collect(Options, program);
+      //   stablePrioritizedImpls = impls.OrderByDescending(
+      //     impl => impl.Priority != 1 ? impl.Priority : Cache.VerificationPriority(impl, Options.RunDiagnosticsOnTimeout)).ToArray();
+      // }
+      // else
+      // {
+      //   stablePrioritizedImpls = impls.OrderByDescending(impl => impl.Priority).ToArray();
+      // }
```
atomb, line 848: Is this comment obsolete, part of refactoring experimentation?

atomb (changes requested):
This looks very nice! The only thing I'd change is to remove a chunk of commented-out code that looks obsolete.

Source/BoogieDriver/BoogieDriver.cs
```diff
@@ -11,11 +11,10 @@ public static int Main(string[] args)
     {
       Contract.Requires(cce.NonNullElements(args));
 
-      var options = new CommandLineOptions
+      var options = new CommandLineOptions(new ConsolePrinter())
```
cpitclaudel, line 14: Are we piggybacking on the command line options object to pass the printer around for simplicity because the options object is already being passed around?

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -918,26 +790,86 @@ public void Inline(Program program)
...
+        stablePrioritizedImpls = impls.OrderByDescending(impl => impl.Priority).ToArray();
+      }
+
+      return stablePrioritizedImpls;
+    }
+
+    private void TraceCachingForBenchmarking(PipelineStatistics stats,
+      string requestId, DateTime start)
+    {
+      if (0 <= Options.VerifySnapshots && Options.TraceCachingForBenchmarking) {
```
cpitclaudel, line 870: `0 <=` is surprising (was expecting `0 <`), though I realize this isn't from your code.

cpitclaudel (approved):
Looks good to me.  I added a few naive questions in places where I had trouble following.  Can't claim that I understand everything that's going on 100%.

My only concern is whether we should change the existing API vs adding a new endpoint InferAndVerifyAsync

Source/BoogieDriver/BoogieDriver.cs
keyboardDrummer: I'd say yes but I'd also say I'd like for `CommandLineOptions` to really be `Options`, not pertaining to the command line per se, so we can use that for any configuration we have for 'headless Boogie'

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -710,7 +664,7 @@ public void Inline(Program program)
     ///  - VerificationCompleted if inference and verification completed, in which the out
     ///    parameters contain meaningful values
     /// </summary>
-    public PipelineOutcome InferAndVerify(
+    public async Task<PipelineOutcome> InferAndVerify(
```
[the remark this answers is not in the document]
keyboardDrummer: So we can keep the old function and be more backwards compatible, or to have Async as a suffix in the name?

Keeping an `InferAndVerify` like the following around seems a little overkill to me:

```
public PipelineOutcome InferAndVerify(...) {
  return InferAndVerifyAsync(...).Result;
}
```

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -749,15 +695,11 @@ public void Inline(Program program)
 
       #endregion
 
-      #region Infer invariants using Abstract Interpretation
```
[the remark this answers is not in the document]
keyboardDrummer: You can fold on the `if` statement as well.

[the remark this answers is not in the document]
cpitclaudel: Your call :)

atomb (changes requested):
This seems like it'll make it much easier to manage the execution of verification!

The one change I'd really like to see is that I think it's worth providing a synchronous wrapper around a new `InferAndVerifyAsync` method, for backward compatibility. The rest of my comments are minor and feel free to take them or leave them.

Source/VCGeneration/ConditionGeneration.cs
```diff
@@ -117,39 +117,30 @@ public ConditionGeneration(Program p, CheckerPool checkerPool)
     /// each counterexample consisting of an array of labels.
     /// </summary>
     /// <param name="impl"></param>
-    public Outcome VerifyImplementation(ImplementationRun run, out List<Counterexample> /*?*/ errors,
-      out List<VCResult> vcResults,
-      string requestId, CancellationToken cancellationToken)
+    public async Task<(Outcome, List<Counterexample> /*?*/ errors, List<VCResult> vcResults)>
+      VerifyImplementation(ImplementationRun run, string requestId, CancellationToken cancellationToken)
```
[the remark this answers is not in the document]
MikaelMayer: Which I did now :-)

Source/Houdini/HoudiniSession.cs
```diff
@@ -394,8 +394,8 @@ private VCExpr BuildAxiom(ProverInterface proverInterface, Dictionary<Variable,
       do
       {
         hardAssumptions.Add(controlExprNoop);
-        (outcome, var unsatisfiedSoftAssumptions) = proverInterface.CheckAssumptions(hardAssumptions,
-          softAssumptions, handler, CancellationToken.None).Result;
+        (outcome, var unsatisfiedSoftAssumptions) = proverInterface.CheckAssumptions(hardAssumptions, softAssumptions,
+          handler, CancellationToken.None).Result;
```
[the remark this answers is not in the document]
keyboardDrummer: It 119 chars wide now

Source/Core/Xml.cs
```diff
@@ -44,15 +44,13 @@ public string Open()
...
-        wr.WriteAttributeString("version", options.VersionNumber);
-        wr.WriteAttributeString("commandLine", Environment.CommandLine);
-      }
+      XmlWriterSettings settings = new XmlWriterSettings();
+      settings.Indent = true;
+      wr = XmlWriter.Create(filename, settings);
+      wr.WriteStartDocument();
+      wr.WriteStartElement("boogie");
+      wr.WriteAttributeString("version", options.VersionNumber);
+      wr.WriteAttributeString("commandLine", Environment.CommandLine);
```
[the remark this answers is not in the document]
keyboardDrummer: Do you want me to take them out? They seem like minor improvements.

Source/ExecutionEngine/ExecutionEngine.cs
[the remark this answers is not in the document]
keyboardDrummer: > Another advantage is that Dafny, as-is, would still be compatible with this code in the interim with that wrapper in place. 

It wouldn't be. Here's the two commits needed to adapt Dafny to the latest Boogie including this PR: 
- https://github.com/keyboardDrummer/dafny/commit/2bea7a831eec3f4a4aaf761593412e051e909825
- https://github.com/keyboardDrummer/dafny/commit/01162e3626d2731da3422bc42d0f8b4b7166d179

> I'd suggest having a wrapper like that given that a lot of external folks use Boogie and we can't very easily know how many use this function.

Added one for `InferAndVery` and `ProcessProgram`, with an `[Obsolete]` attribute on them.

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -268,13 +258,19 @@ public bool ProcessFiles(TextWriter output, IList<string> fileNames, bool lookFo
...
-      if (program == null)
-      {
+      if (program == null) {
         return true;
       }
+
       return ProcessProgram(output, program, bplFileName, programId).Result;
     }
 
+    [Obsolete("Please inline this method call")]
```
atomb, line 268: I'd slightly prefer to not make this (or `InferAndVerify`) obsolete quite yet.

Source/VCGeneration/ConditionGeneration.cs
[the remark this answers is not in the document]
atomb: It sounds like @MikaelMayer is happy to make that adjustment as part of #526, so nothing to do quite yet. :)

Source/Houdini/HoudiniSession.cs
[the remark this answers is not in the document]
atomb: No, I think it's fine. I just wanted to make sure it was intentional (and that there wasn't some other functional change you intended to include).

atomb (approved):

MikaelMayer (commented):
Just a comment, I feel good rebasing my PR on top of yours now that I read it.

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -949,129 +932,122 @@ private static void CleanupRequest(string requestId)
...
-        var cachedResults = Cache.Lookup(impl, Options.RunDiagnosticsOnTimeout, out priority);
-        if (cachedResults != null && priority == Priority.SKIP) {
-          printer.Inform(string.Format("Retrieving cached verification result for implementation {0}...", impl.Name),
-            output);
-          if (Options.VerifySnapshots < 3 ||
-              cachedResults.Outcome == ConditionGeneration.Outcome.Correct) {
-            verificationResult = cachedResults;
-          }
-        }
+      if (0 >= Options.VerifySnapshots)
```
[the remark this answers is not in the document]
cpitclaudel: You don't like Yoda conditions? ^^

Source/ExecutionEngine/CommandLineOptions.cs
```diff
@@ -142,11 +143,11 @@ protected virtual bool ParseOption(string name, CommandLineParseState ps)
...
 
-    ---- General options -------------------------------------------------------
+  ---- General options -------------------------------------------------------
 
-    /version      print the " + ToolName + @" version number
-    /help         print this message
-    /attrHelp     print a message about supported declaration attributes";
+  /version      print the " + ToolName + @" version number
+  /help         print this message
+  /attrHelp     print a message about supported declaration attributes";
```
cpitclaudel, line 150: Why remove the indent here?

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -172,29 +173,21 @@ protected static string CleanUp(string msg)
...
 
-    public static int AutoRequestId(string id)
-    {
-      if (id.StartsWith(AutoRequestIdPrefix))
-      {
-        if (int.TryParse(id.Substring(AutoRequestIdPrefix.Length), out var result))
-        {
+    public static int AutoRequestId(string id) {
+      if (id.StartsWith(AutoRequestIdPrefix)) {
+        if (int.TryParse(id.Substring(AutoRequestIdPrefix.Length), out var result)) {
```
cpitclaudel, line 190: Some of the changes above move curly braces on their own lines; this change does the reverse.  Maybe we should settle on one and apply it consistently?

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -660,6 +654,14 @@ public void Inline(Program program)
...
     }
 
+    [Obsolete("Please inline this method call")]
+    public PipelineOutcome InferAndVerify(
+      Program program,
+      PipelineStatistics stats,
+      string programId = null,
+      ErrorReporterDelegate er = null, string requestId = null) {
+      return InferAndVerify(Console.Out, program, stats, programId, er, requestId).Result;
+    }
```
cpitclaudel, line 664: Feels weird to have two overloads with one of them `async` and the other one not..

keyboardDrummer: One of them is only there for making migration easier. Does that help? Do you want me to remove the non-async one?

Source/ExecutionEngine/CommandLineOptions.cs
keyboardDrummer: Did you comment on old changes? I don't see these changes in this PR

Source/ExecutionEngine/ExecutionEngine.cs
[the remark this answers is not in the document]
MikaelMayer: From left to right, numbers usually increase :-P

Source/ExecutionEngine/ExecutionEngine.cs
MikaelMayer: I think it's fine for now to have the two versions,

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -243,41 +236,42 @@ public ExecutionEngine(ExecutionEngineOptions options, VerificationResultCache c
...
       new ConcurrentDictionary<string, CancellationTokenSource>();
 
-    static ThreadTaskScheduler Scheduler = new ThreadTaskScheduler(16 * 1024 * 1024);
-
-    static TextWriter ModelWriter = null;
+    static ThreadTaskScheduler LargeStackScheduler = new ThreadTaskScheduler(16 * 1024 * 1024);
 
-    public bool ProcessFiles(IList<string> fileNames, bool lookForSnapshots = true, string programId = null)
-    {
+    public bool ProcessFiles(TextWriter output, IList<string> fileNames, bool lookForSnapshots = true,
```
cpitclaudel, line 241: Why not make this function async as well?

keyboardDrummer: I didn't need that. It's not called from Dafny, only from BoogieDriver

Source/ExecutionEngine/ExecutionEngine.cs
keyboardDrummer: > Having both is fine. I expected the async one to be called InferAndVerifyAsync

I've seen that in some places, but putting part of the method signature in the name seems like a duplication of information to me. I'm guessing the main purpose of having Async as a name suffix might be for exactly this scenario, where there's an old and a new API. What do you think?

cpitclaudel: I think it's fine, no need to worry about it too much.

Source/ExecutionEngine/CommandLineOptions.cs
cpitclaudel: Not sure why Github showed them to me.  Thanks!

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -882,15 +827,53 @@ private Implementation[] GetPrioritizedImplementations(Program program)
...
+      }
+
+      await verifyImplementationSemaphore.WaitAsync(cts.Token);
+      try {
+
+        ImplIdToCancellationTokenSource.AddOrUpdate(id, cts, (k, ov) => cts);
+
+        var coreTask = new Task<VerificationResult>(() => VerifyImplementation(program, stats, er, requestId,
+          extractLoopMappingInfo, implementation,
+          programId, taskWriter).Result, cts.Token, TaskCreationOptions.None);
```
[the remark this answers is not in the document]
keyboardDrummer: > Why not directly await VerifyImplementation? Is it because you need a different cancellation token?

Because then you can't use a custom thread scheduler.

> Also, I think you can use [TaskFactory.StartNew](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.taskfactory-1.startnew?view=net-6.0#system-threading-tasks-taskfactory-1-startnew(system-func((system-object-0))-system-object-system-threading-cancellationtoken-system-threading-tasks-taskcreationoptions-system-threading-tasks-taskscheduler)) directly instead of new Task + .Start, which is at least what the documentation recomments.

That looks nicer yes, but note that I'm changing this line again a little bit in https://github.com/boogie-org/boogie/pull/538/files/b0214fc8dd58f62c3acf88c8eb914fc05c26be87..6a9a3a7b4c3d757b1af78cd666952878dec8a028#diff-a5b39418ae8731b4b74a8cb24041d0272c8e4a02316c312cdc2c1bb19dff7d02R857

And I want to change it a little more later to `Task.Run(() => VerifyImplemen...)` once we make sure we don't need the large stack thread scheduler: https://github.com/boogie-org/boogie/pull/524

> I'm not 100% sure what's going on here: you're forcing the task VerifyImplementation to run synchronously by taking its result, but the whole thing is still asynchronous because of the new Task around it?

Yes but it does cause the created thread to wait during IO operations so it's not a great use of resources, that's one thing https://github.com/boogie-org/boogie/pull/524 improves.

Source/ExecutionEngine/ExecutionEngine.cs
keyboardDrummer: I think we're putting the braces on the same line, at least that what my Rider does, I assume based on implicit defaults on the `.editorconfig`. I see one place where this PR moves a brace to a new line, not sure how that happened.

Source/ExecutionEngine/ExecutionEngine.cs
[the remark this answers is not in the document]
cpitclaudel: Oh, I see.  Then I guess this is fine and we can leave future improvements to later.

cpitclaudel (approved):

PR #521, opened 2022-03-03
Fix compatibility issues with solvers other than Z3

### LINKED ISSUE #518: Boogie with CVC5
Does Boogie support the latest version of CVC5 (https://cvc5.github.io/downloads.html)?
I downloaded CVC5 (more specifically, cvc5 version 0.0.7 [git tag 0.0.7 branch HEAD] compiled with GCC version 9.3.0), then ran the tests in the following way, but it did not work:
- `lit -v . --param boogie_params='/proverOpt:SOLVER=CVC5'`  --> no passing test, it seems to be hanging all the time
- `lit -v . --param boogie_params='/proverOpt:PROVER_PATH=XXXXXXX/cvc5/cvc5 # my path to CVC5 binary`  --> mt2 error like below:
```
+Prover error: (error "Error in option parsing: Argument 'mt2' for uint64_t option seed did not parse as uint64_t")

-Boogie program verifier finished with 1 verified, 0 errors
+Prover error:
+
+Prover error: Please use --help to get help on command-line options.
+
+Advisory: MouCreateClassObject SKIPPED due to I/O exception: Broken pipe
+MouCreateClassObject.bpl(3759,12): Verification encountered solver exception (MouCreateClassObject)
+
+Boogie program verifier finished with 0 verified, 0 errors, 1 solver exceptions

error: command failed with exit status: 1
```

This is weird because `mt2` does not seem to appear in the source code of boogie.

I also tried running one single test (sequences/intseq.bpl) for CVC5:
- `lit -v sequences/intseq.bpl --param boogie_params='/proverOpt:SOLVER=CVC5'` It hangs, like above.
- `lit -v sequences/intseq.bpl --param boogie_params='/proverOpt:PROVER_PATH=XXXXXXX/cvc5/cvc5` It raises mt2 error like above.

atomb:
Ensures that Boogie's solver interface, including batch mode, work smoothly with all supported solvers.

Fixes #518.


Source/Provers/SMTLib/SMTLibInteractiveTheoremProver.cs
```diff
@@ -713,7 +712,14 @@ protected override void Send(string s, bool isCommon)
 
     public override async Task<int> GetRCount()
     {
-      SendThisVC("(get-info :rlimit)");
+      if (options.Solver != SolverKind.Z3) {
+        // Only Z3 currently supports retrieving this value. CVC5
+        // supports setting a limit, but does not appear to support
+        // reporting how much it took to complete a query.
```
robin-aws, line 718: That's unfortunate! Can we log an issue for it if it doesn't already exist?

Could we say `if (options.Solver == SolverKind.CVC5)` instead (here and in the other spots you changed), so we default to at least attempting this for other solvers, especially new ones?

Test/prover/cvc5.bpl
```diff
@@ -0,0 +1,8 @@
+// RUN: %boogie /trace /proverOpt:SOLVER=cvc5 "%s" > "%t"
```
robin-aws, line 1: I bet after the last review you know what I'm about to ask here. :)

In this case I'm fine with punting deeper coverage for now, but I'm at least curious what percentage of the suite passes with CVC5.

robin-aws (changes requested):

Source/Provers/SMTLib/SMTLibInteractiveTheoremProver.cs
atomb: Z3 is the only currently-supported solver that has that feature, I think. Yices2 also doesn't support it.

Test/prover/cvc5.bpl
atomb: In the brief attempts I've made, most of the tests time out. So I think it may be too early to try running them all in CI just yet. The changes for included for testing batch mode will make it easy to run a separate job just testing CVC5 once that's expected to complete in a reasonable time, though.

Source/Provers/SMTLib/SMTLibInteractiveTheoremProver.cs
robin-aws: Okay, not the end of the world. If we move towards trying to run all tests on all solvers by default, that will be another mechanism to push us towards hooking up this feature when it is supported.

robin-aws (approved):

Issue #494, opened 2022-02-03
Support verifying a single goal with multiple checkers to measure stability


atomb:
## Use Case

The Dafny project has a goal of improving [verification stability](https://github.com/dafny-lang/dafny/issues/1582), to help ensure that once a program has been verified once it continues to be verifiable as it evolves. One proxy for stability is the variance in resource use between different proof attempts (either using different solvers or using the same solver with different configurations). So, we would like Dafny to be able to automatically provide this variance measurement to users by optionally attempting each proof goal multiple times (each time in a different way) and reporting the results.

We expect to use Z3 resource count as the first variance metric and to vary the proof attempts by varying the Z3 random seed (or, more broadly, the Boogie random seed). In the future, comparing proof attempts by multiple solvers may be a natural next step.

## Implementation Proposal

Given that Boogie is responsible for solver invocation, it is the natural place to implement this functionality. Other clients of Boogie may also benefit from it.

The current `CheckerPool` architecture provides a convenient starting point. To compare the results produced by attempting a proof in several different ways we could:

* Fill the `CheckerPool` with one checker for each solver configuration.
* In `SplitAndVerifyWorker.DoWork`, request multiple checkers from the pool and then perform the check with each.

Two factors will complicate this a bit, however:
* **Aggregating results.** Most of the code farther up the call tree is expecting only one `Outcome`, and arguably _should_ expect only one. I propose a function for combining `Outcome` values that fails if they're inconsistent (e.g., both `Correct` and `Errors`) and a callback for reporting statistics that client code can consume as needed.
* **State and I/O** Some of the code accumulates values to instance fields that would be misleading if accumulated over multiple runs, and some of the code writes to the console or external files during the solving process. I propose that we refactor it to avoid these operations, relying on code farther up the call stack to perform I/O. The statistics callback mentioned above can also replace some of the current instance fields (such as `resourceCount`).

This issue should probably be resolved in careful coordination with [this Dafny issue](https://github.com/dafny-lang/dafny/issues/1647), which also propose some Boogie modifications, some of which will be relevant here, as well.

keyboardDrummer:
"Z3 random seed" -> "Boogie random seed" ? ^^

atomb:
> "Z3 random seed" -> "Boogie random seed" ? ^^

That raises a very interesting question. Does any of the randomization that Boogie does happens before calling `SplitAndVeryifyWorker.DoWork`? If so, we'd need to orchestrate multiple verification attempts elsewhere.

[...]

PR #524, opened 2022-03-07
Enable Boogie programs to be deeply nested without overflowing the stack

keyboardDrummer:
### Changes
- Do not create threads with 16MB instead of the default 1 MB stack space.
- Stop using a TaskScheduler that always creates a new Thread to run a task.
- Remove a deadlock cause in `SMTLibProcess` that was caused by `TaskCompletionSource.SetResult` call inside a lock where the continuation could try to lock the same monitor.
- Stop synchronously blocking on Tasks inside an asynchronous context, which can cause deadlocks
- Wasn't needed: replace algorithms in Boogie that traverse Boogie programs using the call stack

### Testing
- Add a test that tries to trigger a stack overflow by using a Boogie program with deep nesting
- Dafny tests pass with this change: https://github.com/dafny-lang/dafny/pull/1917


cpitclaudel:
Seems reasonable to me, though I'd prefer to get some feedback from other Boogie users.  Why was this introduced in the first place?

keyboardDrummer:
> Seems reasonable to me, though I'd prefer to get some feedback from other Boogie users. Why was this introduced in the first place?

I contacted @BarryBo who contributed the initial change that added this stack, who informed me this was needed by Dafny. I've run the Dafny test suite and one large Dafny customer test suite with this change and both pass so it seems Dafny does not need this change at the moment.

cpitclaudel:
Very nice, thanks.

cpitclaudel (dismissed):

Source/ExecutionEngine/ConcurrentToSequentialWriteManager.cs
```diff
@@ -50,9 +53,10 @@ class SubWriter : WriterWrapper {
     }
 
     public string SetTargetAndGetBuffer(TextWriter newTarget) {
-      var result = buffering ? ((StringWriter)target).ToString() : "";
-      buffering = false;
+      var oldTarget = target;
       target = newTarget;
+      var result = buffering ? ((StringWriter)oldTarget).ToString() : "";
+      buffering = false;
       return result;
```
atomb, line 60: Nice catch. This is a subtle issue!

Source/Provers/SMTLib/SMTLibProcess.cs
```diff
@@ -11,6 +11,8 @@ namespace Microsoft.Boogie.SMTLib
 {
   /*
    * Not thread-safe.
+   * The locks inside this class serve to synchronize between the single external user and the
+   * internal IO threads.
```
atomb, line 15: This is nice to have documented.

Source/VCGeneration/SplitAndVerifyWorker.cs
```diff
@@ -160,6 +165,65 @@ private async Task ProcessResult(Split split, CancellationToken cancellationToke
...
+          {
+            return Outcome.Inconclusive;
+          }
+          return currentOutcome;
+        default:
+          Contract.Assert(false);
+          throw new cce.UnreachableException();
+      }
+    }
+
```
atomb, line 226: Thanks for moving these to separate methods. I think I'll want to use them to resolve #494.

Source/Houdini/StagedHoudini.cs
```diff
@@ -163,8 +163,7 @@ public HoudiniOutcome PerformStagedHoudiniInference()
       foreach (var s in plan)
       {
         Debug.Assert(!plan.GetDependences(s).Contains(s));
-        tasks.Add(new StagedHoudiniTask(s,
-          new Task(o => { ExecuteStage((ScheduledStage) o); }, s, TaskCreationOptions.LongRunning)));
+        tasks.Add(new StagedHoudiniTask(s, Task.Run(() => ExecuteStage(s))));
```
atomb, line 166: Will this new task still have the `LongRunning` option? Does it matter?

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -236,23 +236,26 @@ public class ExecutionEngine : IDisposable {
...
+        foreach (var fileName in fileNames) {
+          success &= await ProcessFiles(output, new List<string> { fileName }, lookForSnapshots, fileName);
+        }
+        return success;
       }
 
       if (0 <= Options.VerifySnapshots && lookForSnapshots) {
         var snapshotsByVersion = LookForSnapshots(fileNames);
-        return snapshotsByVersion.All(s => {
-          // BUG: Reusing checkers during snapshots doesn't work, even though it should. We create a new engine (and thus checker pool) to workaround this.
```
atomb, line 252: Does this bug still exist?

atomb (dismissed):
Nice improvement! I don't think any of my comments necessarily require any changes, but I'll leave that to your judgement.

Source/ExecutionEngine/ConcurrentToSequentialWriteManager.cs
MikaelMayer: Great catch. I'm wondering if it is still possible to have another race condition however: Two threads stepping in this code could both assign `oldTarget` together, and then assign `target` together, and one StringWriter would not be rendered.
I think the whole function should be locked someway.

keyboardDrummer: The calls to this method are inside a lock

atomb (approved):

Source/ExecutionEngine/ConcurrentToSequentialWriteManager.cs
cpitclaudel: I think the "right" way to solve these sort of issues is `Interlocked.Exchange`, but I don't actually understand the root cause, here.

If this method is called within an exclusive lock, how can this change fix an issue?  And if not, why not just lock the whole method body on `myLock`?

keyboardDrummer: The issue was that we had:
```
var result = buffering ? ((StringWriter)target).ToString() : "";
...
target = newTarget;
```

So something could be written to `target` after the `var result` line but before the `target = newTarget` line.

cpitclaudel: Yes but do you have a guarantee that writes to these two variables result and target are done in order?  Otherwise the fix is not guaranteed to help, right?

PR #526, opened 2022-03-08
Realtime verification reporting

MikaelMayer:
This PR enables Boogie to report in real-time its current execution state.
Specifically, it enables the output printers to
* `ReportImplementationsBeforeVerification`: Be notified about the implementations to be verified, for example to change priorities on the fly
* `ReportStartVerifyImplementation`: Be notified when an implementation starts to verify
* `ReportEndVerifyImplementation`: Be notified when an implementation finishes to verify
* `ReportSplitResult`: Be notified when an split finishes
  * There are three places in which I needed to call this: 1) When the prover did not fail, 2) When the prover returns a time out and it was the last change to split the assertions 3) When there is no way it can split the assertions further
  * I changed `VCResult` a bit so that stores the counter-examples obtained from the split itself.
  * I added a method to `VCResult` so that it can compute per-assert outcomes and counter-examples, if any.
* I fixed all the tests on Windows as well (a log file was not closed before being accessed, as a null pointer exception that was throwing because Process was null)
* I added a test for the new behavior of the `OutputPrinter`

** Breaking change** The output printer's interface is changed.


Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -800,16 +809,26 @@ public void Inline(Program program)
 
       // operate on a stable copy, in case it gets updated while we're running
       Implementation[] stablePrioritizedImpls = null;
+      var userDefinedPriorities =
```
MikaelMayer, line 812:
```suggestion
      // Ask the printer for custom priorities to verify implementations
      var userDefinedPriorities =
```

Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
```diff
@@ -645,6 +645,7 @@ public override void FullReset(VCExpressionGenerator gen)
         SendThisVC("(reset)");
         SendThisVC("(set-option :" + Z3.RlimitOption + " 0)");
         HasReset = true;
+        Namer.Reset();
```
MikaelMayer, line 648:
```suggestion
```

Source/ExecutionEngine/ConsolePrinter.cs
```diff
@@ -176,4 +178,25 @@ public virtual void ReportBplError(IToken tok, string message, bool error, TextW
...
+
+  public void ReportEndVerifyImplementation(Implementation implementation, VerificationResult result) {
+    // Do not print anything to console
+  }
+
+  public int GetVerificationPriority(IToken implTok) {
+    return 0;
+  }
+
+  public void ReportAssertionBatchResult(Split split, Dictionary<AssertCmd, ConditionGeneration.Outcome> perAssertOutcome,
```
atomb, line 198: I'm wondering whether `perAssertOutcome` should instead be a `VCResult`, as those are what make up the `VerificationResult` passed into `ReportEndVerifyImplementation`.

Also note that there's an `OnVCResult` method in `VerifierCallback`.

Source/VCGeneration/Counterexample.cs
```diff
@@ -516,17 +517,20 @@ void ObjectInvariant()
     }
 
 
-    public CallCounterexample(VCGenOptions options, List<Block> trace, List<object> augmentedTrace, CallCmd failingCall, Requires failingRequires, Model model,
+    public CallCounterexample(VCGenOptions options, List<Block> trace, List<object> augmentedTrace, AssertRequiresCmd assertRequiresCmd, Model model,
```
atomb, line 520: Nice simplification.

Source/VCGeneration/Split.cs
```diff
@@ -87,6 +87,8 @@ void ObjectInvariant()
       readonly Dictionary<Block /*!*/, BlockStats /*!*/> /*!*/
         stats = new Dictionary<Block /*!*/, BlockStats /*!*/>();
 
+      public IEnumerable<AssertCmd> AssertCmds => blocks.SelectMany(block => block.cmds.OfType<AssertCmd>());
```
atomb, line 90: I think there's already an `Asserts` property containing exactly this.

Source/VCGeneration/SplitAndVerifyWorker.cs
```diff
@@ -151,6 +153,44 @@ private async Task ProcessResult(Split split, CancellationToken cancellationToke
       callback.OnProgress?.Invoke("VCprove", splitNumber < 0 ? 0 : splitNumber, total, provenCost / (remainingCost + provenCost));
 
       if (!proverFailed) {
+        if (callback is VerificationResultCollector collector) {
```
atomb, line 156: All the code in here will be simpler, I think, if you use the signature I suggested above.

The logic you have for identifying which assertions in a batch failed, based on the counterexamples, will still need to happen somewhere, though.

You may need to merge or rebase to incorporate the current `master` before the `VCResult` will be conveniently available.

atomb (changes requested):
This looks great! I'm wondering whether one of the arguments of `ReportAssertionBatchResult` should be different to take into account recent changes, though. I'm not 100% sure that's the right way to go, but I think it's worth at least considering.

Source/VCGeneration/SplitAndVerifyWorker.cs
```diff
@@ -151,6 +153,44 @@ private async Task ProcessResult(Split split, CancellationToken cancellationToke
...
+          List<AssertCmd> asserts = split.AssertCmds.ToList();
+          Dictionary<AssertCmd, Outcome> perAssertOutcome = new();
+          Dictionary<AssertCmd, Counterexample> perAssertCounterExamples = new();
+          if (outcome == Outcome.Correct) {
+            perAssertOutcome = asserts.ToDictionary(cmd => cmd, assertCmd => Outcome.Correct);
+          } else {
+            foreach (var counterExample in collector.examples) {
```
keyboardDrummer, line 163: Why does each Split need to look at counterExamples collected by other splits?

I would think that each split reports a CounterExample if it failed, and the CounterExample can say which statements where involved in the failure, which we can then mark red in the IDE if that CounterExample is active.

Btw, I saw that in the IDE you also mark asserts green. Do you mark them green if the assertion passed for each split in which it occurred ? What do you do if an assertion passes for the first completed split, but fails for the split that completes after it?

Source/VCGeneration/SplitAndVerifyWorker.cs
keyboardDrummer, line 156: Relying on `callback` being a specific subtype seems unreliable.

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -24,6 +24,14 @@ public interface OutputPrinter
     void WriteTrailer(PipelineStatistics stats);
     void WriteErrorInformation(ErrorInformation errorInfo, TextWriter tw, bool skipExecutionTrace = true);
     void ReportBplError(IToken tok, string message, bool error, TextWriter tw, string category = null);
+
+    void ReportImplementationsBeforeVerification(Implementation[] implementations);
+    // For real-time feedback.
```
keyboardDrummer, line 29: I'd leave out the comment.

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -24,6 +24,14 @@ public interface OutputPrinter
...
+    void ReportStartVerifyImplementation(Implementation implementation);
+    void ReportEndVerifyImplementation(Implementation implementation, VerificationResult result);
+    
+    // non-positive for default priority.
+    int GetVerificationPriority(IToken implTok);
```
keyboardDrummer, line 34: This method doesn't match the name of the interface `OutputPrinter`. Why do you need this to begin with? Can't you use the `{:priority}` attribute on Implementations to control their priority?

Source/ExecutionEngine/ConsolePrinter.cs
MikaelMayer: Which of these do you have in mind?
- `Dictionary<AssertCmd, VCResult> perAssertOutcome`
- `VCResult perAssertOutcome`

`The first one would contain more information so it's definitely not a big deal, I can make it happily.
The second one would require to store more information in VCResult, which I can do, but I am wondering first if this is what you meant.`

Source/VCGeneration/Split.cs
MikaelMayer: Oh I see, it's in the newly VCsResult.. Let me check how I can use it there.

Source/VCGeneration/SplitAndVerifyWorker.cs
MikaelMayer: This code is invoked after all the counter-examples of that split have been emitted. However, counter-examples are not recorded per split result (not even in the current master), so I have to rely on the entire list of previously emitted counter-examples, and disregarding counter-examples that were emitted by other splits. Let me know how I can do better, e.g. maybe store the counter-examples in VCResult?
Each split reports up to 5 counter examples.

To my understanding, I think that splits are all disjoints, so if an assert happens in one split, it shouldn't appear in another split. Yet I don't rely on that anyway. If a single assert would appear in two different splits, they will be simply be recorded as two different assertions at the same place. Hence, if one of them fail, the error will be given priority on the gutter - the gutter rendering is like a z-buffer where errors decorations are "closer" than verified decorations.

Source/VCGeneration/SplitAndVerifyWorker.cs
MikaelMayer: Agreed. I think we should put the counter-examples in the VCResult so that I don't need to rely on the collector itself.

Source/ExecutionEngine/ExecutionEngine.cs
MikaelMayer: What do you mean by "doesn't match the name of the interface `OutputPrinter`"?

I need this method so that if a particular output printer is temporarily interested by the result of verifying a specific implementation first, it can set a priority independent of the source code, e.g. for an IDE extension where the user wants to verify the last edited implementation.
You can still use the `{:priority}` attribute on implementations to manually control their priority.

One debatable question is, how to merge a manual and an output-printer given priority. In this PR, I give priority to the output printer - in the IDE extension I'm developing, it will always verify the last five implementations. We could also max the priorities. But I found that, in an IDE setting, it makes it possible to choose the order of verification without writing explicit `{:priority}` attributes, which is pretty neat.

Source/ExecutionEngine/ConsolePrinter.cs
atomb: I'm thinking the latter, which includes all of the assertions in the batch. It only includes one outcome, though, so you'll still need the logic you have (somewhere) to use the counterexamples to figure out which of the assertions in a batch failed when the outcome indicates failure.

Source/VCGeneration/SplitAndVerifyWorker.cs
atomb: I think it may make sense, as you're beginning to suggest here, to adapt `VCResult` to include counterexamples and multiple outcomes (using the logic you implement in this PR). That will require a lot less work from clients of that type in most cases, I expect.

Source/ExecutionEngine/ConsolePrinter.cs
MikaelMayer: I'll try to store the counter-examples in this VCResult so that it makes it easier.

Source/ExecutionEngine/ExecutionEngine.cs
keyboardDrummer: I understand why you added the method, but why don't you use the priority attribute instead?

Source/ExecutionEngine/ConsolePrinter.cs
MikaelMayer: I ended up storing the counter-examples in the VCResult, and the code to figure out per-assertion outcome is now in a method of this VCResult that can be invoked.

Source/VCGeneration/Split.cs
MikaelMayer: Removed in the last commit

Source/ExecutionEngine/ExecutionEngine.cs
MikaelMayer: Aahhhh ok got it got it. I historically discovered the {:priority} attribute while writing this code, but I didn't know at that time I could change the implementation, I was trying to only modify the verification behavior. But now you suggestion makes sense to me. I implemented in my branch something where it sets priorities of implementations in the newly added callback, so that there is no need to order them later.
Everything is working ! thanks for the suggestion.

MikaelMayer:
I think I have a much cleaner solution now.

atomb (dismissed):
Looks great!

MikaelMayer:
> Looks great!

I will wait for at least the approval of @keyboardDrummer to merge this PR.

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -175,6 +179,7 @@ protected static string CleanUp(string msg)
 
   public class ExecutionEngine : IDisposable
   {
+    // Should it be set per request?
```
keyboardDrummer, line 182: It's moved into ExecutionEngineOptions in this PR: https://github.com/boogie-org/boogie/pull/520

Could you remove the comment?

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -780,12 +785,15 @@ private Implementation[] GetPrioritizedImplementations(Program program)
 
       // operate on a stable copy, in case it gets updated while we're running
       Implementation[] stablePrioritizedImpls = null;
+      // Ask the printer for custom priorities to verify implementations
```
keyboardDrummer, line 788: This comment seems obsolete

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -959,6 +968,7 @@ private static void CleanupRequest(string requestId)
       printer.Inform("", output); // newline
       printer.Inform($"Verifying {impl.Name} ...", output);
       var verificationResult = GetCachedVerificationResult(impl, output);
+      printer.ReportStartVerifyImplementation(impl);
```
keyboardDrummer, line 971: Do you also want to report this if a cached result was found? If not, please move this call inside the `if`.

Note that `ReportEndVerifyImplementation` is only called when no cached result is found.

I recommend having the `ReportEndVerifyImplementation` and `ReportStartVerifyImplementation` calls in the same method, so it's easier to see that both are called in the same code path.

Source/VCGeneration/Counterexample.cs
```diff
@@ -508,6 +508,7 @@ public class CallCounterexample : Counterexample
   {
     public CallCmd FailingCall;
     public Requires FailingRequires;
+    public AssertRequiresCmd UnderlyingAssert;
```
keyboardDrummer, line 511: Would `FailingAssert` be a more specific name? I'm not able to get any information out of the term `Underlying`

Source/VCGeneration/Counterexample.cs
```diff
@@ -667,4 +674,53 @@ public virtual void OnVCResult(VCResult result)
...
+
+    public SplitVerifierCallback(VerifierCallback underlying, Split split) : base(CoreOptions.ProverWarnings.None) {
+      this.underlying = underlying;
+      this.split = split;
+    }
+    // reason == null means this is genuine counterexample returned by the prover
+    // other reason means it's time out/memory out/crash
+    public override void OnCounterexample(Counterexample ce, string /*?*/ reason)
+    {
+      split.Counterexamples.Add(ce);
```
keyboardDrummer, line 691: Why don't you do place `split.Counterexamples.Add(ce)` before the call to `OnCounterexample` in `SplitAndVerifyWorker`, then you don't need this class.

Source/VCGeneration/IConditionGenerationLogger.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System.Collections.Generic;
+using Microsoft.Boogie;
+
+namespace VC;
+
+public interface IConditionGenerationLogger {
+  public void ReportSplitResult(Split split,
```
keyboardDrummer, line 7: Newline seems unnecessary

Source/VCGeneration/SplitAndVerifyWorker.cs
```diff
@@ -153,21 +155,26 @@ private async Task ProcessResult(Split split, CancellationToken cancellationToke
...
+        split.parent.Logger?.ReportSplitResult(split, result);
         return;
       }
 
-      await HandleProverFailure(split, cancellationToken);
+      await HandleProverFailure(split, callback, result, cancellationToken);
     }
 
-    private async Task HandleProverFailure(Split split, CancellationToken cancellationToken)
+    private async Task HandleProverFailure(Split split, VerifierCallback callback, VCResult vcResult, CancellationToken cancellationToken)
```
keyboardDrummer, line 165: `callback` variable is shadowing the field variable.

Source/VCGeneration/IConditionGenerationLogger.cs
keyboardDrummer, line 7: Consider renaming this to `LogSplitResult` since it's a method in the a `Logger` interface.

Source/VCGeneration/SplitAndVerifyWorker.cs
```diff
@@ -100,16 +102,17 @@ async Task DoWork(Split split, CancellationToken cancellationToken)
 
       try {
         cancellationToken.ThrowIfCancellationRequested();
-        StartCheck(split, checker, cancellationToken);
+        var splitCallback = new SplitVerifierCallback(callback, split);
```
keyboardDrummer, line 105: Could you move the logic for recording which CounterExamples a Split has into the `Split` class? `Split` creates a `VCGen.ErrorReporter` where this callback is also passed.

Source/ExecutionEngine/ConsolePrinter.cs
```diff
@@ -176,4 +178,24 @@ public virtual void ReportBplError(IToken tok, string message, bool error, TextW
...
+
+  public void ReportStartVerifyImplementation(Implementation implementation) {
+    // Do not print anything to console
+  }
+
+  public void ReportEndVerifyImplementation(Implementation implementation, VerificationResult result) {
+    // Do not print anything to console
+  }
+
+  public int GetVerificationPriority(IToken implTok) {
```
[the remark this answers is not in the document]
MikaelMayer: Oops, sorry, looks like a leftover !

Source/ExecutionEngine/ExecutionEngine.cs
MikaelMayer: Yes we want to report this even if a cached result was found. When we invoke Boogie to verify a procedure, it's an internal detail whether it uses the cache or prove it again.
My extension needs the feedback that verification was started and finished for a given implementation, even if there is no split inside.

Good catch, I moved `ReportEndVerifyImplementation ` in the same method as `ReportStartVerifyImplementation `. You actually just discovered a potential bug which did happen to me many times, when my extension reports that something never finishes to verify but Boogie is done.

Source/VCGeneration/Counterexample.cs
MikaelMayer: Makes sense

Source/VCGeneration/Counterexample.cs
MikaelMayer: You mean, before this call?
`callback.OnCounterexample(split.ToCounterexample(split.Checker.TheoremProver.Context), msg);`
The problem is, `OnCounterExample` is not called only from `SplitAndVerifyWorker`. It's also called from `VCGen` 2 times. Look for `ErrorReporter.OnModel`, it uses the callback defined in `Split.cs` line `reporter = new VCGen.ErrorReporter(...`, which callback is a `SplitVerifyCallback` that records the counter-examples in the split as you have seen.

Source/VCGeneration/IConditionGenerationLogger.cs
MikaelMayer: Renamed instead the interface to `IConditionGenerationPrinter` since it's extended by `OutputPrinter`

Source/VCGeneration/SplitAndVerifyWorker.cs
MikaelMayer: I wish I could move this logic, however, it would require me to also move the logic for handling the split's result into the split itself, since `SplitAndVerifyWorker.HandleProverFailure` and `SplitAndVerifyWorker.ProcessResult` both rely on the `SplitVerifierCallback`. Let me know if you would like me to attempt such a refactoring as part of this PR, or if it is fine for now.

MikaelMayer: Looking at your solution, I was able to find another one that did not even need to change the line you pointed at.
The class SplitVerifierCallback is now removed !

keyboardDrummer (dismissed):

MikaelMayer:
* I fixed all the tests on Windows as well (a log file was not closed before being accessed, as a null pointer exception that was throwing because Process was null)
* I added a test for the new behavior of the `OutputPrinter`

Source/Provers/SMTLib/SMTLibInteractiveTheoremProver.cs
```diff
@@ -29,6 +29,10 @@ public class SMTLibInteractiveTheoremProver : SMTLibProcessTheoremProver
 
     public override Task GoBackToIdle()
     {
+      if (Process == null) {
```
keyboardDrummer, line 32: When does this happen?

Source/VCGeneration/Split.cs
```diff
@@ -1558,5 +1558,10 @@ public void ReleaseChecker()
           checker = null;
         }
       }
-    }
+
+      public void Finish(VCResult result) {
+        Checker.TheoremProver.Close();
```
keyboardDrummer, line 1563: Why would be close here? Shouldn't the theoremProver of the checker be reused?

Source/Provers/SMTLib/SMTLibInteractiveTheoremProver.cs
MikaelMayer: I have no idea but it did happen on my machine and caused a crash as I was debugging the Dafny Language Server and heavily loaded it with a lot of requests.
The only reasons why `Process` would be null would be
- If it `GoBackToIdle()` was called before a single call to `SetupProcess()`
- If the method `Close()` was called before `GoBackToIdle()`

It would be worth investigating why, but for now, I found it should be sound to do this change, because if the process is null, it means it was either not started or it was closed, so it can always be restarted and the prover can be reused.

Source/VCGeneration/Split.cs
MikaelMayer: Good question, let me tell you why this is needed in my case, and so we can discuss alternatives.
When I ran the boogie tests on Windows, it kept crashing 90% of the tests on `GetProverLog.cs` on line 35 because it's trying to read the log files and these logs files were being written by the theorem prover and never closed (`StreamWriter`). So, somewhere I need to close these log files and this was never done explicitely except in the public `Close()` method of the theorem prover which was not call.
Any suggestions to do otherwise?

keyboardDrummer: The trick is to Dispose the ExecutionEngine before reading the log files, like it is done in Dafny [here](https://github.com/dafny-lang/dafny/blob/master/Source/DafnyPipeline.Test/IntraMethodVerificationStability.cs#L217).

MikaelMayer: That's a good trick ! It works !

keyboardDrummer (approved):

MikaelMayer:
I managed to solve all the merging conflicts; however, I cannot push my changes as a few early adopters are relying on both my Dafny branch and Boogie branch to be in sync. @keyboardDrummer please let me know when Dafny will be compatible with the newest version of Boogie.

Source/Provers/SMTLib/SMTLibInteractiveTheoremProver.cs
MikaelMayer: Just added a test that fails by replicating the behavior "close()" before "GoBackToIdle()". It's not the panacea, as in the first place we should never have such a call and there might be some more lurking bugs. But it's better than nothing.

Source/VCGeneration/Split.cs
```diff
@@ -21,9 +20,50 @@ public record VCResult
...
+  ) {
+    public void ComputePerAssertOutcomes(out Dictionary<AssertCmd, ProverInterface.Outcome> perAssertOutcome,
+      out Dictionary<AssertCmd, Counterexample> perAssertCounterExamples) {
+      perAssertOutcome = new();
+      perAssertCounterExamples = new();
+      if (outcome == ProverInterface.Outcome.Valid) {
+        perAssertOutcome = asserts.ToDictionary(cmd => cmd, assertCmd => ProverInterface.Outcome.Valid);
+      } else {
+        foreach (var counterExample in counterExamples) {
+          // Only deal with the ocunter-examples that cover the asserts of this split.
```
keyboardDrummer, line 36: Don't all the counterexamples in `Split.CounterExamples` belong to this Split after your latest changes? Does the continue on line 49 ever trigger?

Source/VCGeneration/Split.cs
```diff
@@ -21,9 +20,50 @@ public record VCResult
...
+
+          if (!asserts.Contains(underlyingAssert)) {
+            continue;
+          }
+
+          perAssertOutcome.Add(underlyingAssert, ProverInterface.Outcome.Invalid);
+          perAssertCounterExamples.Add(underlyingAssert, counterExample);
+        }
+
+        var remainingOutcome =
```
keyboardDrummer, line 56: This part is a bit confusing to me. I get line 58, but line 59 I'm not sure about. Shouldn't that be `ProverInterface.Outcome.Undetermined`?

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -971,19 +978,14 @@ private VerificationResult GetCachedVerificationResult(Implementation impl, Text
...
     private async Task<VerificationResult> VerifyImplementationWithoutCaching(Program program,
       PipelineStatistics stats, ErrorReporterDelegate er, string requestId,
       Dictionary<string, Dictionary<string, Block>> extractLoopMappingInfo,
-      string programId, Implementation impl, TextWriter traceWriter)
+      string programId, Implementation impl, TextWriter traceWriter, IConditionGenerationPrinter printer)
     {
       var verificationResult = new VerificationResult(requestId, impl, programId);
 
-      using var vcgen = new VCGen(program, checkerPool);
+      using var vcgen = new VCGen(program, checkerPool, printer);
```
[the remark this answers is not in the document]
MikaelMayer: On the first sight, `ExecutionEngine.Options` is a `ExecutionEngineOptions`, whereas `CheckerPool.Options` is a `VCGenOptions`, which have nothing in common (and not the printer which is only in `ExecutionEngineOptions`).
Let me know if you still prefer to attach the printer to `CheckerPool.Options`

Source/VCGeneration/Split.cs
MikaelMayer: Normally they should. If we were to write this in Dafny, I would just put some asserts. this is just a guard now in case something went wrong. I'll work on the comment.

Source/VCGeneration/Split.cs
MikaelMayer: You have a point, let me explain the other one.
`ProverInterface.Outcome.Undetermined` was my first choice too originally, but then I figured that if we arrive at this point, then either `outcome != ProverInterface.Outcome.Invalid` or `counterExamples.Count >= maxCounterExamples`
Hence.
1. If `outcome != ProverInterface.Outcome.Invalid`, then because `outcome != ProverInterface.Outcome.Valid` as well (we are on the else branch), the remaining outcomes are either `TimeOut`, `OutOfMemory`, `OutOfResource` or `Undetermined` (I'm not sure what `Bounded` means), which are all more precise than just `Undetermined`, and thus it's nice to report such information at the assertion level.
2. Now if outcome == ProverInterface.Outcome.Invalid, then `counterExamples.Count >= maxCounterExamples`, it means we would assign `ProverInterface.Outcome.Invalid` to the remaining assertions, which is arguably incorrect, so let me change this particular case to `Undetermined` instead.

MikaelMayer:
I added also some missing test portion to at least test the method `vcResult.ComputePerAssertOutcomes`

Source/ExecutionEngine/ExecutionEngine.cs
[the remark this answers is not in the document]
keyboardDrummer: > (and not the printer which is only in ExecutionEngineOptions)

I see. Is it possible to put a `IConditionGenerationPrinter Printer {get;}` in `VCGenOptions` which then is automatically filled with the right printer?

keyboardDrummer (dismissed):

Source/VCGeneration/VCGenOptions.cs
```diff
@@ -6,4 +8,5 @@ public interface VCGenOptions : SMTLibOptions
   bool AlwaysAssumeFreeLoopInvariants { get; }
   int LiveVariableAnalysis { get; }
   bool RemoveEmptyBlocks { get; }
+  IConditionGenerationPrinter Printer {get;}
```
keyboardDrummer: Please resolve the missing space warnings

keyboardDrummer (dismissed):

keyboardDrummer (approved):

PR #528, opened 2022-03-09
Extract methods within ExecutionEngine.VerifyImplementation

keyboardDrummer:
This is almost purely done using Rider's extract method refactoring


MikaelMayer (approved):
Great job !

PR #530, opened 2022-03-10
Support @PROC@ for the -mv option

keyboardDrummer:
### Changes
- Support @PROC@ in the filename argument of the -mv option

### Testing
- In the MoreCapturedStates.bpl test, use @PROC@ in the -mv option together with %parallel-boogie

### DIFF
--- a/Source/Core/Helpers.cs
+++ b/Source/Core/Helpers.cs
@@ -270,8 +270,11 @@ namespace Microsoft.Boogie
     {
       filename = SubstituteAtPROC(descriptiveName, cce.NonNull(filename));
 
-      var reused = allowReuse && UsedLogNames.ContainsKey(filename);
-      var index = UsedLogNames.AddOrUpdate(filename, 0, (_, i) => allowReuse ? i : i + 1);
+      var reused = false;
+      var index = UsedLogNames.AddOrUpdate(filename, 0, (_, i) => {
+        reused = allowReuse;
+        return allowReuse ? i : i + 1;
+      });
       var filenameWithIndex = index > 0 ? filename + "." + index : filename;
 
       return (filenameWithIndex, reused);
--- a/Source/Core/Helpers.cs
+++ b/Source/Core/Helpers.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics.Contracts;
 
@@ -264,30 +265,18 @@ namespace Microsoft.Boogie
       }
     }
 
-    private static readonly HashSet<string> UsedLogNames = new();
+    private static readonly ConcurrentDictionary<string, int> UsedLogNames = new();
     public static (string fileName, bool reused) GetLogFilename(string descriptiveName, string filename, bool allowReuse)
     {
       filename = SubstituteAtProc(descriptiveName, cce.NonNull(filename));
-      var curFilename = filename;
-      var reused = false;
 
-      lock (UsedLogNames) {
-        int n = 1;
-        while (!allowReuse && UsedLogNames.Contains(curFilename)) {
-          curFilename = filename + "." + n++;
-        }
+      var reused = allowReuse && UsedLogNames.ContainsKey(filename);
+      var index = UsedLogNames.AddOrUpdate(filename, 0, (_, i) => allowReuse ? i : i + 1);
+      var filenameWithIndex = index > 0 ? filename + "." + index : filename;
 
-        if (UsedLogNames.Contains(curFilename)) {
-          reused = true;
-        }
-
-        UsedLogNames.Add(curFilename);
-      }
-
-      return (curFilename, reused);
+      return (filenameWithIndex, reused);
     }
 
-    // Substitute @PROC@ in a filename with the given descName
     private static string SubstituteAtProc(string descriptiveName, string filename)
     {
       Contract.Requires(filename != null);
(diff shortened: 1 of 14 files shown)


Source/VCGeneration/Split.cs
```diff
@@ -127,7 +127,7 @@ void ObjectInvariant()
 
       // async interface
       private Checker checker;
-      private int splitNum;
+      private int splitIndex;
```
cpitclaudel, line 130: What's wrong with `splitNum`?

Source/Core/Helpers.cs
```diff
@@ -263,18 +264,39 @@ public static void ExtraTraceInformation(CoreOptions options, string point)
       }
     }
 
+    private static readonly HashSet<string> UsedLogNames = new();
```
cpitclaudel, line 267: I think you can simplify the code below a lot by using a concurrent dictionary from file names to indices here: this way you can replace the while loop and the lock with a single call to `AddOrUpdate`.

Source/Core/Helpers.cs
```diff
@@ -263,18 +264,39 @@ public static void ExtraTraceInformation(CoreOptions options, string point)
...
+
+        UsedLogNames.Add(curFilename);
+      }
+
+      return (curFilename, reused);
+    }
+
     // Substitute @PROC@ in a filename with the given descName
-    public static string SubstituteAtPROC(string descName, string fileName)
+    private static string SubstituteAtProc(string descriptiveName, string filename)
```
cpitclaudel, line 291: I wouldn't change the name of this method; it substitutes `@PROC`, not `@Proc`

Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
```diff
@@ -524,22 +524,10 @@ protected TextWriter OpenOutputFile(string descriptiveName)
...
-        while (usedLogNames.Contains(curFilename))
-        {
-          curFilename = filename + "." + n++;
-        }
-
-        usedLogNames.Add(curFilename);
-      }
-
-      return new StreamWriter(curFilename, false);
+      return new StreamWriter(filename, reused);
```
cpitclaudel, line 530: Are there going to be concurrency issues here with multiple streamwriters to a single file?

Source/VCGeneration/Counterexample.cs
```diff
@@ -88,7 +89,8 @@ void ObjectInvariant()
 
     public Dictionary<TraceLocation, CalleeCounterexampleInfo> calleeCounterexamples;
 
-    internal Counterexample(VCGenOptions options, List<Block> trace, List<object> augmentedTrace, Model model, VC.ModelViewInfo mvInfo, ProverContext context)
+    internal Counterexample(VCGenOptions options, List<Block> trace, List<object> augmentedTrace, Model model,
+      VC.ModelViewInfo mvInfo, ProverContext context, ProofRun split)
```
cpitclaudel, line 93: Weird combination of type and variable name.

cpitclaudel (dismissed):

Source/VCGeneration/Split.cs
keyboardDrummer: This is subjective, but to me `splitIndex` makes it more clear that this is the index that the split has in an ordered list of splits.

Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
keyboardDrummer: The previous line passes `allowReuse: false` so `reused` is always false.

Source/Core/Helpers.cs
keyboardDrummer: Rider warns about any names with consecutive capitals, but I'm not sure why so I'll undo this.

Source/Core/Helpers.cs
```diff
@@ -264,31 +265,19 @@ public static void ExtraTraceInformation(CoreOptions options, string point)
...
-        }
-
-        if (UsedLogNames.Contains(curFilename)) {
-          reused = true;
-        }
-
-        UsedLogNames.Add(curFilename);
-      }
+      var reused = allowReuse && UsedLogNames.ContainsKey(filename);
+      var index = UsedLogNames.AddOrUpdate(filename, 0, (_, i) => allowReuse ? i : i + 1);
```
cpitclaudel, line 274: This code is thread safe because the race between `ContainsKey` and `AddOrUpdate` only happens in the nonconcurrent case (`allowReuse`), but maybe this will be confusing to readers?  I *think* `reused` is the same as `allowReuse && index > 0`?

cpitclaudel: Ah, no, sorry, I misread the code.

cpitclaudel: I guess you could leave things as is, or maybe use this:

```
      var index = UsedLogNames.AddOrUpdate(filename, 0, (_, i) => allowReuse ? -1 : i + 1);
      var reused = index == -1
```

Or alternatively use this:

```
if (allowReuse) {
  return filename, UsedLogNames.ContainsKey(filename);
}

… rest of the code with reused = false
```

I'm assuming you can't mix calls with and without allowReuse in a single process?

cpitclaudel (approved):
LGTM, thanks.  I wonder if this added terminology of "ProofRun" will be confusing i addition to "Split" but I think it's minor

cpitclaudel (approved):

Source/Core/Helpers.cs
cpitclaudel: New version is perfect :)

PR #531, opened 2022-03-10
Maintain order of logging that results from use the -traceCaching option when using parallel Boogie

keyboardDrummer:
### Changes
- Maintain order of logging that results from use the `-traceCaching` option when using parallel Boogie
- Replace `OutputCollector` with `ConcurrentToSequentialWriteManager` that allows writing to the final TextWriter immediately when child TextWriters are written to, instead of only after calls to `WriteMoreOutput`.

### Testing
- Use `%parallel-boogie` in `runtest.snapshot` which already uses `-traceCaching`
- Add a test for `ConcurrentToSequentialWriteManager`
- Tweak `test0/SplitOnEveryAssert.bpl` and `commandline/SplitOnEveryAssert.bpl` because the output order has improved slightly


Source/ExecutionEngine/ConcurrentToSequentialWriteManager.cs
```diff
@@ -0,0 +1,60 @@
+using System.Collections.Generic;
+using System.IO;
+
+namespace Microsoft.Boogie;
+
+public class ConcurrentToSequentialWriteManager
```
atomb, line 6: I think it might be helpful to include a comment describing in some detail what this class is supposed to do.

I found reading the test to be the easiest way to get an intuition for it. Given that intuition, I can now see how this class implements it. But before wrapping my head around that, I wasn't sure what the purpose of the code was.

atomb (changes requested):

atomb (approved):
Nice solution!

PR #532, opened 2022-03-13
[Civl] Fixed bug in the injection of preconditions for yielding procedures

shazqadeer:


Source/Concurrency/YieldingProcInstrumentation.cs
```diff
@@ -250,31 +250,34 @@ private List<Cmd> InlineYieldLoopInvariants(List<CallCmd> yieldInvariants)
       foreach (var impl in absyMap.Keys.OfType<Implementation>())
       {
         var initCmds = new List<Cmd>();
+
+        // Global variables must be havoced to model the yield upon entering a yielding procedure
         if (civlTypeChecker.GlobalVariables.Count() > 0)
```
shazqadeer, line 255: This diff is silly.  All I have done is to guard only the havocing of globals with test while adding the other preconditions unconditionally. The change in indentation might have confused the diff tool.

bkragl (approved):

PR #533, opened 2022-03-13
[Civl] Generalizing PingPong

shazqadeer:
This PR generalizes PingPong to make it more modular so that the code does not depend on global constants and it becomes possible to create any number of PingPong pairs. The revision is implemented atop the abstraction of bidirectional channels. Implementing bidirectional channels atop ordinary channels is a topic of a separate PR.

A side note: it is becoming increasingly problematic to develop Civl programs in the absence of modules for reusing common libraries.

### DIFF
--- a/Test/civl/inductive-sequentialization/PingPong.bpl
+++ b/Test/civl/inductive-sequentialization/PingPong.bpl
@@ -1,32 +1,41 @@
 // RUN: %parallel-boogie "%s" > "%t"
 // RUN: %diff "%s.expect" "%t"
 
-// This example shows how to use a bidirectional shared channel to communicate between two processes.
+// This example shows how to use a bidirectional shared channel to communicate between
+// two processes. The modeling of bidirectional channels is generic.
+// Its usage is specifically illustrated here on a PingPong example.
 
 // A bidirectional channel is a pair of ordinary channels with two ends---left and right.
 type {:datatype} ChannelPair;
 function {:constructor} ChannelPair(left: [int]int, right: [int]int): ChannelPair;
 
+// The id type for indexing into the pool of bidirectional channels.
 type ChannelId;
+
+// The following global variables models al instances of a bidirectional channel indexed
+// the ChannelId type. A single instance of PingPong will only use a single channel id.
 var {:layer 0,3} channel: [ChannelId]ChannelPair;
 
 // The id of a bidirectional channel can be split into two permissions---Left and Right.
 // Left permission is used to receive from the left channel and send to the right channel.
 // Right permission is used to receive from the right channel and send to the left channel.
-type {:linear "cid"} {:datatype} Permission;
-function {:constructor} Left(cid: ChannelId): Permission;
-function {:constructor} Right(cid: ChannelId): Permission;
-function {:inline} ChannelId(p: Permission) : ChannelId {
+type {:linear "cid"} {:datatype} ChannelHandle;
+function {:constructor} Left(cid: ChannelId): ChannelHandle;
+function {:constructor} Right(cid: ChannelId): ChannelHandle;
+function {:inline} ChannelId(p: ChannelHandle) : ChannelId {
   if is#Left(p) then cid#Left(p) else cid#Right(p)
 }
 
-function {:inline} {:linear "cid"} ChannelIdCollector(cid: ChannelId) : [Permission]bool {
+function {:inline} {:linear "cid"} ChannelIdCollector(cid: ChannelId) : [ChannelHandle]bool {
   MapConst(false)[Left(cid) := true][Right(cid) := true]
 }
 
+// This datatype declares the pending asyncs for Ping and Pong processes.
+// These two processes share a channel pair with Ping holding its left channel handle
+// and Pong holding its right channel handle.
 type {:pending_async}{:datatype} PA;
-function {:constructor} PING(x: int, left: Permission): PA;
-function {:constructor} PONG(x: int, right: Permission): PA;
+function {:constructor} PING(x: int, left: ChannelHandle): PA;
+function {:constructor} PONG(x: int, right: ChannelHandle): PA;
 
 function {:inline} NoPAs () : [PA]int
 { (lambda pa:PA :: 0) }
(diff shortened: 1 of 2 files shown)


shazqadeer:
@bkragl:  Ping ...

Test/civl/inductive-sequentialization/PingPong.bpl
```diff
@@ -1,16 +1,32 @@
...
+
+// The id of a bidirectional channel can be split into two permissions---Left and Right.
+// Left permission is used to receive from the left channel and send to the right channel.
+// Right permission is used to receive from the right channel and send to the left channel.
+type {:linear "cid"} {:datatype} Permission;
+function {:constructor} Left(cid: ChannelId): Permission;
+function {:constructor} Right(cid: ChannelId): Permission;
+function {:inline} ChannelId(p: Permission) : ChannelId {
+  if is#Left(p) then cid#Left(p) else cid#Right(p)
+}
```
bkragl, line 21: I'm wondering if renaming `Permission` to something like `ChannelHandle` makes sense.

In any case, I think there should be an explanation here or at the definitions of `SEND`/`RECEIVE` (or both) that the name of a permission/handle comes from the receiving end. I.e., with a `Left` we receive from the left channel and send to the right, and with a `Right` we recieve from the right channel and send to the left.

Test/civl/inductive-sequentialization/PingPong.bpl
```diff
@@ -1,16 +1,32 @@
...
-const unique pong_id:int;
+function {:inline} {:linear "cid"} ChannelIdCollector(cid: ChannelId) : [Permission]bool {
+  MapConst(false)[Left(cid) := true][Right(cid) := true]
+}
 
 type {:pending_async}{:datatype} PA;
-function {:constructor} PING(x:int, pid:int) : PA;
-function {:constructor} PONG(x:int, pid:int) : PA;
+function {:constructor} PING(x: int, left: Permission): PA;
+function {:constructor} PONG(x: int, right: Permission): PA;
```
bkragl, line 29: Maybe instead of `left` and `right` we should use the same generic name for both. Like `permission` or `channel_handle`. Same in the corresponding action definitions.

On the one hand the names provide hints about the intended values, but on the other hand they could be misleading to assume something about the values before we prove it (like `assert is#Left(left);` in `PING`).

Test/civl/inductive-sequentialization/PingPong.bpl
```diff
@@ -21,213 +37,254 @@ function {:inline} EmptyChannel () : [int]int
...
 {
-  assert (exists {:pool "INV"} m:int :: ping_channel[m] > 0) && (forall m:int :: pong_channel[m] == 0);
-  call PAs := PING(x, pid);
+  var cid: ChannelId;
+  var left_channel: [int]int;
+  var right_channel: [int]int;
+
+  cid := ChannelId(left);
+  left_channel := left#ChannelPair(channel[cid]);
+  right_channel := right#ChannelPair(channel[cid]);
```
bkragl, line 88: This is done in multiple places. Would be nice to extract it. Unfortunately, as far is I know a function cannot return multiple values. An auxiliary atomic action would work, but it must take `channel` as input value instead of accessing the global variable in order to not break commutativity.

Also, in this context we could call the channels `ping_channel` and `pong_channel`.

Test/civl/inductive-sequentialization/PingPong.bpl
shazqadeer: I have renamed Permission to ChannelHandle.  

I didn't understand what you meant by the extra comment. The existing comment seems to be doing what you want.

Test/civl/inductive-sequentialization/PingPong.bpl
shazqadeer: I left the field names as they are.  But I added a comment.

Test/civl/inductive-sequentialization/PingPong.bpl
shazqadeer: Your comment about the extracted atomic action interfering with commutativity is interesting. Now that atomic actions are allowed to call each other, users may create many auxiliary helper atomic actions which may be called only from other atomic actions but never directly from a procedure. I am wondering if we should use an analysis inside Civl to remove such actions from the pool over which commutativity checking performed.

shazqadeer: Indeed, extracting into an atomic action would help

```
var cid: ChannelId;
  var left_channel: [int]int;
  var right_channel: [int]int;

  cid := ChannelId(right);
  left_channel := left#ChannelPair(channel[cid]);
  right_channel := right#ChannelPair(channel[cid]);
```
The above six lines would become a single line of code. But to get the full effect, we would need the analysis I mentioned in my last comment.

I thought about renaming left_channel to ping_channel (and right_channel to pong_channel). It didn't seem to help my understanding so I am not doing it. You are welcome to make changes as you please once this PR lands.

shazqadeer:
@bkragl : I am ready to land the PR since it looks like I have addressed the changes you desired. The suggestion about extracting an atomic action is indeed good but perhaps a separate PR is needed to address it.

Test/civl/inductive-sequentialization/PingPong.bpl
bkragl: > I didn't understand what you meant by the extra comment. The existing comment seems to be doing what you want.

Oh right, sorry.

Test/civl/inductive-sequentialization/PingPong.bpl
bkragl: > The above six lines would become a single line of code.

We would still have the declarations of the three local variables, but only one assignment.

> But to get the full effect, we would need the analysis I mentioned in my last comment.

Yes, having some mechanism to exclude helper actions from commutativity checking sounds useful. I'm wondering though if a more transparent solution (e.g., where the programmer explicitly marks such actions) would be better. Otherwise code becomes hard to read, since the programmer needs to keep in mind which actions (and potential bogus mover types) get ignored.

Either way, we'll do this separately from this PR.

bkragl (approved):

PR #537, opened 2022-03-15
Remove project Predication

keyboardDrummer:
This project is not used by Boogie, [Corral](https://github.com/boogie-org/corral) or Dafny


bkragl:
I think the predication algorithm and uniformity analysis in this project were written for [GPUVerify](https://github.com/mc-imperial/gpuverify). Looks like the project has not been active recently, and the [boogie binaries](https://github.com/mc-imperial/gpuverify/tree/master/BoogieBinaries) in the repo are a few years old.

@shazqadeer: Since you were part of the GPUVerify team, do you think this code should be preserved?

shazqadeer:
@keyboardDrummer : The motivation of this PR is written as if Corral and Dafny are the only users of Boogie infrastructure.  This is certainly not true.

I don't know how active GPUVerify is.  @afd ?

keyboardDrummer:
> @keyboardDrummer : The motivation of this PR is written as if Corral and Dafny are the only users of Boogie infrastructure. This is certainly not true.

Sorry, that was not my intention. I opened this PR as a way of asking whether this project is still used.

afd:
@jeroenk has been more active on maintaining GPUVerify than I have recently. While GPUVerify is now a "quiet" project, some folks are actively contributing to it right now, so we do want to make sure it does not rot.

@jeroenk: while the binaries that ship with GPUVerify are rather old, am I right that the "build from source" instructions do use the latest Boogie?

All: the reason we contributed predication to Boogie, rather than keeping it in GPUVerify, is that we wondered if other projects might find it useful (it's a potentially cool way of flattening control flow).

As that clearly hasn't happened, there could indeed be a case for moving it into the GPUVerify source base, to simplify and clean up the Boogie code base. I'd be open to that, but we'd need a bit of time to do it.

Thanks!

keyboardDrummer:
> All: the reason we contributed predication to Boogie, rather than keeping it in GPUVerify, is that we wondered if other projects might find it useful (it's a potentially cool way of flattening control flow).

That sounds good to me. Can we add a `README.md` to Predication to explain when to use it?

afd:
Yes - good idea. I have opened #543 to track this (can't do it immediately due to other commitments).

jeroenk:
@afd bit late to the party, but:

> while the binaries that ship with GPUVerify are rather old, am I right that the "build from source" instructions do use the latest Boogie?

Not quite we have a binary version of Boogie in the GPUVerify repo, which we update using a separate set of instructions.

We also had discussion before about it might being better for predication to live in the GPUVerify repo. Those discussions were never followed up on. I wouldn't be against moving it over. The important question then is whether we have access to all the Boogie interfaces that predication depends upon.

keyboardDrummer:
> The important question then is whether we have access to all the Boogie interfaces that predication depends upon.

I think you would because Predication in Boogie is already in a separate .NET project/assembly, so it already only has access to Boogie APIs that are public.

PR #538, opened 2022-03-15
Concurrency fixes

keyboardDrummer:
### Changes
Several changes to make using Boogie concurrently either safer or faster.

- Make `AbsyQuant.cs.EqualWithAttributesAndTriggers` thread-safe
- Remove locks from `CheckerPool`
- Remove two calls that waited for a `Task` to complete, one in `ExecutionEngine` and one in `SMTLibInteractiveTheoremProver.cs`

### Testing
Changes should be covered by regression suite. Most Boogie tests are run parallelly.

### DIFF
--- a/Source/ExecutionEngine/ExecutionEngine.cs
+++ b/Source/ExecutionEngine/ExecutionEngine.cs
@@ -804,23 +804,11 @@ namespace Microsoft.Boogie
         foreach (var task in tasks) {
           task.Result.ProcessXml(this);
         }
-      }
-      catch (AggregateException ae) {
-        ae.Flatten().Handle(e =>
-        {
-          if (e is ProverException) {
-            Options.Printer.ErrorWriteLine(Console.Out, "Fatal Error: ProverException: {0}", e.Message);
-            outcome = PipelineOutcome.FatalError;
-            return true;
-          }
-
-          if (e is OperationCanceledException) {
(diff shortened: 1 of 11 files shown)


Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -804,23 +804,11 @@ private Implementation[] GetPrioritizedImplementations(Program program)
...
-            return true;
-          }
-
-          return false;
-        });
+      } catch(TaskCanceledException) {
+        outcome = PipelineOutcome.Cancelled;
+      } catch(ProverException e) {
+        Options.Printer.ErrorWriteLine(Console.Out, "Fatal Error: ProverException: {0}", e.Message);
+        outcome = PipelineOutcome.FatalError;
```
atomb, line 811: Do these exceptions no longer occur in `AggregateException` bundles?

atomb (approved):
I think this looks good, assuming you're confident that there's no longer a need to catch exceptions wrapped in an `AggregateException` at the point where I commented.

Source/ExecutionEngine/ExecutionEngine.cs
keyboardDrummer: I think so. An AggregateException can be created when synchronously blocking on a Task, like when using `.Result`, of which two have been removed in this PR. Without this change a test failed.

Issue #546, opened 2022-03-20
/vcsLoad no longer parses non-integers


tchajed:
In Dafny 3.5.0, `/vcsLoad:0.5` now fails with `Error: Invalid argument "0.5" to option /vcsLoad`. This seems to be due to a regression in Boogie.

I don't have the setup to run Boogie tests directly, but looking at the code I think this call is suspicious:
https://github.com/boogie-org/boogie/blob/b090548f4c5e2b444b61884fb5b17f3fadf0a12c/Source/ExecutionEngine/CommandLineOptions.cs#L1422-L1424

There's no variant of `GetNumericArgument` that takes an `Action<double>`, so I don't know how this type-checks (is C# unsoundly casting it to `Action<int>`?). From the current API of CommandLineParseState I think this is the version to use:

https://github.com/boogie-org/boogie/blob/b090548f4c5e2b444b61884fb5b17f3fadf0a12c/Source/ExecutionEngine/CommandLineParseState.cs#L308-L312

cpitclaudel:
> There's no variant of `GetNumericArgument` that takes an `Action<double>`, so I don't know how this type-checks (is C# unsoundly casting it to `Action<int>`?).

`x: int => load = x` is a well-typed `Action<int>`; the cast happens when assigning to `load`.  The bug was introduced in 7867a2ef as a side effect of removing `ref` parameters (cc @keyboardDrummer):

```diff
         case "vcsLoad":
           double load = 0.0;
-          if (ps.GetNumericArgument(ref load))
+          if (ps.GetNumericArgument(x => load = x))
           {
```

tchajed:
The Dafny part makes sense. Not sure if you want to go back to `ref load` to the option parsing code or add an `Action<double>` variant to `GetNumericArgument`.

I shouldn't be surprised that the interaction of specialized type implementations, type inference, and capturing a reference would lead to confusing behaviors.

keyboardDrummer:
Thanks for reporting. I will look at fixing this. Seems like avoiding overloaded method names in this instance will help prevent bugs like this in the future.

PR #547, opened 2022-03-21
Fix incorrect parsing of double options

keyboardDrummer:
Fixes https://github.com/boogie-org/boogie/issues/546

### Testing

I decided not to add tests for this regression. I think in the future we'll refactor options so there's no more duplication in handling which type an option is then the single type specification can serve as both specification and implementation.

### DIFF
--- a/Source/Directory.Build.props
+++ b/Source/Directory.Build.props
@@ -2,7 +2,7 @@
 
   <!-- Target framework and package configuration -->
   <PropertyGroup>
-    <Version>2.13.1</Version>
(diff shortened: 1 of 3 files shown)


tchajed:
You could still have a single method like `GetNumericArgument` and just call it as `GetNumericArgument<int>` or `GetNumericArgument<double>` to distinguish the two. (Just a thought, I'm not a C# programmer and don't know which style is better.)

cpitclaudel (approved):
Works for me.  Why are `Action`s better than ref parameters though?

PR #548, opened 2022-03-22
Allow related messages from loop invariants

atomb:
The original PR to create structured descriptions of proof obligations (#505) broke the "Related message" functionality for loop invariants. This restores it.

Needed by https://github.com/dafny-lang/dafny/pull/1915.


MikaelMayer (changes requested):
I hope these comments are straightforward! It's good to see the error reporting system get better.

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -1537,34 +1537,54 @@ private ErrorInformation CreateErrorInformation(Counterexample error, VC.VCGen.O
...
-            assertError.FailingAssert.Description.FailureDescription,
+          errorInfo = ErrorInformationFactory.CreateErrorInformation(failingAssert.tok,
+            failingAssert.Description.FailureDescription,
             assertError.RequestId, assertError.OriginalRequestId, cause);
-          errorInfo.Kind = assertError.FailingAssert is LoopInitAssertCmd ?
+          errorInfo.Kind = failingAssert is LoopInitAssertCmd ?
             ErrorKind.InvariantEntry : ErrorKind.InvariantMaintainance;
-          if ((assertError.FailingAssert.ErrorData as string) != null)
+          string relatedMessage = null;
+          if ((failingAssert.ErrorData as string) != null)
```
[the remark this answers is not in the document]
atomb: Good suggestion.

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -1537,34 +1537,54 @@ private ErrorInformation CreateErrorInformation(Counterexample error, VC.VCGen.O
...
           {
-            errorInfo.AddAuxInfo(assertError.FailingAssert.tok, assertError.FailingAssert.ErrorData as string,
-              "Related message");
+            relatedMessage = failingAssert.ErrorData as string;
+          }
+          else if (failingAssert is LoopInitAssertCmd initCmd)
```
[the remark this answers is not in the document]
atomb: Because of the way the `Counterexample` type is structured, with the failing command retrieved using differently-named properties depending on what type of counterexample it is, reducing duplication is more complicated than it should be. It's a refactoring I'd like to do, but since this PR is blocking https://github.com/dafny-lang/dafny/pull/1915, I'd prefer to do it as part of a separate, purely refactoring PR.

MikaelMayer (approved):
Looks good to me! Thanks!

Source/Core/AbsyCmd.cs
```diff
@@ -3898,9 +3898,12 @@ public override Absy StdDispatch(StandardVisitor visitor)
   // An AssertCmd that is a loop invariant check before the loop iteration starts
   public class LoopInitAssertCmd : AssertCmd
   {
-    public LoopInitAssertCmd(IToken /*!*/ tok, Expr /*!*/ expr)
+    public readonly AssertCmd originalAssert;
```
keyboardDrummer, line 3901: What's an original assertion in the context of a loop invariant?

keyboardDrummer (commented):
Could you add a regression test? I have trouble reviewing the code because I don't know what it's supposed to do.

atomb:
> Could you add a regression test? I have trouble reviewing the code because I don't know what it's supposed to do.

There are quite a few regression tests in Dafny.  A lit test in Boogie unfortunately wouldn't be able to trigger this behavior, because you can't add descriptions in Boogie source. I suppose it would be possible to write a C# test of it, but it's quite a bit easier to test it from a client such as Dafny.

keyboardDrummer:
> I suppose it would be possible to write a C# test of it

What's difficult about that?

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -1537,34 +1537,54 @@ private ErrorInformation CreateErrorInformation(Counterexample error, VC.VCGen.O
...
+          if (failingAssert.ErrorData is string)
           {
-            errorInfo.AddAuxInfo(assertError.FailingAssert.tok, assertError.FailingAssert.ErrorData as string,
-              "Related message");
+            relatedMessage = failingAssert.ErrorData as string;
+          }
+          else if (failingAssert is LoopInitAssertCmd initCmd)
+          {
+            var desc = initCmd.originalAssert.Description;
+            if (desc.ShortDescription != "assert")
```
keyboardDrummer, line 1556: Why don't you move this check to where the `LoopInitAssertCmd` is created and then immediately replace its description with the description of the original?

Also, checking against a hardcoded string seems frail. Don't you want to replace "assert" with `AssertionDescription.ShortDescription` ?

atomb: My thinking is that there may wind up being other aspects of the original assertion that it's useful to access in the future.

And, yeah, checking against AssertionDescription.ShortDescription is probably the right thing. I'll change that.

atomb: But a type test does work.

atomb:
> > I suppose it would be possible to write a C# test of it
> 
> Why not do that? You seem to imply it's difficult but why is that?

I'm looking into it to see how hard it really is.

Source/ExecutionEngine/ExecutionEngine.cs
keyboardDrummer: My suggestion was to make AssertionDescription a singleton, but a type test works too.

> My thinking is that there may wind up being other aspects of the original assertion that it's useful to access in the future.

Why not make that change then?

atomb:
> > > I suppose it would be possible to write a C# test of it
> > 
> > 
> > Why not do that? You seem to imply it's difficult but why is that?
> 
> I'm looking into it to see how hard it really is.

And it turns out not to be so hard. Added! Thanks for pushing for it.

Source/UnitTests/ExecutionEngineTests/ExecutionEngineTest.cs
```diff
@@ -51,4 +61,54 @@ procedure Good(y: int)
...
+
+  [Test]
+  public async Task LoopInvariantDescriptions()
+  {
+    var options = CommandLineOptions.FromArguments();
+    var engine = ExecutionEngine.CreateWithoutSharedCache(options);
+
+    var writer = new StringWriter();
+    var concurrentWriterManager = new ConcurrentToSequentialWriteManager(writer);
+    var taskWriter = concurrentWriterManager.AppendWriter();
```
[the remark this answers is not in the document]
atomb: Right you are. Pushed.

Source/ExecutionEngine/ExecutionEngine.cs
atomb: I'm not sure I know what you mean by "that change".

Although `AssertionDescription` itself could be a singleton, other subclasses of `ProofObligationDescription` can't be.

keyboardDrummer: "That change" would be adding the "originalAssert" field, as opposed to only using the Description field.

atomb: > > Although AssertionDescription itself could be a singleton
> 
> That would be enough right?
> 
> > , other subclasses of ProofObligationDescription can't be.
> 
> Why not? None of them take parameters or are mutable right?

The subclasses in Boogie don't take parameters, but most of the ones [in Dafny](https://github.com/dafny-lang/dafny/pull/1915) do.

keyboardDrummer (approved):

Source/ExecutionEngine/ExecutionEngine.cs
atomb: > Have you considered moving this check to where the `LoopInitAssertCmd` is created and then immediately replacing its description with the description of the original? That way you avoid having to introduce the `originalAssert` field.

This is unfortunately not directly possible, since the current logic uses _both_ descriptions: the one associated with the `LoopInitAssertCmd` and the one associated with the `AssertCmd` it was derived from.

PR #549, opened 2022-03-23
Use static CVC5 version

keyboardDrummer:
Depending on a static version instead of `latest` prevents us from breaking because of changes in CVC5.

One such breakage occurred today when a file inside latest changed its name from to `cvc5-Linux` to `cvc5-Linux-2022-03-23-7c895b7`


MikaelMayer (approved):
Makes totally sense.

PR #551, opened 2022-03-23
Don't allocate a checker if the checker waiter is canceled.

MikaelMayer:
I'm wondering if this can help.
While debugging, I have seen a lot of checkers being allocated and prepared to just be thrown away because the task awaiting it has been canceled.
This PR does not even allocate the tasks.

However, it contains something I would like someone (e.g. @cpitclaudel ) to confirm it is not a problem: If we use WaitASync(cancellationToken), can it cause a deadlock in any way?

### DIFF
--- a/Source/VCGeneration/CheckerPool.cs
+++ b/Source/VCGeneration/CheckerPool.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics.Contracts;
+using System.Net;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Boogie;
@@ -11,7 +12,7 @@ namespace VC
   public class CheckerPool
   {
     private readonly ConcurrentStack<Checker> availableCheckers = new();
-    private readonly ConcurrentQueue<TaskCompletionSource<Checker>> checkerWaiters = new();
+    private readonly ConcurrentQueue<(TaskCompletionSource<Checker>, CancellationToken?)> checkerWaiters = new();
     private int notCreatedCheckers;
     private bool disposed;
 
@@ -23,7 +24,7 @@ namespace VC
       notCreatedCheckers = options.VcsCores;
     }
 
-    public Task<Checker> FindCheckerFor(ConditionGeneration vcgen, Split split = null)
+    public Task<Checker> FindCheckerFor(ConditionGeneration vcgen, Split split = null, CancellationToken? cancellationToken = null)
     {
       if (disposed) {
         return Task.FromException<Checker>(new Exception("CheckerPool was already disposed"));
@@ -45,7 +46,7 @@ namespace VC
       Interlocked.Increment(ref notCreatedCheckers);
 
       var source = new TaskCompletionSource<Checker>();
-      checkerWaiters.Enqueue(source);
+      checkerWaiters.Enqueue((source, cancellationToken));
       return source.Task.ContinueWith(task =>
       {
         PrepareChecker(vcgen.program, split, task.Result);
@@ -93,7 +94,13 @@ namespace VC
         checker.Close();
         return;
       }
-      if (checkerWaiters.TryDequeue(out var waiter)) {
+      if (checkerWaiters.TryDequeue(out var waiterCancellation)) {
+        var (waiter, cancellationToken) = waiterCancellation;
+        if (cancellationToken is {IsCancellationRequested: true}) {
+          waiter.TrySetCanceled();
+          AddChecker(checker);
+          return;
+        }
         if (waiter.TrySetResult(checker)) {
           return;
         }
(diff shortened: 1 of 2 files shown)


Source/VCGeneration/SplitAndVerifyWorker.cs
```diff
@@ -98,10 +98,9 @@ private void TrackSplitsCost(List<Split> splits)
 
     async Task DoWork(Split split, CancellationToken cancellationToken)
     {
-      var checker = await split.parent.CheckerPool.FindCheckerFor(split.parent, split);
+      var checker = await split.parent.CheckerPool.FindCheckerFor(split.parent, split, cancellationToken).WaitAsync(cancellationToken);
```
keyboardDrummer, line 101: Why did you add `.WaitAsync(cancellationToken)`? I think that's only useful when the Task doesn't have cancellation support, but that's what you're already adding in this PR

Source/VCGeneration/CheckerPool.cs
```diff
@@ -11,7 +12,7 @@ namespace VC
   public class CheckerPool
   {
     private readonly ConcurrentStack<Checker> availableCheckers = new();
-    private readonly ConcurrentQueue<TaskCompletionSource<Checker>> checkerWaiters = new();
+    private readonly ConcurrentQueue<(TaskCompletionSource<Checker>, CancellationToken?)> checkerWaiters = new();
```
keyboardDrummer, line 15: I think you don't need to change the field, because you can cancel the `TaskCompletionSource` in the field already.

Here's an example of what I think you could do:

```csharp
  public class CheckerPool
  {
    private readonly ConcurrentStack<Checker> availableCheckers = new();
    private readonly ConcurrentQueue<TaskCompletionSource<Checker>> checkerWaiters = new();
    private int notCreatedCheckers;
    private bool disposed;

    public VCGenOptions Options { get; }

    public CheckerPool(VCGenOptions options)
    {
      this.Options = options;
      notCreatedCheckers = options.VcsCores;
    }

    public Task<Checker> FindCheckerFor(ConditionGeneration vcgen, Split split = null, CancellationToken? cancellationToken = null)
    {
      if (disposed) {
        return Task.FromException<Checker>(new Exception("CheckerPool was already disposed"));
      }
        
      if (availableCheckers.TryPop(out var result)) {
        PrepareChecker(vcgen.program, split, result);
        Contract.Assert(result != null);
        return Task.FromResult(result);
      }

      var afterDecrement = Interlocked.Decrement(ref notCreatedCheckers);
      if (afterDecrement >= 0) {
        var checker = CreateNewChecker();
        PrepareChecker(vcgen.program, split, checker);
        Contract.Assert(checker != null);
        return Task.FromResult(checker);
      }
      Interlocked.Increment(ref notCreatedCheckers);

      var source = new TaskCompletionSource<Checker>();
      cancellationToken?.Register(() =>
      {
        source.SetCanceled(cancellationToken.Value);
      });
      checkerWaiters.Enqueue(source);
      return source.Task.ContinueWith(task =>
      {
        PrepareChecker(vcgen.program, split, task.Result);
        Contract.Assert(task.Result != null);
        return task.Result;
      });
    }

    private Checker CreateNewChecker()
    {
      var log = Options.ProverLogFilePath;
      if (log != null && !log.Contains("@PROC@") && availableCheckers.Count > 0) {
        log = log + "." + availableCheckers.Count;
      } 

      return new Checker(this, log, Options.ProverLogFileAppend);
    }

    public void Dispose()
    {
      while (availableCheckers.TryPop(out var checker)) {
        checker.Close();
      }
      disposed = true;
    }

    void PrepareChecker(Program program, Split split, Checker checker)
    {
      if (checker.WillingToHandle(program) && (split == null || checker.SolverOptions.RandomSeed == split.RandomSeed && !Options.Prune))
      {
        checker.GetReady();
        return;
      }

      checker.Target(program, checker.TheoremProver.Context, split);
      checker.GetReady();
    }

    public void AddChecker(Checker checker)
    {
      if (checker.IsClosed) {
        throw new Exception();
      }
      if (disposed) {
        checker.Close();
        return;
      }

      while (checkerWaiters.TryDequeue(out var waiter)) {
        if (waiter.TrySetResult(checker)) {
          return;
        }
      }
      availableCheckers.Push(checker);
    }

    public void CheckerDied()
    {
      Checker newChecker = null;
      while (checkerWaiters.TryDequeue(out var waiter)) {
        if (waiter.Task.Status == TaskStatus.Running) {
          newChecker ??= CreateNewChecker();
          if (waiter.TrySetResult(newChecker)) {
            return;
          }
        }
      }

      if (newChecker != null) {
        AddChecker(newChecker);
      } else {
        Interlocked.Increment(ref notCreatedCheckers);
      }
    }
  }
```

keyboardDrummer:
I got inspired by this PR and made one that also provides the cancellation feature but including a bugfix and refactoring: https://github.com/boogie-org/boogie/pull/552

MikaelMayer:
Wonderful ! I was thinking that your PR is way better than this one as the refactoring is much cleaner. I'll close this PR.

PR #552, opened 2022-03-24
Introduce AsyncQueue

keyboardDrummer:


Source/Core/AsyncQueue.cs
```diff
@@ -0,0 +1,70 @@
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Boogie;
+
+/// <summary>
+/// A queue from which items can be dequeued even if they have not been enqueued yet.
```
cpitclaudel, line 9: I don't understand that description

Source/Core/AsyncQueue.cs
```diff
@@ -0,0 +1,70 @@
...
+/// Also supports pushing elements to the front of the queue.
+/// </summary>
+public class AsyncQueue<T>
+{
+  private readonly LinkedList<T> items = new();
+  private readonly Queue<TaskCompletionSource<T>> customers = new();
+
+  public void Enqueue(T value)
+  {
+    lock (this) {
```
cpitclaudel, line 20: Avoid locking on `this`

Source/Core/AsyncQueue.cs
```diff
@@ -0,0 +1,70 @@
...
+      return source.Task.ContinueWith(t => t.Result, cancellationToken,
+        TaskContinuationOptions.RunContinuationsAsynchronously, TaskScheduler.Current);
+    }
+  }
+
+  public T[] ClearItems()
+  {
+    lock (this) {
+      var result = new T[items.Count];
+      items.CopyTo(result, 0);
```
cpitclaudel, line 65: Is there no AsArray method on LinkedList?

Source/Core/AsyncQueue.cs
```diff
@@ -0,0 +1,70 @@
...
+  private readonly Queue<TaskCompletionSource<T>> customers = new();
+
+  public void Enqueue(T value)
+  {
+    lock (this) {
+      while (customers.TryDequeue(out var customer)) {
+        if (customer.TrySetResult(value)) {
+          return;
+        }
+      }
```
cpitclaudel, line 25: This logic is duplicated in Push and Enqueue

Source/Core/AsyncQueue.cs
```diff
@@ -0,0 +1,70 @@
...
+public class AsyncQueue<T>
+{
+  private readonly LinkedList<T> items = new();
+  private readonly Queue<TaskCompletionSource<T>> customers = new();
+
+  public void Enqueue(T value)
+  {
+    lock (this) {
+      while (customers.TryDequeue(out var customer)) {
+        if (customer.TrySetResult(value)) {
```
cpitclaudel, line 22: When can this be false?

cpitclaudel (changes requested):
I wonder if the queeue could just be implemented with a semaphore.  You would just release it when enqueuing and wait on it when dequeing.

Source/Core/AsyncQueue.cs
keyboardDrummer: No, I guess because you need to traverse twice to do that so it's a bit unnatural.

Source/Core/AsyncQueue.cs
keyboardDrummer: When the customer Task has been cancelled

Source/Core/AsyncQueue.cs
```diff
@@ -0,0 +1,70 @@
...
+using System.Threading.Tasks;
+
+namespace Microsoft.Boogie;
+
+/// <summary>
+/// A queue from which items can be dequeued even if they have not been enqueued yet.
+/// All methods are thread-safe.
+/// Also supports pushing elements to the front of the queue.
+/// </summary>
+public class AsyncQueue<T>
```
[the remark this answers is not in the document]
keyboardDrummer: This is different from a `ConcurrentQueue` because it allows asynchronously retrieving elements that haven't been added yet.

Source/VCGeneration/CheckerPool.cs
```diff
@@ -10,69 +8,54 @@ namespace VC
...
 
-      var afterDecrement = Interlocked.Decrement(ref notCreatedCheckers);
-      if (afterDecrement >= 0) {
-        var checker = CreateNewChecker();
-        PrepareChecker(vcgen.program, split, checker);
-        Contract.Assert(checker != null);
-        return Task.FromResult(checker);
-      }
-      Interlocked.Increment(ref notCreatedCheckers);
+      var checker = await checkerLine.Dequeue(cancellationToken) ?? CreateNewChecker();
```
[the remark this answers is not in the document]
keyboardDrummer: Because it spins up a process and you might not use them.

Source/VCGeneration/CheckerPool.cs
```diff
@@ -10,69 +8,54 @@ namespace VC
...
 
     public void Dispose()
     {
-      while (availableCheckers.TryPop(out var checker)) {
-        checker.Close();
-      }
       disposed = true;
+      foreach (var checker in checkerLine.ClearItems()) {
+        checker?.Close();
+      }
```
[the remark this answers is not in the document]
keyboardDrummer: Good point. I think in practice there's no issue because CheckerPool is only disposed when all work threads have already been awaited, so no concurrent calls to `CheckerPool.AddChecker` and `CheckerPool.Dispose` are done, but I'll add a lock anyways.

Source/Core/AsyncQueue.cs
MikaelMayer: We might to have to be careful with `TaskCompletionSource`
See here for example.
https://devblogs.microsoft.com/premier-developer/the-danger-of-taskcompletionsourcet-class/

In short, it will invoke the continuation attached to customers.... synchronously. Hence, what happens after `await asyncQueue.Dequeue()` until the next `await` will executed during this call, which might take some time. I looked statically at the hypothetical trace, it would go back to `DoWork`, then `StartCheck`, then `BeginCheck`, then a lot of computation to prepare the next VC condition... and during all that time, the queue is locked and released checkers cannot be used by other tasks.

The solution here is to lock only extracting a customer.

Source/Core/AsyncQueue.cs
keyboardDrummer: > I wonder if the queue could just be implemented with a semaphore. You would just release it when enqueuing and wait on it when dequeing.

Wouldn't the dequeue become synchronous then?

Source/Core/AsyncQueue.cs
```diff
@@ -0,0 +1,88 @@
...
+        return;
+      }
+
+      items.AddLast(value);
+    }
+  }
+
+  private bool TryEnqueue(T value)
+  {
+    while (customers.TryDequeue(out var customer)) {
```
MikaelMayer, line 40: Here is a potential deadlock or at least slow down
https://devblogs.microsoft.com/premier-developer/the-danger-of-taskcompletionsourcet-class/
In short, `customer.TrySetResult` will continue to execute all continuations synchronously, preventing other threads to access this queue.

The solution to that is to lock only when dequeuing customers, not on `customer.TrySetResult(value)`

Possible solution
1) Don't lock around `TryEnqueue`
2) 
```
var consumer;
var dequeued = true;
while(dequeued) {
  lock(lock) {
    dequeued = customers.TryDequeue(out customer)
  }
  if(dequeued) {
    if (customer.TrySetResult(value)) {
      return true;
    }
  }
}
```

MikaelMayer (changes requested):
I think I found a potential deadlock or at least slow down. Can you please investigate it?

Source/Core/AsyncQueue.cs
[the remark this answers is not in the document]
MikaelMayer: Oh this comment is obsolete, looks like you already did it.

Source/Core/AsyncQueue.cs
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  public void Push(T value)
+  {
+    lock (this) {
+      while (customers.TryDequeue(out var customer)) {
+        if (customer.TrySetResult(value)) {
+          return;
+        }
+      }
```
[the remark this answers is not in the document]
MikaelMayer: Obsolete comment

Source/Core/AsyncQueue.cs
keyboardDrummer: This problem is handled on line 74, and the test `EnqueueReturnsBeforeCompletingDequeueTask` tests against this deadlock

Source/Provers/SMTLib/SMTLibProcess.cs
```diff
@@ -427,15 +428,7 @@ async IAsyncEnumerable<SExpr> ParseSExprs(bool top)
...
-      lock (this) {
-        if (proverOutput.TryDequeue(out var result)) {
-          return Task.FromResult(result);
-        }
-
-        var taskCompletionSource = new TaskCompletionSource<string>();
-        outputReceivers.Enqueue(taskCompletionSource);
-        return taskCompletionSource.Task;
-      }
+      return proverOutput.Dequeue(CancellationToken.None);
```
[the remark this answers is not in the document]
keyboardDrummer: I'm not using one on purpose. I've added a comment to explain why.

> This might result in a leak of tasks.

What do you mean by that?

Source/Core/AsyncQueue.cs
keyboardDrummer: Replied in your other comment.

Source/Provers/SMTLib/SMTLibProcess.cs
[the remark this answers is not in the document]
MikaelMayer: I might be wrong, but "Dequeue" enqueues a customer on a list of consumers, and this consumer consists of roughly the continuation of what happens after `ReadProver`
Perhaps should should also set the consumers to cancel when the items are cleared?

Source/Core/AsyncQueue.cs
cpitclaudel: Yeah, I would have just locked a regular queue.  Something like `await semaphore.WaitAsync(); lock(q) { q.Deque(); }`?

Source/Core/AsyncQueue.cs
cpitclaudel: Doesn't this work?

```
using System;
using System.Linq;

var l = new LinkedList<int>();
Console.WriteLine(l.ToArray());
```

Source/Core/AsyncQueue.cs
cpitclaudel: Gosh, good catch.

Source/Core/AsyncQueue.cs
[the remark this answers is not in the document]
cpitclaudel: But what does it mean to retrieve an element that hasn't been added?

Source/Core/AsyncQueue.cs
keyboardDrummer: O yes it does. I guess I didn't have the enumerable extensions when trying this before.

Source/Core/AsyncQueue.cs
MikaelMayer: Great job.
Just a small question then:
`ContinueWith` will always return a task that succeeds, it does not propagate exceptions like the fact that t was canceled. Is that intended?

Source/Core/AsyncQueue.cs
keyboardDrummer: It works well, and it makes AsyncQueue so simple I might remove it altogether. Thanks.

keyboardDrummer: One downside of the `await semaphore.WaitAsync(); lock(q) { q.Dequeue(); }` approach is that the items don't come out in the same order as they're put in. For `CheckerPool` that is fine but for `SMTLibProcess` a little less so.

MikaelMayer: How is that? Isn't dequeuing unloading items the same order?

keyboardDrummer: The test `ItemsAreKeptInOrder` fails in the implementation that uses a SemaphoreSlim, in commit https://github.com/boogie-org/boogie/pull/552/commits/5db0f4dcdaab49f9d605dbf980e02deb5677f278

MikaelMayer: Oh I see. It's probably because `await semaphore.WaitAsync(); ...` does not guarantee that the first thread to reach the semaphore will be triggered next? In that case, you might consider to keep your previous code.

keyboardDrummer: > Oh I see. It's probably because await semaphore.WaitAsync(); ... does not guarantee that the first thread to reach the semaphore will be triggered next? In that case, you might consider to keep your previous code.

Yes that.

Source/Provers/SMTLib/SMTLibProcess.cs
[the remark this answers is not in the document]
MikaelMayer: Your answer is detailed but I am not sure it answers my concern.
I'm probably misunderstanding a point about tasks, I thought that if a task is not canceled or raise an exception, it's going to remain in memory forever. But in our case, since we call CheckerDied which eventually kills the process, even though the task is not cancelled, it will still raise an exception.

MikaelMayer (approved):
Looks good ! Happy to see the code get simpler and even solve potential deadlocks.

Source/Provers/SMTLib/SMTLibProcess.cs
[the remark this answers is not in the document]
keyboardDrummer: > I thought that if a task is not canceled or raise an exception, it's going to remain in memory forever.

I'm not sure I understand what you're unsure of, but I'll share some related thoughts.

My understanding is that a Task is stored on the call stack of whatever thread it's being executed on, and when the Task completes its `ExecuteEntry` method then the call frame that references the Task is popped and the Task can be GC'd.

If there are continuations attached to the Task, then those will execute after the Task has reached some post running state. If the continuations are executed on a different thread than the Task then I imagine the Task will already get GC'd while the continuations run.

> But in our case, since we call CheckerDied which eventually kills the process, even though the task is not cancelled, it will still raise an exception.

In the case of `SMTLibProcess`, the Tasks are referenced by the `Queue<TaskCompletionSource<string>> outputReceivers` field stored in `SMTLibProcess`. When a `Checker` gets GC'd, so will its `SMTLibProcess` and the `SMTLibProcess.prover` field, and then the Task references in `outputReceivers` are lost. The Tasks don't complete with an Exception but their continuations are GC'd and never executed.

Source/VCGeneration/CheckerPool.cs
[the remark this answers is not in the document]
cpitclaudel: 👍

PR #555, opened 2022-03-24
Fix a missing output bug in ConcurrentToSequentialWriteManager

### LINKED ISSUE #553: Assertion failures are sometimes not reported
There seems to be a race condition that can cause part of Boogie's output to be dropped.

The output trailer in the form "Boogie program verifier finished with 0 verified, 3 errors" is still correct, but details about results from one particular implementation can be dropped. This seems to occur in about 1 in a 1000 or 1 in 10.000 times.

Here's an example of what may be missing in the output: 

```
DisjointDomains.bpl(18,5): Error: This assertion might not hold.
Execution trace:
    DisjointDomains.bpl(15,3): start
- DisjointDomains.bpl(25,5): Error: This assertion might not hold.
- Execution trace:
-    DisjointDomains.bpl(24,3): start
DisjointDomains.bpl(31,5): Error: This assertion might not hold.
Execution trace:
    DisjointDomains.bpl(30,3): start

Boogie program verifier finished with 0 verified, 3 errors
```

keyboardDrummer:
### Changes
- Fix a concurrency bug in `ConcurrentToSequentialWriteManager` that would result in lost output. Fixes #553
- Turn off switching hot buffered writers to live writers, since it's difficult to get right. Will reintroduce in a different PR.

### Testing
- Add a test for `ExecutionEngine` that reproduces the missing output reported here: https://github.com/boogie-org/boogie/issues/553
- Add a test for `ConcurrentToSequentialWriteManager` that exercises a certain concurrent scenario which reveals the bug this PR fixes

### DIFF
--- a/Source/ExecutionEngine/ConcurrentToSequentialWriteManager.cs
+++ b/Source/ExecutionEngine/ConcurrentToSequentialWriteManager.cs
@@ -53,11 +53,12 @@ public class ConcurrentToSequentialWriteManager
     }
 
     public string SetTargetAndGetBuffer(TextWriter newTarget) {
-      var oldTarget = target;
-      target = newTarget;
-      var result = buffering ? ((StringWriter)oldTarget).ToString() : "";
-      buffering = false;
-      return result;
+      lock (target) {
+        var result = buffering ? ((StringWriter)target).ToString() : "";
+        target = newTarget;
+        buffering = false;
+        return result;
+      }
     }
 
     protected override void Dispose(bool disposing) {
--- a/Source/ExecutionEngine/ConcurrentToSequentialWriteManager.cs
+++ b/Source/ExecutionEngine/ConcurrentToSequentialWriteManager.cs
@@ -26,9 +26,6 @@ public class ConcurrentToSequentialWriteManager
         var disposedWriter = writers.Dequeue();
         Writer.Write(disposedWriter.SetTargetAndGetBuffer(null));
       }
-      if (writers.Count > 0) {
-        Writer.Write(writers.Peek().SetTargetAndGetBuffer(Writer));
-      }
     }
     Writer.Flush();
   }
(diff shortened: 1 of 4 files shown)


Source/ExecutionEngine/ConcurrentToSequentialWriteManager.cs
```diff
@@ -53,11 +53,12 @@ class SubWriter : WriterWrapper {
...
-      target = newTarget;
-      var result = buffering ? ((StringWriter)oldTarget).ToString() : "";
-      buffering = false;
-      return result;
+      lock (target) {
+        var result = buffering ? ((StringWriter)target).ToString() : "";
+        target = newTarget;
+        buffering = false;
+        return result;
+      }
```
MikaelMayer, line 61: Nice lock that makes sense.

MikaelMayer (approved):
Great job fixing that bug AND writing tests for ensure it won't happen again.

atomb (approved):
This looks nice, and I'm glad you're continuing the removed tests in a separate PR.

cpitclaudel (approved):
LGTM, but why not just use a synchronized textwriter? https://docs.microsoft.com/en-us/dotnet/api/system.io.textwriter.synchronized?view=net-6.0

keyboardDrummer:
> why not just use a synchronized textwriter? https://docs.microsoft.com/en-us/dotnet/api/system.io.textwriter.synchronized?view=net-6.0

That only synchronises between write calls to itself, but I need to synchronise between write calls and a `.ToString` call.

atomb (approved):
Looks good. Nice to have more explicit concurrency tests.

cpitclaudel:
> That only synchronises between write calls to itself, but I need to synchronise between write calls and a .ToString call.

Great point, my bad

PR #561, opened 2022-04-01
Print uses declarations

keyboardDrummer:
### Changes

- Correctly print axioms that are part of uses declarations, instead of as independent axioms.
- Slightly prettier /printPruned option

### Testing

- Test the printing of uses clauses
- Test the output of /printPruned


Source/Core/Absy.cs
```diff
@@ -1565,6 +1551,8 @@ public PredicateCmd CreateCandidateInvariant(Expr e, string tag = null)
   [ContractClass(typeof(DeclarationContracts))]
   public abstract class Declaration : Absy, ICarriesAttributes
   {
+    public virtual IEnumerable<Declaration> SelfAndChildren => Enumerable.Repeat(this, 1);
```
RustanLeino, line 1554: I would have done `yield return this;`, but I now see that `Enumerable.Repeat(this, 1);` will emit less code. Nice.

Test/pruning/UsesClauses.bpl.pruned.after-doesNotHaveAxioms.expect
```diff
@@ -0,0 +1,33 @@
...
+  ensures ConsumerFunc(4) == 3;
+  ensures four == 4;
+
+
+
+implementation hasAxioms()
+{
+    assert ConsumerFunc(4) == 3;
+    assert four == 4;
+}
```
RustanLeino, line 17: Why is `hasAxioms` printed in the `*-doesNotHaveAxioms` output? The `implementation` is not needed at all. The `procedure` is not used. And since these mention `ProducerConst` and `ProducerFunc`, the `*after-*` file prints as as an invalid Boogie program. Wouldn't it be useful if what's printed here does not include unused (`procedure`? and) `implementation` declarations and is something you can feed back into Boogie as a parsing, resolving program?

Test/pruning/UsesClauses.bpl
```diff
@@ -1,5 +1,9 @@
-// RUN: %parallel-boogie /prune /printPruned:"%tpruned" /errorTrace:0 "%s" > "%t"
+// RUN: %parallel-boogie /prune /printPruned:"%t.pruned" /errorTrace:0 "%s" > "%t"
 // RUN: %diff "%s.expect" "%t"
+// RUN: %parallel-boogie /print:"%t.print" /noVerify "%s"
+// RUN: %diff -I "//*" "%s.print.expect" "%t.print"
+// RUN: %diff "%s.pruned.before-doesNotHaveAxioms.expect" "%t.pruned-before-doesNotHaveAxioms"
+// RUN: %diff "%s.pruned.after-doesNotHaveAxioms.expect" "%t.pruned-after-doesNotHaveAxioms"
```
RustanLeino, line 6: I suggest include these (`before-` and `after-`) for `hasAxioms` as well.

Test/pruning/UsesClauses.bpl.pruned.before-doesNotHaveAxioms.expect
```diff
@@ -0,0 +1,53 @@
...
+
+const unique ProducerConst: bool uses {
+  axiom four == 4;
+}
+
+axiom four == 4;
+
+axiom four == 4;
+
+axiom four == 4;
```
RustanLeino, line 11: How come these axioms are printed so many times, included inside a `uses` clause (on L4)?

RustanLeino (changes requested):
Print `uses` clauses is a good (and necessary!) feature. It allows us to `/print` the Boogie program that is actually fed to Boogie.

I don't understand what the `/printPruned` output is printing. I'm confused that `after-` includes `implementation` declarations that are not being checked, and that includes unused `procedure` declarations. After all, the `after-` version does not print unused functions and constants. Also, it confusing that the `before-` version duplicates axioms.

I have a related question regarding the API. If you give Boogie a file to parse, then it's clear that each `axiom` is included just once. But the API seems to want an `axiom` in a `uses` clause to also be added to the list of top-level declarations. It would be nice to say that an `axiom` (just like any other declaration) is supposed to be included only once through the Boogie API, but the fact that you're supposed to include some axioms more than once, then it also makes it more difficult to explain clearly what the rules are regarding including an `axiom` in several `uses` clauses. Can you please provide/document some clarity in this?

Test/pruning/UsesClauses.bpl.pruned.after-doesNotHaveAxioms.expect
keyboardDrummer: > Wouldn't it be useful if what's printed here does not include unused (procedure? and) implementation declarations and is something you can feed back into Boogie as a parsing, resolving program?

Yes, but note that pruning works on a split, which may include only part of an implementation and thus prune some things that are required for the implementation.

Test/pruning/UsesClauses.bpl.pruned.before-doesNotHaveAxioms.expect
keyboardDrummer: That's a bug

keyboardDrummer:
> Print `uses` clauses is a good (and necessary!) feature. It allows us to `/print` the Boogie program that is actually fed to Boogie.
> 
> I don't understand what the `/printPruned` output is printing. I'm confused that `after-` includes `implementation` declarations that are not being checked, and that includes unused `procedure` declarations. 

Pruning only prunes constants, functions and axioms, not implementations and procedures. We could change printPruned not to show implementation and procedures at all, or to only show the implementation and procedure for which pruning was applied. Since pruning applies to a split, which may contain only some of the blocks in an implementation, pruning can still remove declarations used by part of that implementation, so I suggest we don't show any implementations and procedures in `/printPruned`.

> Also, it confusing that the `before-` version duplicates axioms.

That's a bug. Fixed now.

> I have a related question regarding the API. If you give Boogie a file to parse, then it's clear that each `axiom` is included just once. But the API seems to want an `axiom` in a `uses` clause to also be added to the list of top-level declarations. It would be nice to say that an `axiom` (just like any other declaration) is supposed to be included only once through the Boogie API, but the fact that you're supposed to include some axioms more than once, then it also makes it more difficult to explain clearly what the rules are regarding including an `axiom` in several `uses` clauses. Can you please provide/document some clarity in this?

This PR changes that API. After this PR an axiom that's only in a uses clause, but not separately part of the list of TopLevelDeclarations, will still be considered an axiom in the Boogie program. Adding an axiom to two uses clauses will cause it to be part of the top level declarations twice, but I haven't seen a need for doing that in practice.

RustanLeino (approved):
Great output! Thanks for the changes.

keyboardDrummer:
> Great output! Thanks for the changes.

And thanks for the feedback! Especially spotting the somewhat severe bug 😅

cpitclaudel:
Hey folks, I'm seeing performance issues after upgrading to this commit.  Is it possible that this commit is the culprit, or is it unlikely / should I dig deeper?

Session transcript below, but basically running Dafny with this commit on the project I'm currently working on takes 14-16 seconds, and running without it takes 10-11 seconds.

```console
$ git apply customBoogie.patch 

$ dotnet clean Source/Dafny.sln && dotnet build -c Release Source/Dafny.sln
…
Time Elapsed 00:00:06.15

$ time Binaries/net6.0/Dafny -compile:0 -proverOpt:SOLVER=noop example/Example.dfy
…
Dafny program verifier finished with 0 verified, 0 errors, 67 inconclusives

real	0m13.868s
user	0m0.000s
sys	0m0.031s

$ time Binaries/net6.0/Dafny -compile:0 -proverOpt:SOLVER=noop example/Example.dfy
…
Dafny program verifier finished with 0 verified, 0 errors, 67 inconclusives

real	0m14.213s
user	0m0.000s
sys	0m0.031s

$ (cd boogie; git checkout ca49a8c7)
…

HEAD is now at ca49a8c7 Bring back buffering to live writer switch (#557)

$ dotnet clean Source/Dafny.sln && dotnet build -c Release Source/Dafny.sln
…
Time Elapsed 00:00:06.98

$ time Binaries/net6.0/Dafny -compile:0 -proverOpt:SOLVER=noop example/Example.dfy
…
Dafny program verifier finished with 0 verified, 0 errors, 67 inconclusives

real	0m11.470s
user	0m0.000s
sys	0m0.015s

$ time Binaries/net6.0/Dafny -compile:0 -proverOpt:SOLVER=noop example/Example.dfy
…
Dafny program verifier finished with 0 verified, 0 errors, 67 inconclusives

real	0m11.559s
user	0m0.000s
sys	0m0.031s

$ time Binaries/net6.0/Dafny -compile:0 -proverOpt:SOLVER=noop example/Example.dfy
…
Dafny program verifier finished with 0 verified, 0 errors, 67 inconclusives

real	0m11.739s
user	0m0.000s
sys	0m0.015s

$ (cd boogie; git checkout 831708b6)
Previous HEAD position was ca49a8c7 Bring back buffering to live writer switch (#557)
HEAD is now at 831708b6 Print uses declarations (#561)

$ dotnet clean Source/Dafny.sln && dotnet build -c Release Source/Dafny.sln
…
Time Elapsed 00:00:06.83

$ time Binaries/net6.0/Dafny -compile:0 -proverOpt:SOLVER=noop example/Example.dfy
…
Dafny program verifier finished with 0 verified, 0 errors, 67 inconclusives

real	0m14.375s
user	0m0.000s
sys	0m0.031s

$ time Binaries/net6.0/Dafny -compile:0 -proverOpt:SOLVER=noop example/Example.dfy
…
Dafny program verifier finished with 0 verified, 0 errors, 67 inconclusives

real	0m16.943s
user	0m0.000s
sys	0m0.000s
```

This was with my own code as `example/Example.dfy`, but here's what I get with the test suite:

- For `Test/concurrency/12-MutexLifetime-long.dfy`, I get 15-16 seconds on `ca49a8c7 * Bring back buffering to live writer switch`, 18-19s on `831708b6 * Print uses declarations (#561)`
- For `cloudmake/CloudMake-CachedBuilds.dfy`, I get 5-6 seconds on `ca49a8c7 * Bring back buffering to live writer switch`, and about 7s on `831708b6 * Print uses declarations (#561)`.

Can you you check if you see the same thing? Thanks!

keyboardDrummer:
@cpitclaudel Thanks for the research, seems like a bug. I'll look into it.

cpitclaudel:
Thanks.  Please do consider that I may be mismeasuring something in some way! If so, sorry for the noise :/

Issue #562, opened 2022-04-01
RFC: deprecate or remove ErrorData?


atomb:
Many of the command types in Boogie have an `ErrorData` property which is typically a `string` but in many cases allowed to be an arbitrary `object`. I also recently added a `Description` property, of type `ProofObligationDescription` to those same commands. The primary goal of this new property was to allow separate messages for failed and successful proof of a particular proof obligation. Dafny subclasses this type extensively, and may eventually extend its own subclasses to include things like the Dafny syntax of the expression leading to a particular proof obligation. The benefit I see to this approach is that the `Description` is then always guaranteed to contain at least a success and failure message, and may also contain any client-specific data that might be useful.

The question I want to raise here is this: does `ProofObligationDescription` meet the needs of everyone who's currently using `ErrorData`? And would it therefore make sense to deprecate (and eventually remove?) the `ErrorData` property?

keyboardDrummer:
You could check https://github.com/boogie-org/corral and https://github.com/mc-imperial/gpuverify, those are the two projects that consume Boogie through the C# API that I'm aware of.

bkragl:
Civl uses `ErrorData` extensively to provide meaningful error messages (mainly through [this helper](https://github.com/boogie-org/boogie/blob/831708b6005ed8ac7f72f37a691be6deee375b13/Source/Concurrency/CivlUtil.cs#L128)). What needs to be done to preserve these messages if `ErrorData` would be removed? (cc @shazqadeer)

atomb:
The migration path would be to construct a subclass of [`ProofObligationDescription`](https://github.com/boogie-org/boogie/blob/master/Source/Core/ProofObligationDescription.cs) and store instances of it in the `Description` property instead of the `ErrorData` property. To retain identical functionality to what you have right now, it would be possible to create a class that just stored the string you're currently using for the message. To make things slightly nicer, you could have a different subclass for each possible type of property that Civl checks, potentially including extra metadata, with messages to use for the case of both successful and unsuccessful proof.

keyboardDrummer:
But since Civl is part of Boogie, the migration would be done in the PR that removes ErrorData, right @atomb ?

atomb:
I'm happy to make these changes, @shazqadeer. I think migrating Civl should be easy, and I'll make sure to include @bkragl on the reviewers list, as the one who raised that point.

PR #564, opened 2022-04-11
Implementation verification api

keyboardDrummer:
### Changes

- Add the method `ExecutionEngine.GetImplementationTasks`, an API for identifying, starting and stopping individual verification tasks, which is used by https://github.com/dafny-lang/dafny/pull/2004
- Fix bug in `SMTLibProcessTheoremProver.InitializeGlobalInformation()`
- Refactor: Move `Program.ExtractLoops` and related method to a new file `LoopExtractor.cs`
- Refactor: Move `OutputPrinter` from ExecutionEngine.cs to `OutputPrinter.cs`
- Refactor: Move various error information related classes from `ExecutionEngine.cs` to `ErrorInformation.cs`
- Refactor: Remove the unused fields `Counterexample.OriginalRequestId` and `Counterexample.RequestId`

### Tests
- Add test for `ExecutionEngine.GetImplementationTasks`


Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
```diff
@@ -1108,7 +1108,7 @@ protected void AddTypeDecl(string decl)
 
     ////////////////////////////////////////////////////////////////////////////
 
-    private static string _backgroundPredicates;
+    private string _backgroundPredicates;
```
MikaelMayer, line 1111: Is this fixing the bug you mention? Can you cover it with a test so that it won't be repeated again?

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -11,174 +11,16 @@
...
-      }
-    }
-  }
-
-  #endregion
-
   public class ExecutionEngine : IDisposable {
-    public static ErrorInformationFactory ErrorInformationFactory { get; } = new();
+
+    private static readonly ConditionalWeakTable<Program, Action<VCGen, Implementation, VerificationResult>> ResultPostProcessors = new ();
```
MikaelMayer, line 23: I understand that you are setting a callback for when the Boogie program is done verifying, so that the routines of `ExtractLoop` can be performed independently of the verification engine and these callbacks that don't prevent the garbage collector to be collected if no reference points to Program anymore (e.g. when verification is canceled)

I also understand that you want to run these action synchronously, so you are not using TaskCompletionSource.

I'm puzzled because you are not deleting the callback when it is processed, which I think it should. I'm not sure about garbage collection, but for example in Java it's lazy and might not follow the links of a ConditionalWeakTable to determine whether garbage collection is needed, but programs can be quite heavy.

* The principle of "hiding away" this logic is hindered by the fact that there is also at most one Action to run per program, so it's not even an architecture that will enable more use cases.
* Could you perhaps have a synchronously executed task instead?
* I would prefer to need not a static shared variable caching all these callbacks, but rather somehow store the callback in the execution that needs it, so that there is no need for a table.

Please let me know what you think of the above.

MikaelMayer (changes requested):
It feels great to be able to access the verification tasks of different implementations ! This is going to be super useful to control verification.

Source/VCGeneration/Counterexample.cs
```diff
@@ -81,8 +83,6 @@ void ObjectInvariant()
     public Model Model;
     public VC.ModelViewInfo MvInfo;
     public ProverContext Context;
-    public string OriginalRequestId;
-    public string RequestId;
```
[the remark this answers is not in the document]
keyboardDrummer: The reason I wanted to remove the fields is that they enable removing one parameter from `VerifyImplementationWithLargeStackScheduler`, which is called by the new code in `ImplementationTask`, and which would otherwise pass in `null` for this parameter since it doesn't have a requestId to pass in. There's other parameters of `VerifyImplementationWithLargeStackScheduler` like `ErrorReporterDelegate er` that I think are redundant, but I've left those alone for now.

I checked that Corral and GPUVerify do not use the fields that I've removed, and AFAIK there aren't other consumers of the Boogie's C# API.

Source/ExecutionEngine/ExecutionEngine.cs
keyboardDrummer: I've replaced the use of `ConditionalWeakTable` with an extra record. Curious if you like the new implementation better.

Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
keyboardDrummer: I just tried doing that but to enable defining the test without a lot of code requires some more refactoring in options related code that I'd do as part of another PR. 
Also, it would be a test that closely follows the implementation of `InitializeGlobalInformation` so I don't think it would be too interesting.
Also, without this change the current test suite fails, although it requires executing a combination of existing tests.
Also, having a static field like this was is a smell and I don't think we'll bring that back anytime soon.

Can we leave this test out for now?

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -995,20 +848,14 @@ private VerificationResult GetCachedVerificationResult(Implementation impl, Text
...
-            }
-          }
+          await vcgen.VerifyImplementation(new ImplementationRun(impl, traceWriter), cancellationToken);
+        if (ResultPostProcessors.TryGetValue(program, out var postProcess)) {
+          postProcess!(vcgen, impl, verificationResult);
         }
       } catch (VCGenException e) {
-        var errorInfo = ErrorInformationFactory.CreateErrorInformation(impl.tok,
-          $"{e.Message} (encountered in implementation {impl.Name}).", requestId, "Error");
+        var errorInfo = ErrorInformationFactory.Instance.CreateErrorInformation(impl.tok,
```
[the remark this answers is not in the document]
MikaelMayer: Oh I see, the field was declared there
https://github.com/boogie-org/boogie/pull/564/files#diff-a5b39418ae8731b4b74a8cb24041d0272c8e4a02316c312cdc2c1bb19dff7d02L181
Good job of getting further inlining that static method.

Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
MikaelMayer: Ok for the explanation. Could you then please move this field with the other fields at the beginning, so that at least it's obvious why it should not be static. Thanks :-)

Source/ExecutionEngine/ExecutionEngine.cs
MikaelMayer: That's great ! I was wondering how you would do the change, and what you did made me better realize the flow, i.e. where is this callback is being used. At least for me it's definitely more readable that way, compared to the previous "stateful" implementation. I thank you for the effort!
I'll have a few suggestions about the naming conventions.

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -1018,20 +862,20 @@ private VerificationResult GetCachedVerificationResult(Implementation impl, Text
         }
 
         verificationResult.Errors = null;
-        verificationResult.Outcome = VCGen.Outcome.Inconclusive;
+        verificationResult.Outcome = ConditionGeneration.Outcome.Inconclusive;
```
cpitclaudel, line 865: Why rename this?  `VCGen` is nice and readable

Source/VCGeneration/Counterexample.cs
```diff
@@ -251,6 +240,22 @@ public void PrintModel(TextWriter tw, Counterexample counterexample)
       }
     }
 
+    public void InitializeStates()
+    {
+      if (Model is { ModelHasStatesAlready: false }) {
```
cpitclaudel, line 245: Is that the same as `Model?.ModelHasStatesAlready == false`?

Source/VCGeneration/Counterexample.cs
```diff
@@ -251,6 +240,22 @@ public void PrintModel(TextWriter tw, Counterexample counterexample)
...
+        if (this is AssertCounterexample assertError) {
```
cpitclaudel, line 246: Could we use a `case` instead of an if with `is` in each branch?

Source/VCGeneration/Counterexample.cs
```diff
@@ -386,6 +391,113 @@ public Model.Element GetModelValue(Variable v)
...
+        cause = "Out of memory on";
+      }
+      else if (outcome == VCGen.Outcome.SolverException)
+      {
+        cause = "Solver exception on";
+      }
+      else if (outcome == VCGen.Outcome.OutOfResource)
+      {
+        cause = "Out of resource on";
+      }
```
cpitclaudel, line 414: Another good candidate for a switch

Source/VCGeneration/Counterexample.cs
```diff
@@ -386,6 +391,113 @@ public Model.Element GetModelValue(Variable v)
...
+        }
+      }
+      else if (this is ReturnCounterexample returnError)
+      {
+        if (returnError.FailingEnsures.ErrorMessage == null || forceBplErrors)
+        {
+          errorInfo = ErrorInformation.Create(returnError.FailingReturn.tok, returnError.FailingReturn.Description.FailureDescription, cause);
+          errorInfo.Kind = ErrorKind.Postcondition;
+          errorInfo.AddAuxInfo(returnError.FailingEnsures.tok,
+            returnError.FailingEnsures.ErrorData as string ?? returnError.FailingEnsures.Description.FailureDescription,
```
cpitclaudel, line 439: Does this pattern of using `ErrorData as string ?? …` deserve an extension method?

Source/VCGeneration/ErrorInformation.cs
```diff
@@ -0,0 +1,94 @@
...
+    if (msg.ToLower().StartsWith("error: "))
+    {
+      return msg.Substring(7);
+    }
+    else
+    {
+      return msg;
+    }
+  }
+}
```
cpitclaudel, line 92: How does this interact with @atomb 's classification of errors?

Source/VCGeneration/LoopExtractor.cs
```diff
@@ -0,0 +1,621 @@
...
+          continue;
+        }
+
+        immSuccBlks.Add(bl);
+      }
+    }
+
+    return immSuccBlks;
+  }
+}
```
cpitclaudel, line 621: Moving things like this makes reviewing and tracing bugs harder.  It may still be worth it, but the usual git facilities for inspecting code break badly in the face of this.

Source/VCGeneration/StratifiedVC.cs
```diff
@@ -14,6 +14,9 @@ namespace VC
 {
   using Bpl = Microsoft.Boogie;
 
+  /// <summary>
+  /// Used in Corral https://github.com/boogie-org/corral
+  /// </summary>
```
cpitclaudel, line 19: Unless that class only exists for that purpose, should this really be an XML doc comment?

Source/ExecutionEngine/ConsolePrinter.cs
```diff
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Diagnostics.Contracts;
 using System.IO;
+using VCGeneration;
```
cpitclaudel, line 5: ?

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -1152,83 +996,93 @@ private async Task<PipelineOutcome> RunStagedHoudini(Program program, PipelineSt
...
-              {
+              foreach (Counterexample error in timedOutAssertions) {
                 var callError = error as CallCounterexample;
                 var returnError = error as ReturnCounterexample;
                 var assertError = error as AssertCounterexample;
                 IToken tok = null;
                 string msg = null;
-                if (callError != null)
-                {
+                if (callError != null) {
```
cpitclaudel, line 1071: Why reformat these?

cpitclaudel (commented):
I'm having trouble getting a sense of what actually changed (vs got moved or refactored) in the PR, so the comments I left are all very superficial :/

Source/VCGeneration/StratifiedVC.cs
MikaelMayer: Do you know of other use cases?

Test/test15/ModelTest.bpl.expect
```diff
@@ -19,6 +21,12 @@ tickleBool -> {
   true -> true
   else -> true
 }
+*** STATE <initial>
```
MikaelMayer, line 24: What are these new lines? Are they just debug lines?

MikaelMayer (changes requested):
A few minor changes (naming conventions, spacing)
But otherwise, I'm ready to approve this PR.
Just a question: How does your feature interact with the `{:priority}` attribute? Can you confirm it works the same?

Source/ExecutionEngine/ExecutionEngine.cs
keyboardDrummer: It's not a rename.`Outcome` is defined `ConditionGeneration` and `VCGen` inherits from `ConditionGeneration`. Rider recommends using the defining class as a reference, which I think is good because it reduces ambiguity.

Source/VCGeneration/Counterexample.cs
keyboardDrummer: Yes

Source/VCGeneration/Counterexample.cs
keyboardDrummer: Used an abstract method instead.

Source/VCGeneration/Counterexample.cs
keyboardDrummer: There's an issue about removing ErrorData: https://github.com/boogie-org/boogie/issues/562

Source/VCGeneration/ErrorInformation.cs
keyboardDrummer: This is a move of code, so there should be no change.

Source/VCGeneration/LoopExtractor.cs
keyboardDrummer: Originally it was sufficient to review the refactor commits separately from the other commits, but I've since added a merge commit in which several conflicts were resolved. I'm not sure what the best workflow is here. Maybe I should have done a rebase instead then you could have still reviewed the individual commits.

Source/VCGeneration/StratifiedVC.cs
keyboardDrummer: I've only found it to be used in Corral.

Source/ExecutionEngine/ConsolePrinter.cs
keyboardDrummer: It is used..

Source/ExecutionEngine/ExecutionEngine.cs
keyboardDrummer: I think the IDE did this when I extracted the code into a separate method

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -10,175 +10,17 @@
...
-  public class ErrorInformationFactory
-  {
-    public virtual ErrorInformation CreateErrorInformation(IToken tok, string msg, string requestId = null,
-      string originalRequestId = null, string category = null)
-    {
-      Contract.Requires(1 <= tok.line && 1 <= tok.col);
-      Contract.Requires(msg != null);
-
-      return ErrorInformation.CreateErrorInformation(tok, msg, requestId, originalRequestId, category);
+  public record ProcessedProgram(Program Program, Action<VCGen, Implementation, VerificationResult> PostProcessResult) {
```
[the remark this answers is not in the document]
keyboardDrummer: I like the last one. The first one I'm not a fan of because I don't like widening a type when the original type is sufficient. The second one is I think slightly less future proof than the current option, since you can imagine a function that takes a `ProcessedProgram` and further processes it to return another `ProcessedProgram`, but that would be difficult when `ProcessedProgram` is already a specific subclass. The third one is more specific than the current solution which I like, but the name exposes slightly more details than I'd like to.

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -853,18 +710,14 @@ private Implementation[] GetPrioritizedImplementations(Program program)
...
           programId, taskWriter), cts.Token);
 
         var verificationResult = await coreTask;
-        var output = verificationResult.GetOutput(Options.Printer, this, stats, er, implementation);
-
-        await taskWriter.WriteAsync(output);
         return verificationResult;
       }
       finally {
-        taskWriter.Close();
```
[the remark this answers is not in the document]
keyboardDrummer: It's moved to the caller. I think it's bad practice to close a resource that you didn't create.

Source/VCGeneration/LoopExtractor.cs
cpitclaudel: Github having trouble is fine, because I can review locally.  What I'm more concerned about is `git-log -L`, which is extremely useful for understanding the history of a function, and also completely broken by renames / moves.

Source/VCGeneration/StratifiedVC.cs
cpitclaudel: I was suggesting to convert it to a regular comment.

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -1152,83 +996,93 @@ private async Task<PipelineOutcome> RunStagedHoudini(Program program, PipelineSt
...
     }
 
     public void ReportOutcome(OutputPrinter printer,
       ConditionGeneration.Outcome outcome, ErrorReporterDelegate er, string implName,
-      IToken implTok, string requestId, string msgIfVerifies, TextWriter tw, uint timeLimit, List<Counterexample> errors)
+      IToken implTok, string msgIfVerifies, TextWriter tw, uint timeLimit, List<Counterexample> errors) {
+
+      var errorInfo = GetOutcomeError(Options, outcome, implName, implTok, msgIfVerifies, tw, timeLimit, errors);
+      if (errorInfo != null)
+      {
```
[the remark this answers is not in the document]
cpitclaudel: Agreed, if the code is getting moved around anyway then it's not a bad time to clean up indentation/formatting

Test/test15/ModelTest.bpl.expect
keyboardDrummer: There's a line somewhere
`error.Model.Write(Options.ModelWriter ?? errorInfo.Out);`
that can print the model without the model states having been initialised. A different invocation that would automatically initialise the states would be:
`error.PrintModel(Options.ModelWriter ?? errorInfo.Out, error);`

Because now we strictly initialise the model states, these model prints are slightly larger.

Source/VCGeneration/StratifiedVC.cs
keyboardDrummer: Okay, why? It serves as a (poor) explanation of why this exists, not how it works.

MikaelMayer (dismissed):

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -10,17 +10,17 @@
...
-using System.Net.Mime;
 using System.Runtime.CompilerServices;
-using Core;
-using Microsoft.Boogie.Houdini;
 using VCGeneration;
 
 namespace Microsoft.Boogie
 {
-  public class ExecutionEngine : IDisposable {
+  public record ProcessedProgram(Program Program, Action<VCGen, Implementation, VerificationResult> PostProcessResult) {
```
cpitclaudel, line 18: How many different instances of this record do you create?  I'm more used to seeing abstract methods than fields of type `Action`, so if there are not too many different post-processors an abstract class may be better?

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -10,17 +10,17 @@
...
+    public ProcessedProgram(Program program) : this(program, (_, _, _) => { }) {
+    }
+  }
 
-    private static readonly ConditionalWeakTable<Program, Action<VCGen, Implementation, VerificationResult>> ResultPostProcessors = new ();
```
cpitclaudel: Ah, never mind, I didn't notice that the code was already like that

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -672,8 +671,8 @@ private Implementation[] GetPrioritizedImplementations(Program program)
         CleanupRequest(requestId);
       }
 
-      if (Options.PrintNecessaryAssumes && program.NecessaryAssumes.Any()) {
-        Console.WriteLine("Necessary assume command(s): {0}", string.Join(", ", program.NecessaryAssumes.OrderBy(s => s)));
+      if (Options.PrintNecessaryAssumes && processedProgram.Program.NecessaryAssumes.Any()) {
+        Console.WriteLine("Necessary assume command(s): {0}", string.Join(", ", processedProgram.Program.NecessaryAssumes.OrderBy(s => s)));
```
cpitclaudel, line 675: I don't understand the meaning of the word `processed` here.  What makes a program "processed"?

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -1097,30 +1092,30 @@ private async Task<PipelineOutcome> RunStagedHoudini(Program program, PipelineSt
...
           }
 
           break;
 
         case VCGen.Outcome.Inconclusive:
           if (implName != null && implTok != null) {
-            errorInfo = ErrorInformationFactory.Instance.CreateErrorInformation(implTok,
-              "Verification inconclusive (" + implName + ")");
+            string msg = "Verification inconclusive (" + implName + ")";
+            errorInfo = ErrorInformation.Create(implTok, msg, null);
```
cpitclaudel, line 1118: Maybe `ErrorInformation.Create` should have a default argument?

Source/VCGeneration/ConditionGeneration.cs
```diff
@@ -118,17 +118,16 @@ public ConditionGeneration(Program p, CheckerPool checkerPool)
...
       Contract.Requires(run != null);
 
       Contract.EnsuresOnThrow<UnexpectedProverOutputException>(true);
       Helpers.ExtraTraceInformation(Options, "Starting implementation verification");
 
       var collector = new VerificationResultCollector(Options);
-      collector.RequestId = requestId;
       Outcome outcome = await VerifyImplementation(run, collector, cancellationToken);
-      List<Counterexample> /*?*/ errors = null;
+      List<Counterexample> /*?*/ errors = new List<Counterexample>();
```
cpitclaudel, line 130:
```suggestion
      List<Counterexample> /*?*/ errors = new ();
```

Source/VCGeneration/Counterexample.cs
```diff
@@ -134,7 +134,7 @@ public void AddCalleeCounterexample(Dictionary<TraceLocation, CalleeCounterexamp
     }
 
     // Looks up the Cmd at a given index into the trace
-    public Cmd getTraceCmd(TraceLocation loc)
+    public Cmd GetTraceCmd(TraceLocation loc)
```
cpitclaudel, line 137: Is it OK to rename these public methods?

Source/VCGeneration/Counterexample.cs
```diff
@@ -251,6 +240,22 @@ public void PrintModel(TextWriter tw, Counterexample counterexample)
       }
     }
 
+    public void InitializeStates()
+    {
+      if (!Model.ModelHasStatesAlready) {
```
cpitclaudel: I would use `if (Model.ModelHasStatesAlready) { return; }`

Source/VCGeneration/Counterexample.cs
```diff
@@ -251,6 +240,22 @@ public void PrintModel(TextWriter tw, Counterexample counterexample)
       }
     }
 
+    public void InitializeStates()
```
cpitclaudel, line 243: `InitializeModelStates` maybe?

Source/VCGeneration/Counterexample.cs
```diff
@@ -386,6 +391,98 @@ public Model.Element GetModelValue(Variable v)
...
+        else
+        {
+          errorInfo = ErrorInformationFactory.Instance.CreateErrorInformation(null,
+            returnError.FailingEnsures.ErrorMessage);
+        }
+      }
+      else // error is AssertCounterexample
+      {
+        Debug.Assert(this is AssertCounterexample);
+        var assertError = (AssertCounterexample)this;
```
cpitclaudel, line 434: This line will already throw an invalid-cast exception if the type isn't right; do you need the Debug.Assert above?

Source/VCGeneration/Counterexample.cs
keyboardDrummer: I think so, GPUVerify has said they're OK with minor API changes. The other C# customer AFAIK is Corral but I think similarly small changes are OK.

Source/VCGeneration/Counterexample.cs
keyboardDrummer: Resolved in next diff :/ https://github.com/boogie-org/boogie/pull/564/files/e9319c3660598907ec82859218df878a1896fa61..a24711bafc7bf540bf0579ba7e38cce3d6787cb6

Source/VCGeneration/Counterexample.cs
keyboardDrummer: Removed.

cpitclaudel (approved):

PR #570, opened 2022-05-02
Bump version to 2.14.2 and revert https://github.com/boogie-org/boogie/pull/561/files

keyboardDrummer:
Reverting because of a performance regression: https://github.com/boogie-org/boogie/pull/561#issuecomment-1099877898


Source/Directory.Build.props
```diff
@@ -2,7 +2,7 @@
 
   <!-- Target framework and package configuration -->
   <PropertyGroup>
-    <Version>2.14.1</Version>
+    <Version>2.14.2</Version>
```
atomb, line 5: I'd lean slightly toward calling this 2.15.0 and including https://github.com/boogie-org/boogie/pull/567. I don't feel very strongly about the version, but I would very much like it to include that PR.

atomb (approved):

Source/Directory.Build.props
robin-aws: Definitely should be 2.15.0 now that #567 is included.

robin-aws (approved):

PR #571, opened 2022-05-04
Add a verification status API 

keyboardDrummer:
### Changes

Add an API to IImplementationTask that allows observing when the status of an implementation's verification changes

### Testing

Added a unit test


Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -694,7 +694,7 @@ private Implementation[] GetPrioritizedImplementations(Program program)
       return GetPrioritizedImplementations(program).Select(implementation => new ImplementationTask(this, processedProgram, implementation)).ToList();
     }
 
-    public async Task<VerificationResult> VerifyImplementationWithLargeStackScheduler(
+    public async Task<Task<VerificationResult>> EnqueueVerifyImplementation(
```
MikaelMayer, line 697:
```suggestion
    /// The outer task is to wait for a semaphore to let verification start
    /// The inner task is the actual verification of the implementation
    public async Task<Task<VerificationResult>> EnqueueVerifyImplementation(
```
```suggestion
    public async Task<Task<VerificationResult>> EnqueueVerifyImplementation(
```

When something is weird, such as a nested task here, I'd strongly encourage adding a comment to the method, even if it is private. It took me substantial time to figure out why two nested tasks were needed.

MikaelMayer (changes requested):
This is an interesting API. A few minor things to fix I believe.

What is the use case of this API ? What are the cases where it will it be useful to proactively query for the status of the verification of an implementation, rather than inferring it from the callbacks implemented in the OutputPrinter?

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -706,21 +706,17 @@ private Implementation[] GetPrioritizedImplementations(Program program)
...
+        implementation,
+        programId, taskWriter), cts.Token);
 
-        var verificationResult = await coreTask;
-        return verificationResult;
-      }
-      finally {
+      var _ = coreTask.ContinueWith(t => {
         verifyImplementationSemaphore.Release();
         ImplIdToCancellationTokenSource.TryRemove(id, out old);
```
cpitclaudel, line 717: Which thread does `ContinueWith` run on?  Is it safe to release on the thread?

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -706,21 +706,17 @@ private Implementation[] GetPrioritizedImplementations(Program program)
...
-      }
+      });
+      return coreTask;
```
cpitclaudel, line 719: I don't understand the  the double `Task` :/

Source/ExecutionEngine/ImplementationTask.cs
```diff
@@ -1,35 +1,97 @@
...
+
+        var enqueueTask = engine.EnqueueVerifyImplementation(ProcessedProgram, new PipelineStatistics(), null, null,
+          Implementation, taskCancellationSource, TextWriter.Null);
+
+        if (enqueueTask.IsCompleted) {
+          CurrentStatus = VerificationStatus.Verifying;
+        } else {
+          CurrentStatus = VerificationStatus.Queued;
+          enqueueTask.ContinueWith(_ => { CurrentStatus = VerificationStatus.Verifying; });
+        }
```
cpitclaudel, line 62: Could the task itself send notifications, rather than having the two layers of `Task`?

keyboardDrummer:
> What is the use case of this API ?

The use-case is for publishing verification status notifications in the Dafny language server.

> What are the cases where it will it be useful to proactively query for the status of the verification of an implementation, rather than inferring it from the callbacks implemented in the OutputPrinter?

I think you can most of the same information as this API provides from the `ReportStartVerifyImplementation` and `ReportEndVerifyImplementation` methods in `OutputPrinter`. One semantic difference is that this API also allows you to see if results for an implementation are available in the cache without starting verification of that implementation, by checking if the status starts at `Completed` instead of `Stale`.

There's also some UX differences. I prefer these but I think that's subjective:
- It uses an Observable<> as a return type instead of a callback function as an argument.
- There's a `CurrentState` field next to the Observable in case you want to know the current state.
- The state changes are differentiated by enum values instead of different callback functions.
- The API is contained inside an IImplementationTask which groups APIs related to the verification of an Implementation.

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -706,21 +706,17 @@ private Implementation[] GetPrioritizedImplementations(Program program)
...
-          programId, taskWriter), cts.Token);
+      var coreTask = Task.Run(() => VerifyImplementation(processedProgram, stats, er, cts.Token,
+        implementation,
+        programId, taskWriter), cts.Token);
 
-        var verificationResult = await coreTask;
-        return verificationResult;
-      }
-      finally {
+      var _ = coreTask.ContinueWith(t => {
```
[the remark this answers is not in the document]
keyboardDrummer: I think it will. The callback in `ContinueWith` is called if the Task goes to the completed state, either successfully, cancelled or if an exception occurred.

Source/ExecutionEngine/ExecutionEngine.cs
keyboardDrummer: Good suggestion, thanks.

Source/ExecutionEngine/ExecutionEngine.cs
keyboardDrummer: Where the continuation runs is scheduled by `TaskScheduler.Current`. Does it matter?

> Is it safe to release on the thread?

Why wouldn't it be? `verifyImplementationSemaphore.Release();` is thread-safe.

Source/UnitTests/ExecutionEngineTests/ExecutionEngineTest.cs
```diff
@@ -178,4 +181,54 @@ public async Task LoopInvariantDescriptions()
...
+    Assert.AreEqual((secondName, VerificationStatus.Queued), statusList[1]);
+    Assert.AreEqual((firstName, VerificationStatus.Error), statusList[2]);
+    Assert.AreEqual((secondName, VerificationStatus.Verifying), statusList[3]);
+    Assert.AreEqual((secondName, VerificationStatus.Correct), statusList[4]);
+    
+    var tasks2 = engine.GetImplementationTasks(program);
+    Assert.AreEqual(VerificationStatus.Error, tasks2[0].CurrentStatus);
+    Assert.AreEqual(VerificationStatus.Correct, tasks2[1].CurrentStatus);
+    var statuses2 = first.ObservableStatus.Merge(second.ObservableStatus);
+    Assert.IsFalse(await statuses2.Any().ToTask());
```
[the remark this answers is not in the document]
keyboardDrummer: I've added:

```
    Assert.AreEqual(VerificationStatus.Stale, first.CurrentStatus);
    Assert.AreEqual(VerificationStatus.Stale, second.CurrentStatus);
```

somewhere in the tests.

Source/UnitTests/ExecutionEngineTests/ExecutionEngineTest.cs
```diff
@@ -214,20 +214,23 @@ public async Task LoopInvariantDescriptions()
     var statuses = first.ObservableStatus.Select(s => (implementationTask: first.Implementation.Name, s)).
       Merge(second.ObservableStatus.Select(status => (second.Implementation.Name, s: status)));
     statuses.Subscribe(t => statusList.Add(t));
-    
+
+    Assert.AreEqual(VerificationStatus.Stale, first.CurrentStatus);
+    Assert.AreEqual(VerificationStatus.Stale, second.CurrentStatus);
```
MikaelMayer, line 219: Yes, that's what I wanted, thanks.

MikaelMayer (dismissed):
Just three non-blocking remarks. The last one could result in a change though.
Very good job simplifying the PR.

MikaelMayer (approved):
Great job !

PR #574, opened 2022-05-18
[Civl] Fix reserve example

shazqadeer:
This example was unverified earlier. This PR is an attempt to verify using lemma procedures that introduce facts about cardinality of sets.


shazqadeer:
I don't understand the mysterious failure of the CI run.

bkragl:
The file `Test/civl/lit.local.cfg` should be removed, because it excludes `reserve.bpl`.

shazqadeer:
https://github.com/boogie-org/boogie/pull/576

PR #579, opened 2022-05-26
Don't generate SMT-Lib with empty attribute lists

atomb:
When using `(! ... :attr)` to add a attributes to an SMT-Lib expression, fix the calculation of whether or not any attributes exist (because it's invalid SMT-Lib syntax if there are none).

### DIFF
--- a/Source/Provers/SMTLib/SMTLibLineariser.cs
+++ b/Source/Provers/SMTLib/SMTLibLineariser.cs
@@ -447,7 +447,10 @@ namespace Microsoft.Boogie.SMTLib
           weight = 1;
         }
 
-        var hasAttrs = node.Triggers.Count > 0 || info.qid != null || weight != 1 || info.uniqueId != -1;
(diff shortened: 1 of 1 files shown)


Source/Provers/SMTLib/SMTLibLineariser.cs
```diff
@@ -447,7 +447,10 @@ public bool Visit(VCExprQuantifier node, LineariserOptions options)
           weight = 1;
         }
 
-        var hasAttrs = node.Triggers.Count > 0 || info.qid != null || weight != 1 || info.uniqueId != -1;
```
keyboardDrummer, line 450: Is it possible to have a test for this fix?

keyboardDrummer (approved):

shazqadeer:
Let us either land or abandon this approved PR.

atomb:
Can you merge it, @shazqadeer? I don't have permissions to do so.

PR #580, opened 2022-05-26
Handle hard timeout gracefully in batch solver

atomb:
This is much more difficult to do when the solver is in interactive mode, because a timeout kills the whole process.

### DIFF
--- a/Source/Provers/SMTLib/SMTLibBatchTheoremProver.cs
+++ b/Source/Provers/SMTLib/SMTLibBatchTheoremProver.cs
@@ -163,6 +163,9 @@ namespace Microsoft.Boogie.SMTLib
     private async Task<Outcome> GetResponse(CancellationToken cancellationToken)
     {
       var outcomeSExp = await Process.GetProverResponse().WaitAsync(cancellationToken);
+      if (outcomeSExp.Name.Equals("timeout")) {
+        return Outcome.OutOfResource;
+      }
       var result = ParseOutcome(outcomeSExp, out var wasUnknown);
 
       var unknownSExp = await Process.GetProverResponse().WaitAsync(cancellationToken);


Source/Provers/SMTLib/SMTLibBatchTheoremProver.cs
```diff
@@ -163,6 +163,9 @@ public override async Task<Outcome> CheckOutcome(ErrorHandler handler, int error
     private async Task<Outcome> GetResponse(CancellationToken cancellationToken)
     {
       var outcomeSExp = await Process.GetProverResponse().WaitAsync(cancellationToken);
+      if (outcomeSExp.Name.Equals("timeout")) {
```
keyboardDrummer, line 166: Can we have a unit test for this?

keyboardDrummer (approved):

Source/Provers/SMTLib/SMTLibBatchTheoremProver.cs
atomb: Yeah, we can. I should have left this as a draft until I added one. :)

Test/prover/z3-hard-timeout.bpl
```diff
@@ -0,0 +1,11 @@
+// RUN: %boogie /proverOpt:C:"-T:1" /proverOpt:BATCH_MODE=true /trace "%s" > "%t"
```
keyboardDrummer, line 1: Is `T` a timeout option? Might be nice to use the unabbreviated version for readability.

keyboardDrummer (dismissed):

Test/prover/z3-hard-timeout.bpl
atomb: Yeah, it is a timeout option. I'm not sure it has a long form, actually. There is a `timeout` parameter, but I _think_ that corresponds to the "soft" timeout set with `-t` (lowercase).

keyboardDrummer (approved):

PR #585, opened 2022-06-01
Enable rerunning a cancelled ImplementationTask

keyboardDrummer:
### Changes

- `Task<VerificationResult> ActualTask`, `VerificationStatus CurrentStatus` and `IObservable<VerificationStatus>` jave been turned into `IVerificationStatus CacheStatus` and `IObservable<IVerificationStatus>`, which leads to cleaner code on the client side.
- Enable calling `IImplementationTask.Run` after cancelling a previous run.

### Testing
Added a test `RunCancelRun`

### DIFF
--- a/Source/ExecutionEngine/ExecutionEngine.cs
+++ b/Source/ExecutionEngine/ExecutionEngine.cs
@@ -696,7 +696,7 @@ namespace Microsoft.Boogie
 
     /// The outer task is to wait for a semaphore to let verification start
     /// The inner task is the actual verification of the implementation
-    public async Task<Task<VerificationResult>> EnqueueVerifyImplementation(
+    public Task<Task<VerificationResult>> EnqueueVerifyImplementation(
       ProcessedProgram processedProgram, PipelineStatistics stats,
       string programId, ErrorReporterDelegate er, Implementation implementation,
       CancellationTokenSource cts,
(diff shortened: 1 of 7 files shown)


Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -707,17 +707,33 @@ private Implementation[] GetPrioritizedImplementations(Program program)
...
+      });
+      return result;
+    }
+
+    /// The outer task is to wait for a semaphore to let verification start
+    /// The inner task is the actual verification of the implementation
+    public async Task<Task<VerificationResult>> EnqueueVerifyImplementation(
+      ProcessedProgram processedProgram, PipelineStatistics stats,
+      string programId, ErrorReporterDelegate er, Implementation implementation,
+      CancellationToken cancellationToken,
```
keyboardDrummer, line 724: Was needed because the upper one requires a `CancellationTokenSource` and this one only has a `CancellationToken`.

Source/VCGeneration/VCGen.cs
```diff
@@ -356,18 +366,26 @@ void ExpandAsserts(Implementation impl)
       watch.Start();
 #endif
 
-      ConvertCFG2DAG(run.Implementation);
+      var data = implementationData.GetOrCreateValue(run.Implementation)!;
+      if (!data.ConvertedToDAG) {
+        data.ConvertedToDAG = true;
```
atomb, line 371: I don't know how much it matters, but this sets `convertedToDAG` to `true` before converting it, whereas `Passified`, below, is the other way around.

atomb (approved):
I think this looks good, from what I can tell, though it's the sort of change I'd be happier to see pass CI before saying anything conclusive about. The CI seems to have hung in some way.

The two comments I made are trivial, so feel free to make changes in response, or not, as you like.

atomb (dismissed):

PR #586, opened 2022-06-08
Add support for a `verboseName` attribute

atomb:
This attribute changes the name Boogie uses when printing messages, choosing procedures to verify with `/proc`, and generating XML output. It also changes the name recorded in `VerificationResult` objects, as may be used by clients of Boogie as a library.


Source/Core/Absy.cs
```diff
@@ -1307,6 +1307,13 @@ void ObjectInvariant()
...
     }
 
+    public string /*!*/ DisplayName
+    {
+      get {
+        return QKeyValue.FindStringAttribute(Attributes, "displayName") ?? Name;
+      }
+    }
+
     public string /*!*/ Name
```
keyboardDrummer, line 1317: What's the purpose of name, if not to display things?

atomb: The value of `Name` is used in Boogie source files, and conforms to the naming restrictions on Boogie identifiers. The value of `DisplayName` can be anything, and even include spaces or other things that would be invalid in Boogie names.

Source/Core/Absy.cs
```diff
@@ -1307,6 +1307,13 @@ void ObjectInvariant()
       Contract.Invariant(name != null);
     }
 
+    public string /*!*/ DisplayName
```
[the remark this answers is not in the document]
atomb: Yeah, that's probably a bit better. I'll change it. I originally was considering having a setter, but ultimately found it wasn't necessary.

Source/Core/Absy.cs
keyboardDrummer: Do you that when parsing a Boogie file, Name has certain restrictions. Why would these restrictions also apply when setting name programmatically?

keyboardDrummer: Fair, then I wonder what the right name for DisplayName is, since it's not always used to display the AST node. How do I know when to use which?

atomb: It's now `VerboseName`, with a comment clarifying what it's intended to be used for.

bkragl: There is an attribute help [here](https://github.com/boogie-org/boogie/blob/7b389a524fe1cb475108688cb97b3a9ea1e7e639/Source/ExecutionEngine/CommandLineOptions.cs#L1699). Probably worth documenting the new attribute there too.

atomb: Oh, yes, good point. I'll open another PR to add that.

PR #587, opened 2022-06-08
Fixed: Concurrency problem on the Process

MikaelMayer:
Here is a bug that @cpitclaudel  and I identified, when working with Dafny Language Server
1) A verification request is sent to the SMTLibProcess
2) The user press "space" before the verification request is completed

The problem is that the process output can be ignored and never read.

However, because the Dafny Language Server often immediately sends a new verification request, and that this verification requests tries to use the same Checker and thus the same Z3 process, the first "PingPong" request fails because, after sending "ping", it collects the result that was never read in 1)

Moreover, a single proverErrors.Add() and then everything becomes inconclusive, as in this screenshot below:
image

Note that, if we used the same Ping/Pong pair for PingPong (where it is now time-sensitive) than for the other places (where it can be cancelled using a CancellationToken), we run into the problem that other places could send a Ping, get cancelled, and then on process recovery PingPong sends a Ping and reads their Pong, and assumes it's ok, but there remains an unread Pong

Thus, this PR solves this problem once and forall by the following change
* PingPong now requires a timeout
* If the time expires, it throws a ProverDiedException.
* PingPong uses a Ping2/Pong2 scheme (based on version rather than name)
* Other places that use Ping/Pong to get all the output now use Ping1/Pong1 so that it won't interfere with PingPong.

I also refactored other places that use Ping1/Pong1 so that their logic is centralized into `ReadOutputAsync`

This is not the end of the journey. I still get these kinds of errors when debugging why sometimes the language server crashes:
image

TODO: I still need to create some tests.... but how?

### DIFF
--- a/Source/Provers/SMTLib/SMTLibInteractiveTheoremProver.cs
+++ b/Source/Provers/SMTLib/SMTLibInteractiveTheoremProver.cs
@@ -517,62 +517,68 @@ namespace Microsoft.Boogie.SMTLib
       }
 
       SendThisVC("(get-model)");
-      Process.Ping();
-      Model theModel = null;
+
+      Model model = await ReadOutputAsync<Model>(
+        null,
+        (resp, theModel) => {
+          if (theModel != null)
+          {
+            HandleProverError("Expecting only one model but got many");
+          }
+
+          return ParseErrorModel(resp);
+        },
+        cancellationToken);
+
+      return model;
+    }
+    
+    /// <summary>
+    /// Sends a ping, and until the result is a pong,
+    /// passes intermediate results to collector
+    /// Returns the final result returned by collector, or the default value
+    /// </summary>
+    private async Task<T> ReadOutputAsync<T>(T defaultValue, Func<SExpr, T, T> collector, CancellationToken cancellationToken) {
+      Process.Ping1();
+
+      T result = defaultValue;
+
       while (true)
       {
         var resp = await Process.GetProverResponse().WaitAsync(cancellationToken);
-        if (resp == null || Process.IsPong(resp))
+        if (resp == null || Process.IsPong1(resp))
         {
           break;
         }
 
-        if (theModel != null)
-        {
-          HandleProverError("Expecting only one model but got many");
-        }
-
-        theModel = ParseErrorModel(resp);
+        result = collector(resp, result);
       }
 
-      return theModel;
+      return result;
     }
 
     private async Task<Outcome> GetResponse(CancellationToken cancellationToken)
     {
-      var result = Outcome.Undetermined;
       var wasUnknown = false;
 
-      Process.Ping();
-
-      while (true)
-      {
-        var resp = await Process.GetProverResponse().WaitAsync(cancellationToken);
-        if (resp == null || Process.IsPong(resp))
-        {
-          break;
-        }
-
-        result = ParseOutcome(resp, out wasUnknown);
-      }
+      var result = await ReadOutputAsync(
+        Outcome.Undetermined,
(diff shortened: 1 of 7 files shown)


Source/Provers/SMTLib/SMTLibInteractiveTheoremProver.cs
```diff
@@ -99,7 +101,7 @@ public override async Task BeginCheck(string descriptiveName, VCExpr vc, ErrorHa
 
       if (Process != null)
       {
-        await Process.PingPong(); // flush any errors
+        await Process.PingPong(MS_BEFORE_ASSUMING_PROVER_DIED); // flush any errors
         Process.NewProblem(descriptiveName);
```
cpitclaudel, line 105: Shouldn't you catch errors here?

Source/Provers/SMTLib/SMTLibSolver.cs
```diff
@@ -16,43 +16,57 @@ public abstract class SMTLibSolver
...
-  public void Ping()
+  public void Ping1()
   {
     Send("(get-info :name)");
   }
+  
+  public void Ping2()
+  {
+    Send("(get-info :version)");
+  }
```
cpitclaudel, line 27: Not a fan of pong1 / pong2.  Can we give `ping` an argument saying what to send / look for?

cpitclaudel: Maybe we can keep a single `Ping` but use `echo` with a counter instead of these `get-info` queries

Source/Provers/SMTLib/SMTLibInteractiveTheoremProver.cs
```diff
@@ -514,62 +517,68 @@ private async Task<Model> GetErrorModel(CancellationToken cancellationToken)
...
-      {
-        var resp = await Process.GetProverResponse().WaitAsync(cancellationToken);
-        if (resp == null || Process.IsPong(resp))
-        {
-          break;
-        }
-
-        result = ParseOutcome(resp, out wasUnknown);
-      }
+      var result = await ReadOutputAsync(
```
cpitclaudel, line 564: nice generalization but maybe we don't need to pass in a lambda; could we just return the SExpr instead?

cpitclaudel (changes requested):

keyboardDrummer (commented):
> The problem is that the process output can be ignored and never read.

Is that so? I think `SMTLibInteractiveTheoremProver` is designed to always complete read operations even when a cancellation occurs, although there might be a bug. However, I do think that there can be a race condition where different read operations can read each other's results from the solver, because `SMTLibInteractiveTheoremProver.GoBackToIdle` can read results from ongoing operations.

I think we should change `SMTLibProcess` so it has a `Task<SExpr> SendRequest(string input)` method instead of `Task<SExpr> GetProverResponse()` that ensures the returned result came from the sent input.

keyboardDrummer:
Related PR: https://github.com/boogie-org/boogie/pull/588

Source/Provers/SMTLib/SMTLibSolver.cs
MikaelMayer: Nice, thanks for the pointer to the `echo` command in smt-lib ! it made the code nicer

Source/Provers/SMTLib/SMTLibInteractiveTheoremProver.cs
cpitclaudel: Maybe a lot of sexprs, then?

PR #591, opened 2022-06-14
Add IsRunning property

keyboardDrummer:


Source/Provers/SMTLib/SMTLibProcess.cs
```diff
@@ -133,12 +133,15 @@ public override void Send(string cmd)
...
-    // TODO, consider building in a PingPong to catch a missing response.
     public override async Task<SExpr> SendRequest(string request) {
       SExpr previousResponse = null;
       try {
         await asyncLock.WaitAsync();
         Send(request);
+        // Because Z3 may return a response multiple times for a single request,
+        // We use a ping/pong to determine when Z3 has finished sending responses.
+        // We assume the last response is the correct one, although in practice
+        // We've seen the responses are duplicates.
```
MikaelMayer, line 144: in practice, not only we have seen duplicate responses, but also sometimes z3 emits an error before returning the response.

MikaelMayer (approved):
Makes sense

PR #596, opened 2022-06-27
Feat: Restart process if prover errors, and clear error messages

MikaelMayer:
In Dafny, it happened that a prover errors in the interactive theorem prover results in a a checker always returning Undetermined, until the executable is restarted.
This PR makes it possible, on a new interactive request, that there is no prover error. If there are some, the prover is restarted.

### DIFF
--- a/Test/civl/inductive-sequentialization/ProducerConsumer.bpl
+++ b/Test/civl/inductive-sequentialization/ProducerConsumer.bpl
@@ -1,17 +1,28 @@
 // RUN: %parallel-boogie "%s" > "%t"
 // RUN: %diff "%s.expect" "%t"
 
-type {:linear "pid"} Pid = int;
-var {:layer 0,3} C:[int]int; // FIFO channel
-var {:layer 0,3} head:int;   // head index of the channel
-var {:layer 0,3} tail:int;   // tail index of the channel
+type {:datatype} Channel; // FIFO channel
+function {:constructor} Channel(C: [int]int, head: int, tail: int): Channel;
 
-const unique prod_id:int;
-const unique cons_id:int;
+type ChannelId; // id of a FIFO channel
+
+type {:linear "cid"} {:datatype} ChannelHandle; // permission for sending to or receiving from a channel
+function {:constructor} Send(cid: ChannelId): ChannelHandle;
+function {:constructor} Receive(cid: ChannelId): ChannelHandle;
+
+function {:inline} Cid(handle: ChannelHandle): ChannelId {
+  if is#Send(handle) then cid#Send(handle) else cid#Receive(handle)
+}
+
+function {:inline} {:linear "cid"} ChannelIdCollector(cid: ChannelId) : [ChannelHandle]bool {
+  MapConst(false)[Send(cid) := true][Receive(cid) := true]
+}
+
+var {:layer 0,3} channels: [ChannelId]Channel; // pool of FIFO channels
 
 type {:pending_async}{:datatype} PA;
-function {:constructor} PRODUCER(x:int, pid:int) : PA;
-function {:constructor} CONSUMER(x:int, pid:int) : PA;
+function {:constructor} PRODUCER(x: int, send_handle: ChannelHandle) : PA;
+function {:constructor} CONSUMER(x: int, receive_handle: ChannelHandle) : PA;
 
 function {:inline} NoPAs () : [PA]int
 { (lambda pa:PA :: 0) }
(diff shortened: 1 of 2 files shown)


Source/Provers/SMTLib/SMTLibInteractiveTheoremProver.cs
```diff
@@ -41,6 +41,19 @@ public override Task GoBackToIdle()
...
 
+    public void ClearProverErrors()
+    {
+      lock (proverErrors)
+      {
+        if (proverErrors.Count > 0)
+        {
+          proverWarnings.RemoveAll(x => true);
+          proverErrors.RemoveAll(x => true);
+          SetupProcess();
```
atomb, line 52: Would it make sense to, instead, set `processNeedsRestart` to `true` here so that it gets restart in `PossiblyRestart` which is called a few lines below? That would guarantee that the restarted process would always receive the text in `common`.

keyboardDrummer (commented):
Looks good, +1'd Arron's comment.

Source/Provers/SMTLib/SMTLibInteractiveTheoremProver.cs
```diff
@@ -41,6 +41,19 @@ public override Task GoBackToIdle()
       return Process.PingPong();
     }
 
+    public void ClearProverErrors()
+    {
+      lock (proverErrors)
```
keyboardDrummer, line 46: Since `proverErrors` and `proverWarnings` are always locked before they're modified, we might as well use a concurrent data structure.

MikaelMayer: In think this is equivalent, but in either case I'm only worried about the following scenario, in which we empty the list, release the lock, and then the previous process adds one more error before we restart the process.

MikaelMayer: To avoid the problem I mention, I am attaching the proverErrors and proverWarnings to the Process field through a ConditionalWeakTable. 

Moreover, I changed their types to ConcurrentStack so that there is no need for locking mechanisms; however, when they are invoked once in a conditional statement, and once in the body of that conditional statement, I ensured that it's the same object we are considering instead of accessing the field twice.

Can you have a look to my new solution?

Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
```diff
@@ -26,8 +25,37 @@ public abstract class SMTLibProcessTheoremProver : ProverInterface
     protected TypeAxiomBuilder CachedAxBuilder;
     internal abstract ScopedNamer Namer { get; }
     protected TypeDeclCollector DeclCollector;
-    protected readonly List<string> proverErrors = new();
-    protected readonly List<string> proverWarnings = new();
+
+    private readonly ConditionalWeakTable<SMTLibSolver, ConcurrentStack<string>> allProverErrors = new();
```
keyboardDrummer, line 29: Alternatively, you could change move `HandleProverError` into a new class `ProverProblemCollector` that contains the errors and warnings, and create a new collector each time you create a new process:

```csharp
    protected void SetupProcess()
    {
      Process?.Close();
      Process = options.Solver == SolverKind.NoOpWithZ3Options
        ? new NoopSolver()
        : new SMTLibProcess(libOptions, options);
      this.proverProblemCollector = new ProverProblemCollector(this);
      Process.ErrorHandler += proverProblemCollector.HandleProverError;
    }
```

MikaelMayer: I like this better. I'm not fan of conditional weak tables, so this looks like a good solution.

Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
```diff
@@ -1564,6 +1524,59 @@ public override void Push()
       SendThisVC("(push 1)");
       DeclCollector.Push();
     }
+    
+    
+  
+  
+    public class ProverProblemCollector
```
keyboardDrummer, line 1531: Could you move this class to a separate file? This file is already very large.

Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
```diff
@@ -527,14 +529,19 @@ protected void FlushProverWarnings()
...
+        while (proverProblems.Warnings.TryDequeue(out var warning))
         {
-          proverWarnings.Iter(handler.OnProverWarning);
-          proverWarnings.Clear();
+          handler.OnProverWarning(warning);
         }
       }
     }
 
+    protected void HandleProverError(string s)
```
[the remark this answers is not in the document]
MikaelMayer: Yes, it is used in SMTErrorModelConverter.
I added this helper, otherwise, I had to change every call there to `ProblemCollector.HandleProverErrors`

Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
MikaelMayer: Warnings are enqueued in the case of a parse error until they are displayed when an error occurs (in the case of an error). For example, in [`ParseRCount`]() within CheckSat
https://github.com/boogie-org/boogie/blob/7b389a524fe1cb475108688cb97b3a9ea1e7e639/Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs#L1544-L1558
warnings are processed in `HandleProverError`: 
https://github.com/boogie-org/boogie/blob/7b389a524fe1cb475108688cb97b3a9ea1e7e639/Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs#L575

For errors, you're right, they are read nowhere except for their `.Count > 0` property.
So I made the field `Errors` private and added the public property `HadErrors` which reads nicely. I think it's not too costly to keep the errors in the concurrent stack anyway and it had been already useful for me in the past.

keyboardDrummer: Good insight, but `FlushProverWarnings()` is also called if there's no error right? It is called _before_ the check `if (error == "") return;`

Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
```diff
@@ -1564,6 +1524,60 @@ public override void Push()
...
       DeclCollector.Push();
     }
+    
+    
+  
+  
+    public class ProverProblemCollector
+    {
+      public ConcurrentQueue<string> Warnings = new();
+      private readonly ConcurrentStack<string> errors = new();
```
atomb, line 1534: I'm not sure I understand why this one's a stack instead of a queue. Just calling that out in case it's a mistake. If it's intentional, I'm otherwise happy with this PR.

atomb (approved):
Nice robustness improvement!

Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
MikaelMayer: It seems that the `proverWarnings` are flushed using an `ErrorHandler handler` that is provided when `CheckSat` is called in either `SMTLibBatchTheoremProver` or `SMTLibInteractiveTheoremProver`

Removing `proverWarnings` by immediately using `currentErrorHandler` might result in nothing if `currentErrorHandler` is not set at best, and in a null pointer exception if we don't check this and it's null.

There is no unit test guarding us against failures of the underlying SMT solver. I implemented the change in the last commit, but I would like you to review it to ensure it is coherent.

keyboardDrummer (approved):

Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
```diff
@@ -1522,7 +1522,7 @@ protected int ParseRCount(SExpr resp)
       {
         // If anything goes wrong with parsing the response from the solver,
         // it's better to be able to continue, even with uninformative data.
-        currentErrorHandler.OnProverWarning($"Failed to parse resource count from solver. Got: {resp}");
+        currentErrorHandler?.OnProverWarning($"Failed to parse resource count from solver. Got: {resp}");
```
MikaelMayer, line 1525: I still have a bad feeling about not stacking warnings before  an output handler is available. A nullpointerexception here would have meant that we are dismissing this warning. This is not desirable. All warnings should make their way to the user. That's probably why they were "flushed", before, when we are sure that there is a currentErrorHandler available. Could you perhaps use an async queue for that purpose?

Source/Provers/SMTLib/SMTLibBatchTheoremProver.cs
```diff
@@ -97,7 +97,7 @@ public override int FlushAxiomsToTheoremProver()
 
         var result = CheckSat(cancellationToken);
```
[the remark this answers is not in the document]
MikaelMayer: You're right, it is existing behavior. And it looks weird.
Looking at the control flow, the CheckSat command is supposed to close the input after sending all the commands, synchronously, so the pop will never make it to the process's input. But it will make it to the log.
So I guess it's ok, although I'm not sure why it's necessary.

PR #601, opened 2022-07-15
add field access syntax for datatypes

shazqadeer:
This PR adds "->" as the token for accessing fields in datatypes. 

I recommend reading Test\datatypes\field_access_*.bpl, the newly-added tests to get an idea of the new syntax and backward compatibility with old syntax. 

Next, I recommend looking at the new classes FieldAccess and FieldAssignLhs. The latter is similar to MapAssignLhs and allows datatype values to be updated by specifying a single field access as an assignment target.

Finally, I recommend reading the changes in BoogiePL.atg.


shazqadeer:
@RustanLeino: 

Thank you for the email exchange earlier about this design. The implementation in this PR differs from what we discussed. First, I chose to implement something specific for datatypes rather than general function calls. The reason is that I wanted to allow the field name after the `->` token to refer to the native field name rather than the `#`-decorated datatype selector name. I would much appreciate especially your feedback on this PR.

I thought about using the token `.` instead of `->` but that does not work because `.` is a valid character inside an identifier. But if we choose to outlaw `.` inside identifiers, then we could use it for the field access which would be nicer than `->`.

shazqadeer:
I was curious about the possibility of using `.` for field access. I disabled `.` in identifiers and ran the regressions. About 30 tests failed. So it is not uncommon. I suppose it would be undesirable to break clients and disallow `.` in identifiers.  Thoughts?

I am done with whatever I wanted to achieve in this PR. I am ready to land but would love to get feedback.

Source/Core/AbsyExpr.cs
```diff
@@ -3837,6 +3843,197 @@ public T Dispatch<T>(IAppliableVisitor<T> visitor)
...
+        return this._tok;
+      }
+      set
+      {
+        Contract.Requires(value != null);
+        this._tok = value;
+      }
+    }
+    
+    public string fieldName
```
RustanLeino, line 3868: It is typical in C# to use an initial capital letter for properties. I don't remember what the rest of the Boogie source code does, but I propose change `tok`, `fieldName`, and `selector` to `Tok`, `FieldName`, and `Selector`.

Source/Core/AbsyExpr.cs
```diff
@@ -3837,6 +3843,197 @@ public T Dispatch<T>(IAppliableVisitor<T> visitor)
...
+    {
+      return 1;
+    }
+
+    public void Emit(IList<Expr> args, TokenTextWriter stream, int contextBindingStrength, bool fragileContext)
+    {
+      stream.SetToken(this);
+      Contract.Assert(args.Count == 1);
+      stream.push();
+      stream.Write("(");
```
RustanLeino, line 3949: The `contextBindingStrength` parameter can be used to let you omit these parentheses. I think something like that is done with the `...[...]` expression.

Test/datatypes/t4.bpl.expect
```diff
@@ -0,0 +1,2 @@
+
+Boogie program verifier finished with 3 verified, 0 errors
```
RustanLeino, line 2: It would be good to have some negative verification tests, too.

Source/Core/Absy.cs
```diff
@@ -1457,6 +1458,36 @@ public DatatypeTypeCtorDecl(TypeCtorDecl typeCtorDecl)
...
+    public void AddConstructor(DatatypeConstructor constructor)
+    {
+      this.constructors.Add(constructor);
+    }
+
+    public void AddSelector(DatatypeSelector selector)
+    {
+      if (!selectors.ContainsKey(selector.OriginalName))
+      {
+        selectors.Add(selector.OriginalName, new List<DatatypeSelector>());
```
RustanLeino, line 1473: Is this where `selector.OriginalName` is something like `a` and `selector.Name` is something like `a#Pair`. With the new `->` syntax, I suggest supporting only `OriginalName`.

Source/Core/AbsyExpr.cs
```diff
@@ -3837,6 +3843,197 @@ public T Dispatch<T>(IAppliableVisitor<T> visitor)
...
+      var selectors = datatypeTypeCtorDecl.GetSelectors(fieldName);
+      if (selectors == null)
+      {
+        tc.Error(this.tok, "datatype {0} does not have a field with name {1}", ctorType, fieldName);
+        return null;
+      }
+      Contract.Assert(selectors.Count > 0);
+      if (selectors.Count > 1)
+      {
+        tc.Error(this.tok, "datatype {0} has several fields with name {1}", ctorType, fieldName);
```
RustanLeino, line 3996: Just as a point of comparison, Dafny allows a datatype to have several destructors with the same name, as long as they all have the same type. No need to change anything at this time, but if this further functionality is needed in the future, it could be supported in Boogie.

Also, in Dafny, expressions can be partial, so a precondition of such a field-access operation on a datatype is that the variant (constructor) of the datatype is one that has the requested field. Since Boogie is total, I imagine that Boogie always allows the field access, but sometimes acts like just an uninterpreted function.

RustanLeino (changes requested):
The new functionality and the `->` syntax look fine to me. I suggest allowing only one of `a` and `a#Pair` to follow the `->`.

On the subject of using `.` instead of `->`, Dafny makes heavy use of `.` in identifiers. For Dafny, this makes the fully qualified identifiers more readable than using some other "delimiter" character. I believe other verifiers (like AutoProof for Eiffel) also make use of `.` in identifier names.

Test/datatypes/t4.bpl
```diff
@@ -0,0 +1,42 @@
...
+function{:constructor} Pair(a: int, b: int): Pair;
+
+procedure P0(p: Pair) returns (q: Pair)
+  requires p->a#Pair == 0;
+  ensures  q->a#Pair == 1;
+{
+  assert p->a == p->a#Pair;
+  q := p;
+  q->a#Pair := 1;
+  assert q->b == p->b#Pair;
```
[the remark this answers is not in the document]
shazqadeer: b#Pair is the name of the selector function automatically generated for the field b. I was allowing the name of the selector function to be used in a field access as well. I have now disallowed it. See the revised t4.bpl.

Source/Core/AbsyExpr.cs
shazqadeer: I have made the changes except for `Tok`. I see that lower case `tok` is used for the property consistently across all other equivalent classes.

Source/Core/AbsyExpr.cs
shazqadeer: Done. I am using the same opBinding strength for `->` as for `...[...]`. 

```
const int opBindingStrength = 0x90;
```

Sound good?

Source/Core/Absy.cs
shazqadeer: Fixed now.

Source/Core/AbsyExpr.cs
shazqadeer: The second paragraph in your comment above is accurate.

Test/datatypes/t4.bpl.expect
shazqadeer: Added t7.bpl.

Test/datatypes/t4.bpl
```diff
@@ -0,0 +1,42 @@
+// RUN: %parallel-boogie "%s" > "%t"
+// RUN: %diff "%s.expect" "%t"
+
+type{:datatype} Pair;
```
keyboardDrummer, line 4: What's the benefit of `type{:datatype} Pair` over `datatype Pair` ?

Test/datatypes/t4.bpl
```diff
@@ -0,0 +1,42 @@
...
+function{:constructor} Pair(a: int, b: int): Pair;
```
keyboardDrummer, line 5: What's the benefit of `function{:constructor} Pair(a: int, b: int): Pair;` over `constructor Pair(a: int, b: int): Pair;` ?

Source/Core/AbsyExpr.cs
```diff
@@ -3837,6 +3843,206 @@ public T Dispatch<T>(IAppliableVisitor<T> visitor)
...
+
+    private string fieldName;
+    
+    private DatatypeSelector selector;
+
+    public IToken tok
+    {
+      get
+      {
+        Contract.Ensures(Contract.Result<IToken>() != null);
```
keyboardDrummer, line 3858: Are these contracts useful? They don't run at compile-time. I would prefer not using them and then this `tok` becomes a regular auto-property `public IToken tok {get;set;}`

Source/Core/AbsyCmd.cs
```diff
@@ -2345,6 +2345,106 @@ public override Absy StdDispatch(StandardVisitor visitor)
...
+    public override IdentifierExpr DeepAssignedIdentifier
+    {
+      get
+      {
+        Contract.Ensures(Contract.Result<IdentifierExpr>() != null);
+        return Datatype.DeepAssignedIdentifier;
+      }
+    }
+
+    public override Variable DeepAssignedVariable
```
keyboardDrummer, line 2379: Consider `public override Variable DeepAssignedVariable => Datatype.DeepAssignedVariable;`

Source/Core/AbsyCmd.cs
```diff
@@ -2345,6 +2345,106 @@ public override Absy StdDispatch(StandardVisitor visitor)
...
+    {
+      Contract.Invariant(Datatype != null);
+      Contract.Invariant(FieldAccess != null);
+    }
+
+    public TypeParamInstantiation TypeParameters = null;
+
+    private Type TypeAttr = null;
+
+    public override Type Type
```
keyboardDrummer, line 2365: Consider `public override Type Type {get; private set;}`

Test/datatypes/t4.bpl
```diff
@@ -0,0 +1,42 @@
+// RUN: %parallel-boogie "%s" > "%t"
```
keyboardDrummer, line 1: The names of the test files don't have any meaning, so they might as well be a single file.

Test/datatypes/t4.bpl
shazqadeer: The syntax for specifying a datatype is not introduced by this PR. It has existed for a while. The original introduction of datatypes in Boogie was experimental with no investment in syntax. So everything was done by using attributes on existing syntax. What you are seeing is a result of that design. It is desirable to have a nice clean syntax for specifying datatypes. Perhaps the subject of another PR.

Test/datatypes/t4.bpl
shazqadeer: See my response to your first comment.

Multiple constructors are allowed and used. There are many such tests in the civl folder.

Source/Core/AbsyExpr.cs
shazqadeer: They are all over the place in Boogie code. I was trying to be consistent. 

@RustanLeino : Could you shed some light? Should we try to drop these contracts gradually?

Test/datatypes/t4.bpl
shazqadeer: I understand. But separate files were helpful to me organizing the nature of different tests. So I would prefer to leave them separate. Would it be helpful if I added a comment in each file about the nature of the tests in it?

keyboardDrummer: Could you give the files more specific names?

Test/datatypes/t4.bpl
keyboardDrummer: Sounds good, thanks :)

Source/Core/AbsyExpr.cs
```diff
@@ -3837,6 +3843,206 @@ public T Dispatch<T>(IAppliableVisitor<T> visitor)
...
+      bool parensNeeded = opBindingStrength < contextBindingStrength ||
+                          (fragileContext && opBindingStrength == contextBindingStrength);
+      stream.SetToken(this);
+      Contract.Assert(args.Count == 1);
+      stream.push();
+      if (parensNeeded)
+      {
+        stream.Write("(");
+      }
+      cce.NonNull(args[0]).Emit(stream, 0x00, false);
```
RustanLeino, line 3956: The recursive call should pass in `opBindingStrength`, not `0x00`.

Suppose that an operator like `+` (whose `opBindingStrength` is `0x40`) could produce a value of a datatype, then the expression `(a + b)->x` would be printed as `a + b->x` if you pass in `0x00`. There aren't any such operators in Boogie today, but the correct parameter value is still `opBindingStrength` (which in your case is `0x90`).

Source/Core/AbsyExpr.cs
RustanLeino: You're right. `.tok` has been an exception this rule, and it seems okay to leave it. (Any change to this in the future would be appropriate for a separate PR.)

Source/Core/AbsyExpr.cs
RustanLeino: Yes, same binding strength as `...[...]` is good. But note my comment in the code--the recursive call should pass in the `opBindingStrength` you use, not `0x00`.

Source/Core/AbsyExpr.cs
RustanLeino: Yes, unfortunately, these contracts are no longer useful, since the tool that would do the bytecode rewriting for them is no longer available. `Contract.Assert` and `Contract.Assume` are the only ones that (in a DEBUG build) generate run-time checks.

I think that dropping them gradually is a good plan. Those that remain in the source code still serve as documentation, until they will eventually be removed.

Source/Core/AbsyCmd.cs
shazqadeer: Done.

Source/Core/AbsyCmd.cs
shazqadeer: I used `public override Type Type => TypeAttr;` instead. I was getting an error with your suggestion.

Source/Core/AbsyExpr.cs
shazqadeer: I have removed most of the contracts in the classes I added. I also converted to properties wherever I could.

Source/Core/AbsyExpr.cs
shazqadeer: Done.

Test/datatypes/t4.bpl
shazqadeer: Done.

Source/Core/AbsyExpr.cs
shazqadeer: Done.

RustanLeino (approved):

bkragl:
Great feature @shazqadeer, thanks for adding it! It makes Boogie programs with data types much nicer to read.

Issue #519, opened 2022-02-28
Boogie hangs if there is no z3 on Linux


anaumchev:
**Problem**: Boogie just hangs when executed with no command-line arguments if there is no z3 installed.
**Expected**: Boogie reports the missing dependency and terminates.

shazqadeer:
@anaumchev : I just committed a PR that changes how Boogie invokes the SMT solver. Could you please try again and see if your problem is resolved?

shazqadeer:
Closing for now. Please reopen if necessary.

Issue #483, opened 2022-01-16
Boogie should just try to execute z3 instead of looking for a .exe in PATH


cpitclaudel:
It looks like Boogie explicitly searches the PATH for `z3` or `z3.exe` instead of just trying to run `z3`.  The result is that it doesn't run if z3 is in your path under the name `z3.cmd`:

```
c:\Users\cpitcla>z3
Error: input file was not specified.
For usage information: z3 -h

c:\Users\cpitcla>boogie test2.dfy.bpl 
Fatal Error: ProverException: Cannot find any prover executable
```

Windows symlinks don't work very well, so having a .cmd script on your path that calls Z3 is the best alternative AFAIK.  Most applications seem fine with it; Boogie is an exception.

PR #602, opened 2022-07-18
Fixes issue #483

shazqadeer:
This PR eliminates the search for the prover binary. Now the user has to either supply ProverPath as a command-line option or have the solver binary present in the path. I verified that if the z3 executable in not present in my path, I get the following error message which should clue the user.

Fatal Error: ProverException: Unable to start the process z3: An error occurred trying to start process 'z3' with working directory '/Users/shaz/Source/boogie/Test/datatypes'. No such file or directory

It is possible that this PR will also fix issue #519 .


rakamaric (approved):
Looks good to me.
I do wonder though why the more complicated path-searching implementation was there in the first place. I would guess there was some use case when just invoking `z3` was not enough.

PR #605, opened 2022-07-20
Make TryRun thread safe

keyboardDrummer:
### Changes
- Make `IImpelementationTask.TryRun` thread-safe
- Add `IImpelementationTask.IsIdle`

### Testing
- Update tests to exercise new behavior.

### DIFF
--- a/Source/Directory.Build.props
+++ b/Source/Directory.Build.props
@@ -2,7 +2,7 @@
 
   <!-- Target framework and package configuration -->
   <PropertyGroup>
-    <Version>2.15.6</Version>
+    <Version>2.15.7</Version>
     <TargetFramework>net6.0</TargetFramework>
     <GeneratePackageOnBuild>false</GeneratePackageOnBuild>
     <Authors>Boogie</Authors>
--- a/Source/ExecutionEngine/ImplementationTask.cs
+++ b/Source/ExecutionEngine/ImplementationTask.cs
@@ -119,7 +119,7 @@ public class ImplementationTask : IImplementationTask {
       lock (mayAccessCancellationSource) {
         // Clear cancellationSource before calling status.OnCompleted, so ImplementationTask.IsIdle returns true
         cancellationSource = null;
-        if (cancellationToken.IsCancellationRequested) {
+        if (cancellationToken.IsCancellationRequested && CacheStatus is not Completed) {
           Console.WriteLine("publishing stale");
           status.OnNext(new Stale());
         } else {
--- a/Source/ExecutionEngine/ImplementationTask.cs
+++ b/Source/ExecutionEngine/ImplementationTask.cs
@@ -1,7 +1,9 @@
 #nullable enable
 using System;
 using System.IO;
+using System.Reactive.Linq;
 using System.Reactive.Subjects;
+using System.Reactive.Threading.Tasks;
 using System.Threading;
 using System.Threading.Tasks;
 
@@ -36,11 +38,13 @@ public interface IImplementationTask {
   Implementation Implementation { get; }
 
   IObservable<IVerificationStatus>? TryRun();
+  bool IsIdle { get; }
   void Cancel();
 }
 
 public class ImplementationTask : IImplementationTask {
   private readonly ExecutionEngine engine;
+  private readonly object myLock = new();
 
   public IVerificationStatus CacheStatus { get; private set; }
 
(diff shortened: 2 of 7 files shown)


Source/ExecutionEngine/ImplementationTask.cs
```diff
@@ -62,40 +72,59 @@ public class ImplementationTask : IImplementationTask {
...
   public IObservable<IVerificationStatus>? TryRun()
   {
     if (CacheStatus is Completed) {
       return null;
     }
 
-    var alreadyRunning = cancellationSource != null;
-    if (alreadyRunning) {
-      return null;
+    lock (myLock) {
```
MikaelMayer, line 89: Shouldn't the lock also contain the previous if statement?
Two threads could arrive at this lock at the same time and a third one sets `CacheStatus` as `Completed` before it enters the lock.
Then there could be race conditions.

MikaelMayer (commented):
Pending questions. Please explain me what you are achieving with enough detail so that I can understand your design.

Source/ExecutionEngine/ImplementationTask.cs
```diff
@@ -62,40 +72,78 @@ public class ImplementationTask : IImplementationTask {
...
+
+      cancellationSource = new();
     }
 
-    var alreadyRunning = cancellationSource != null;
-    if (alreadyRunning) {
+    // We claimed the right to run. No other thread is running nor can they get to this point,
+    // so from this point we can read and write from and to the fields CacheStatus and status as if we're the only thread.
+
+    if (CacheStatus is Completed) {
```
cpitclaudel, line 120: Could we do this part before creating the new `cancellationSource` above?

Source/ExecutionEngine/ImplementationTask.cs
```diff
@@ -62,40 +72,78 @@ public class ImplementationTask : IImplementationTask {
...
     task.ContinueWith(r =>
     {
-      if (r.Exception != null) {
-        observableStatus.OnError(r.Exception);
-      } else {
-        observableStatus.OnCompleted();
+      // Lock so we may do operations after clearing cancellationSource,
+      // which releases our control over the field status.
+      lock (myLock) {
+        // Clear cancellationSource before calling status.OnCompleted, so ImplementationTask.IsIdle returns false
```
cpitclaudel, line 137: `false` or `true`?

cpitclaudel (changes requested):
I left a few comments.  Happy to re-review after that, I think this is pretty close to approval.

Source/Directory.Build.props
```diff
@@ -2,7 +2,7 @@
 
   <!-- Target framework and package configuration -->
   <PropertyGroup>
-    <Version>2.15.6</Version>
+    <Version>2.15.7</Version>
```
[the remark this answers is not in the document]
keyboardDrummer: Bumps the Boogie version number. I do this on most of my Boogie PRs so I'm then also able to use the Boogie changes in Dafny.

Source/ExecutionEngine/ImplementationTask.cs
```diff
@@ -62,40 +72,78 @@ public class ImplementationTask : IImplementationTask {
   }
 
   private CancellationTokenSource? cancellationSource;
+  private ReplaySubject<IVerificationStatus>? status;
```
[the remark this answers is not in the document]
keyboardDrummer: status always ends in either Completed or Stale, and CacheStatus reflects that final value. If no run has been done, so no status has been set, CacheStatus is Stale.

Source/ExecutionEngine/ImplementationTask.cs
```diff
@@ -62,40 +72,78 @@ public class ImplementationTask : IImplementationTask {
...
+        return result;
+      }
+
+      cancellationSource = new();
     }
 
-    var alreadyRunning = cancellationSource != null;
-    if (alreadyRunning) {
+    // We claimed the right to run. No other thread is running nor can they get to this point,
+    // so from this point we can read and write from and to the fields CacheStatus and status as if we're the only thread.
```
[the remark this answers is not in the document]
keyboardDrummer: The large lock makes it harder to refactor, but I've made an attempt :)

Source/ExecutionEngine/ImplementationTask.cs
```diff
@@ -62,40 +72,93 @@ public class ImplementationTask : IImplementationTask {
...
+
+      if (cancellationSource.IsCancellationRequested) {
+        // Another thread is running but was cancelled,
+        // so we may immediately start a new run after the cancellation completes.
+        return QueueRun();
+      }
+
+      // Another thread is running and is not cancelled, so this run fails.
       return null;
     }
```
cpitclaudel, line 98: Does the whole method now run inside the lock?

Source/ExecutionEngine/ImplementationTask.cs
```diff
@@ -62,40 +72,93 @@ public class ImplementationTask : IImplementationTask {
...
+        }
       }
     }, TaskScheduler.Current);
-    return observableStatus;
+    return status;
+  }
+
+  private IObservable<IVerificationStatus>? QueueRun()
+  {
+
```
cpitclaudel, line 138:
```suggestion
```

cpitclaudel (approved):
Looks good to me.  The one thing I'm not 100% clear on is whether there would be a way to unify the two cases (enqueuing to wait for the cancelled task to exit vs starting a task right away).  I'm also not 100% sure that I understand whether the `status` variable is sufficiently protected: is there a chance that the status can change while we hold the lock, and would that cause an issue?

Source/ExecutionEngine/ImplementationTask.cs
keyboardDrummer: Jep

keyboardDrummer:
>  is there a chance that the status can change while we hold the lock, and would that cause an issue?

I don't think so, but it seems impossible to say for sure :-)

cpitclaudel (approved):

PR #606, opened 2022-07-20
Create GitHub releases alongside NuGet uploads

### LINKED ISSUE #572: Cut GitHub releases automatically when releasing
New releases of Boogie, indicated by pushing tags, are automatically pushed to Nuget by GitHub actions, which is super nice! Some consumers have been confused by the fact that there aren't corresponding GitHub releases as well, however. There used to be up until 2.8.22, and this makes it easy to not realize there are much newer versions on Nuget.

It should be very simple to publish a GH release from [this workflow](https://github.com/boogie-org/boogie/blob/master/.github/workflows/test.yml) - I set up the same thing for Dafny [here](https://github.com/dafny-lang/dafny/blob/master/.github/workflows/publish-release.yml#L67-L77), and in this case it can be published immediately rather than creating a draft, as there won't be any release notes to fill in.

atomb:
Fixes #572.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,6 +31,9 @@ jobs:
         uses: actions/checkout@v2
         with:
           fetch-depth: 0
+      - name: Determine version
+        id: get_version
+        uses: battila7/get-version-action@v2
       - name: Install tools, build Boogie, test Boogie
         run: |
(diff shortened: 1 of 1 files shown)


robin-aws (approved):
Nice, thanks for getting to this since I couldn't :)

Issue #371, opened 2021-04-13
[Civl] bug in inlining of atomic actions


shazqadeer:
For introduction actions, Civl automatically add the inline attribute (as we do for all atomic actions). This is done using AddAttribute in Absy.cs, which, if an attribute with the given name is already present, adds the given parameters to this existing attribute. So we end up with {:inline 1, 1} (as opposed to {:inline 1} {:inline 1}), which causes the inlining not to happen.

Civl should check (on all atomic actions) if an {:inline i} attribute is present and output an error if i != 1.  {:inline 1} should be added only if the inline attribute is absent.

shazqadeer:
@bkragl :  FYI

PR #610, opened 2022-08-02
[Civl] fixes issue 371

shazqadeer:
Fixes issue #371

### DIFF
--- a/Source/Concurrency/CivlTypeChecker.cs
+++ b/Source/Concurrency/CivlTypeChecker.cs
@@ -331,11 +331,6 @@ namespace Microsoft.Boogie
         {
           Error(proc, "unnecessary to provide inline attribute on action");
         }
-        var impl = actionProcToImpl[proc];
-        if (impl.FindAttribute("inline") != null)
-        {
-          Error(impl, "unnecessary to provide inline attribute on action");
-        }
       });
       
       if (checkingContext.ErrorCount > 0)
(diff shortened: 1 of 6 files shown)


Source/Concurrency/CivlTypeChecker.cs
```diff
@@ -324,7 +324,20 @@ private void TypeCheckActions()
...
+        if (proc.FindAttribute("inline") != null)
+        {
+          Error(proc, "unnecessary to provide inline attribute on action");
+        }
+        var impl = actionProcToImpl[proc];
+        if (impl.FindAttribute("inline") != null)
+        {
+          Error(impl, "unnecessary to provide inline attribute on action");
+        }
+      });
```
bkragl, line 339: I know that in Boogie procedures and implementations can be declared separately, but that if declared together (i.e., as a procedure with body) the set of supplied attributes is given to both. What I'm wondering is: does Boogie processes the inline attribute on both, or if it is meaningful on only one of them? If it is the latter, we could only do the corresponding check and avoid duplicate error messages.

bkragl (approved):
Just one minor comment below.

Source/Concurrency/CivlTypeChecker.cs
shazqadeer:
```
protected int GetInlineCount(Implementation impl)
    {
      Contract.Requires(impl != null);
      Contract.Requires(impl.Proc != null);

      string /*!*/
        procName = impl.Name;
      Contract.Assert(procName != null);
      if (recursiveProcUnrollMap.TryGetValue(procName, out var c))
      {
        return c;
      }

      c = -1; // TryGetValue above always overwrites c
      impl.CheckIntAttribute("inline", ref c);
      // procedure attribute overrides implementation
      impl.Proc.CheckIntAttribute("inline", ref c);

      recursiveProcUnrollMap[procName] = c;
      return c;
    }
```

The code above in Inline.cs examines the `inline` attribute. It seems that the attribute on the procedure overrides the one on the implementation. So I think I can avoid both the check and the injection of the inline attribute on the implementation.

PR #616, opened 2022-09-01
Ensure binary operations always have a token.

MikaelMayer:
Previously, binary operations created out of the blue did not have a token, which caused a problem for reporting some errors in Dafny.
Rather than manually adding these tokens in Dafny, I am adding a mechanism by which the token of a newly built BinaryExpr through the helper is by default the token of the first element.

### DIFF
--- a/Source/Core/AbsyExpr.cs
+++ b/Source/Core/AbsyExpr.cs
@@ -169,7 +169,7 @@ namespace Microsoft.Boogie
       Contract.Requires(e1 != null);
       Contract.Requires(e0 != null);
       Contract.Ensures(Contract.Result<NAryExpr>() != null);
-      return Binary(Token.NoToken, op, e0, e1);
+      return Binary(e0.tok ?? Token.NoToken, op, e0, e1);
     }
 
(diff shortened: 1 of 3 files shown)


Source/Core/AbsyExpr.cs
```diff
@@ -169,7 +169,7 @@ public static NAryExpr Binary(BinaryOperator.Opcode op, Expr e0, Expr e1)
       Contract.Requires(e1 != null);
       Contract.Requires(e0 != null);
       Contract.Ensures(Contract.Result<NAryExpr>() != null);
-      return Binary(Token.NoToken, op, e0, e1);
+      return Binary(e0.tok ?? Token.NoToken, op, e0, e1);
```
RustanLeino, line 172: This looks fine and the `??` is a good catch. If one wanted to go one step further, I suppose one could write

``` cs
e0.tok ?? e1.tok ?? Token.NoToken
```

RustanLeino (approved):

RustanLeino (approved):

Issue #583, opened 2022-05-29
Prover error when using /randomSeedIterations:n


shazqadeer:
Repro: run boogie on Test/inductive-sequentialization/PingPong.bpl as

```
path_to_boogie PingPong.bpl /randomSeedIterations:1
```

Lots of Prover errors are spewed out:

```
Prover error: line 63 column 221: unknown function/constant random1806478530@@0

Prover error: line 154 column 104: ambiguous constant reference, more than one constant with the same sort, use a qualified expression (as <symbol> <sort>) to disambiguate random413413821@@0

(0,0): Verification inconclusive (Civl_IS_choice_MAIN)
Prover error: Invalid PING response from the prover: unknown

Prover error: Invalid PING response from the prover: (:rlimit 378)

Prover error: line 178 column 221: unknown function/constant random1150535845
```

atomb:
It seems that this comes up with `/randomSeed`, as well. Normally, `/randomSeedIterations:1` shouldn't be any different from the default mode of operation, except that it does explicitly set the random seed if it's not already set.

keyboardDrummer:
If I use /proverLog then it creates 2 logs, the second one only contains references to certain renamed variables but not the function definitions for them, so that's why the solver is complaining. Would have to deep dive to understand why.

PR #617, opened 2022-09-03
Fixes issue #583

shazqadeer:
Closes issue #583 

Names of datatype-related functions have special meaning in SMT files. So, we should not rename them when randomizing symbol names.

### DIFF
--- a/Source/VCExpr/ScopedNamer.cs
+++ b/Source/VCExpr/ScopedNamer.cs
@@ -174,7 +174,11 @@ namespace Microsoft.Boogie.VCExprAST
       }
 
       var uniqueInherentName = NextFreeName(thing, inherentName);
-      if (boogieDeterminedNames.Contains(inherentName))
+      if (thing is DatatypeConstructor || thing is DatatypeMembership || thing is DatatypeSelector)
+      {
+        result = uniqueInherentName;
+      }
+      else if (boogieDeterminedNames.Contains(inherentName))
       {
         result = uniqueInherentName;
       }


keyboardDrummer:
> Names of datatype-related functions have special meaning in SMT files.

Sorry I don't understand. Is this a situation where the references to something are renamed, but the definition is not? If so, why is the definition not renamed?

Source/VCExpr/ScopedNamer.cs
```diff
@@ -174,7 +174,11 @@ public virtual string GetName(object thing, string inherentName)
       }
 
       var uniqueInherentName = NextFreeName(thing, inherentName);
-      if (boogieDeterminedNames.Contains(inherentName))
+      if (thing is DatatypeConstructor || thing is DatatypeMembership || thing is DatatypeSelector)
```
keyboardDrummer, line 177: This seems hard to maintain since it's adding specific code in a generic location. New AST constructs could easily forget to update this `if` if they need to be added. I'd love to suggest something but I can't without understanding well why these three should be here.

keyboardDrummer (approved):
I'm OK with merging, although I think we could also improve the maintainability of this change.

keyboardDrummer:
Thanks for the excellent explanation. I wasn't aware of `declare-datatypes` and the corresponding functions it enables.

> I am happy to incorporate any suggestions you might have for improving my fix.

I think adding a property `MayRename` to `NamedDeclaration` that returns false for DatatypeConstructor, DatatypeSelector, and DatatypeMembership, and true otherwise, and which is used in `ScopedNamer.GetName`, would be more maintainable.

And also, I guess `DatatypeConstructor` could be renamed, as long as `DatatypeSelector` and `DatatypeMembership` change the corresponding part of their name with it. Since achieving this might be a hassle, I think disallowing the rename of `DatatypeConstructor` is fine for now.

PR #622, opened 2022-09-29
[Civl] separate function call construction from resolution/typecheck

shazqadeer:
ExprHelper contains a method to create a function call. This method was doing resolution and typechecking on the returned expression even though (1) other expression generators were not and (2) a separate helper method for resolving and typechecking generated Absy's has been provided in CivlUtil.cs. As a result, the situation was confusing, at least for me. This PR eliminates the typechecking from the expression generation leaving it to the caller to make sure things are resolved properly.


bkragl (commented):
LGTM

PR #624, opened 2022-09-30
Add syntax for testing datatype constructor

shazqadeer:
This PR adds native Boogie syntax for testing which constructor constructs a datatype value. The syntax is:
```
expr is F
```

```expr``` is an arbitrary expression. ```is``` is the new keyword. ```F``` is an identifier expected to be the name of a constructor of the type of ```expr```.

This PR also generalizes the field access support via ```->``` to allow the same field name to be used in multiple constructors, provided the types are the same.

Finally, a few bugs in handling of FieldAssignLhs (when a field access appears on the left side of assignment) are fixed.

### CODE REVIEW

**Comment by shazqadeer** [comment]
Once this PR lands, I intend to eliminate the f# and is# syntax for field access and constructor test from Boogie. As far as I know, Civl and Move Prover are the only users of Boogie datatypes. 

@wrwg: Heads up...

**Comment by wrwg** [comment]
> Once this PR lands, I intend to eliminate the f# and is# syntax for field access and constructor test from Boogie. As far as I know, Civl and Move Prover are the only users of Boogie datatypes.
> 
> @wrwg: Just offering a heads up here.

What's the syntax for `f#C`?

**Comment by shazqadeer** [comment]
> What's the syntax for `f#C`?

```f#C(e)``` would become ```e->f```.

Please take a look at the examples in Test/datatypes folder.

**Comment by shazqadeer** [comment]
I am merging this PR. But please leave feedback when you find the time. I will track and fix in subsequent PRs.

File: Source/Core/BoogiePL.atg [hunk 1]
```diff
@@ -1217,19 +1217,25 @@ Power<out Expr/*!*/ e0>
...
-= (. Contract.Ensures(Contract.ValueAtReturn(out e) != null); IToken/*!*/ x;
+= (. Contract.Ensures(Contract.ValueAtReturn(out e) != null); IToken/*!*/ x; IsConstructor isConstructor;
      e = dummyExpr;
   .)
   ( "-"                        (. x = t; .)
     UnaryExpression<out e>     (. e = Expr.Unary(x, UnaryOperator.Opcode.Neg, e); .)
   | NegOp                      (. x = t; .)
     UnaryExpression<out e>     (. e = Expr.Unary(x, UnaryOperator.Opcode.Not, e); .)
-  | CoercionExpression<out e>
+  | CoercionExpression<out e> [IF(la.val == "is") IsConstructor<out x, out isConstructor> (. e = new NAryExpr(x, isConstructor, new List<Expr> { e }); .)]
```
RustanLeino, line 1227: This seems wrong to me. The effect is that `is` binds stronger than `!` (and `-`). So,

    !a is node

means

    !(a is node)

which is not what `!a is node` looks like to me. I suggest you undo the change on L1227 and instead do:

```
Power<out Expr/*!*/ e0>
= (.Contract.Ensures(Contract.ValueAtReturn(out e0) != null); IToken/*!*/ x; Expr/*!*/ e1; .)
  IsConstructor<out e0>
  [
    "**"           (. x = t; .)
    /* recurse because exponentation is right-associative */
    Power<out e1>  (. e0 = Expr.Binary(x, BinaryOperator.Opcode.Pow, e0, e1); .)
  ]
  .

IsConstructor<out Expr/*!*/ e0>
= (. Contract.Ensures(Contract.ValueAtReturn(out e0) != null); IToken/*!*/ x, id; .)
  UnaryExpression<out e0>
  [ "is"            (. x = t; .)
    Ident<out id>
    (. var isConstructor = new IsConstructor(id, id.val);
       e0 = new NAryExpr(x, isConstructor, new List<Expr> { e0 });
    .)
  ]
  .
```

That makes the binding power of `is` stronger than the binary `**` and weaker than the unary `!` and `-`. Then, also change `IsConstructor.Emit` in `AbsyExpr.cs` to use:

``` c#
      const int opBindingStrength = 0x68;
```

PR #629, opened 2022-10-10
Refactor some of the SMT solver interaction code

atomb:
This should be a pure refactoring, maintaining identical behavior. It will make it easier to implement additional interfaces to SMT solvers, including in plugins (where some relevant code was previously inaccessible).

It's not quite done yet, though, so it's in draft form for CI for the moment.


Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
```diff
@@ -149,22 +149,6 @@ protected static string Sanitize(string msg)
...
 
-    protected void SetupProcess()
-    {
-      Process?.Close();
-      Process = libOptions.CreateSolver(libOptions, options);
-      
-      Process.ErrorHandler += HandleProverError;
-    }
-
-    public override void Close()
```
shazqadeer, line 160: I don't understand the purpose of removing Close from here. It looks like an identical implementation has been moved to the derived classes.

Source/Provers/SMTLib/SMTLibProcess.cs
```diff
@@ -314,182 +316,12 @@ private void HandleError(string msg)
       ErrorHandler?.Invoke(msg);
     }
 
-    #region SExpr parsing
```
shazqadeer, line 317: Glad that the parsing code is factored out. Thanks.

Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
atomb: Yep, that's intentional. The idea is that the two existing derived classes do it the same way but future derived classes wouldn't need to.

shazqadeer: Ok, but you could have achieved the same effect by leaving the Close method as is. This method is overridable and derived could change it if necessary. 

In any case, the issue is minor. Think a bit and do what you think is best. I will approve the PR.

shazqadeer (approved):

Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
atomb: Yeah, you're right about that. My preference is to require each subclass to implement it separately so that the superclass doesn't seem to prescribe a specific way to start and stop processes, even though it means slightly more code and some duplication. But it's definitely a matter of taste rather than correctness.

If you're happy with that, would you be up for merging it, @shazqadeer? I'm not yet in the right group to be able to do it myself.

PR #632, opened 2022-10-11
Rework tracing to exclude SMT setup times

### LINKED ISSUE #565: `/trace` prints incorrect time for first assertion in each file
Boogie's `/trace` seems to include some startup/warmup cost in the time of the first implementation in each file; for example, the following program…

```
procedure m1();
implementation m1() {
  assert true;
}

procedure m2();
implementation m2() {
  assert true;
}
```

… prints this:

```
Verifying m1 ...
  [0.149 s, 1 proof obligation]  verified

Verifying m2 ...
  [0.007 s, 1 proof obligation]  verified
```

This makes it hard to reliably find performance issues in a Boogie file.

cpitclaudel:
Closes #565.

This is not very polished but it seems to work for me.  I haven't had the time to test it with splits.

Output before:

```
Verifying m1 ...
[TRACE] Using prover: z3
  [0.345 s, solver resource count: 92, 1 proof obligation]  verified

Verifying m2 ...
  [0.062 s, solver resource count: 92, 1 proof obligation]  verified
```

Output after:

```
Verifying m1 ...
[TRACE] Using prover: z3
  [0.057 s, solver resource count: 92, 1 proof obligation]  verified

Verifying m2 ...
  [0.081 s, solver resource count: 92, 1 proof obligation]  verified
```


Source/VCGeneration/VCGen.cs
```diff
@@ -360,11 +360,6 @@ record ImplementationTransformationData {
       callback.OnProgress?.Invoke("VCgen", 0, 0, 0.0);
 
       Stopwatch watch = new Stopwatch();
```
atomb, line 362: Is this `Stopwatch` still used anywhere after removing the code below?

atomb (dismissed):
This looks good to me. I left one question, but I'm not sure whether it requires any changes.

Source/ExecutionEngine/VerificationResult.cs
```diff
@@ -20,6 +20,13 @@ public sealed class VerificationResult
   public DateTime Start { get; set; }
   public DateTime End { get; set; }
 
+  /// <summary>
+  /// Gets or sets a the elapsed time, excluding SMT solver setup costs.
```
shazqadeer, line 24: a the -> the

more narrow -> narrower

shazqadeer (dismissed):

Source/ExecutionEngine/VerificationResult.cs
cpitclaudel, line 24:
```suggestion
  /// Gets or sets the elapsed time, excluding SMT solver setup costs.
```

Source/ExecutionEngine/VerificationResult.cs
```diff
@@ -20,6 +20,13 @@ public sealed class VerificationResult
...
+  /// (This is more narrow than <c>Start - End</c>.)
```
cpitclaudel, line 25:
```suggestion
  /// (This is narrower than <c>Start - End</c>.)
```

Source/VCGeneration/VCGen.cs
cpitclaudel: Nope, it's a rebase gone wrong.  Thanks!

cpitclaudel, line 362:
```suggestion
```

cpitclaudel:
Thanks all for the reviews!  I'd like to wait for a review from @keyboardDrummer , since he recently modified this bit of the code (and it would be bad if we gave incorrect times)

Source/VCGeneration/Checker.cs
```diff
@@ -248,6 +248,10 @@ public bool HasOutput
       get { return hasOutput; }
     }
 
+    /// <summary>
+    /// Gets the amount of time How that the prover spent working. This measures the cost of processing queries,
```
keyboardDrummer, line 252: 'How' seems superfluous

Source/VCGeneration/Checker.cs
```diff
@@ -248,6 +248,10 @@ public bool HasOutput
...
+    /// but it does not include startup costs (unlike <c>ProverStartTime</c>).
```
keyboardDrummer, line 253: What startup costs, starting the solver is already done before ProverStateTime, right?

shazqadeer:
@cpitclaudel : It would be good to land this PR soon. I see that we already have some conflicts presumably from the recent PR landed by @atomb. Thanks.

shazqadeer:
@cpitclaudel : Should we abandon this PR?

shazqadeer:
Thank you. Much appreciated.

Source/VCGeneration/Checker.cs
cpitclaudel: It was poor phrasing / a typo.  There's no `ProverStartTime`; it's `VerificationResult.Start` that I meant.

Source/Provers/SMTLib/SMTLibInteractiveTheoremProver.cs
```diff
@@ -85,6 +85,7 @@ public override async Task<Outcome> Check(string descriptiveName, VCExpr vc, Err
           processNeedsRestart = true;
         }
 
+        watch.Start();
```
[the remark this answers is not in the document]
cpitclaudel: They are not (intended to be) different; it's just that using `end time - start time` to measure elapsed time is not great (because these clocks are not guaranteed to be monotonic and are not very precise anyway).  

However, there is no good reason to have them set in separate places. I cleaned things up by rewriting the code to start the stopwatch when `ProverStartTime` is set and consult it when the elapsed time was computed.

cpitclaudel:
Updated and cleaned up.  I've simplified the timing code based on @keyboardDrummer feedback, and documented the remaining shortcomings.

keyboardDrummer (approved):

PR #647, opened 2022-11-06
[Civl] UnpackCmd

shazqadeer:
This PR adds syntax for unpacking a constructed value into its components. In addition to being generally useful, this syntax is essential for supporting linear maps/sets/vals nested within datatypes.

To support nested linear values, this PR also generalizes linear type checking and permission instrumentation:
- A variable that has a linear type nested within it is also considered linear and subject to the usual typing rules
- Permissions are computed only for variables that has a linear map/set/val at the root, thereby creating for the first time the possibility that a linear variable may not contribute a permission. Such a variable has permissions nested inside it but these are not revealed to the logical reasoning.

To pull all of the above off, a substantial amount of refactoring of linear type checker was done. This refactoring had the following side effects:
- SignatureMatcher was tightened to disallow different names at the same position.
- GlobalSnapshotInstrumentation and NoninterferenceInstrumentation classes were eliminated and their logic merged more compactly into YieldingProcInstrumentation.


Source/Concurrency/CivlCoreTypes.cs
```diff
@@ -79,12 +79,14 @@ public abstract class Action
     public HashSet<Variable> gateUsedGlobalVars;
     public HashSet<Variable> actionUsedGlobalVars;
     public HashSet<Variable> modifiedGlobalVars;
+    public ConcurrencyOptions options;
```
bkragl, line 82: I think I'd prefer if we don't keep references to options in classes like `Action`, but instead pass options in wherever necessary. Would that be less convenient?

shazqadeer: Ok, I will make an attempt.

PR #648, opened 2022-11-12
[Civl] Replace the use of commutativity witnesses with pools

shazqadeer:


bkragl:
Should we think about removing the "commutativity hints" feature from the code? There are two variants. The "witness" variant to supply a witness for a global variable in a commutativity check, which after this PR is not used in any example anymore. And the "lemma" variant to inject an arbitrary (unchecked) assumption into a commutativity check, which is only used in a contrived example that illustrates the usage of the feature.

Why was `add_to_pool` restricted to predicate commands? (just curious)

shazqadeer:
> Why was add_to_pool restricted to predicate commands?

It is particularly simple to understand the semantics of add_to_pool when it is attached to a predicate command which does not have side effects. The plumbing required to handle add_to_pool correctly for other commands on which attributes can be applied is more complicated. In the interest of more robustness, I decided to be less general.

shazqadeer:
> Should we think about removing the "commutativity hints" feature from the code? There are two variants. The "witness" variant to supply a witness for a global variable in a commutativity check, which after this PR is not used in any example anymore. And the "lemma" variant to inject an arbitrary (unchecked) assumption into a commutativity check, which is only used in a contrived example that illustrates the usage of the feature.

Ok. I will put up a PR.

PR #652, opened 2022-11-16
[Civl] updates to pool instantiation and variable elimination

shazqadeer:
- Pool instantiation was not collecting all quantifiers that could be instantiated. While skolemization of a quantifier requires knowing the polarity, instantiation can be done on all quantifiers.
- Variable elimination in transition relation computation tries different sets of defined variables one after another. Changed the last attempt to be based on all variables that have a pool associated with them.

### DIFF
--- a/Source/Concurrency/TransitionRelationComputation.cs
+++ b/Source/Concurrency/TransitionRelationComputation.cs
@@ -370,7 +370,7 @@ namespace Microsoft.Boogie
       {
         TryElimination(Enumerable.Empty<Variable>());
         TryElimination(trc.allLocVars.Select(v => varCopies[v][0]));
-        TryElimination(trc.allLocVars.Where(v => varCopies[v].Count > 1).Select(v => varCopies[v][1]));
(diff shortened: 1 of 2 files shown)


Source/Concurrency/TransitionRelationComputation.cs
```diff
@@ -370,7 +370,7 @@ private void EliminateIntermediateVariables()
       {
         TryElimination(Enumerable.Empty<Variable>());
         TryElimination(trc.allLocVars.Select(v => varCopies[v][0]));
-        TryElimination(trc.allLocVars.Where(v => varCopies[v].Count > 1).Select(v => varCopies[v][1]));
```
bkragl, line 373: Why not have both?

shazqadeer: We could have both. It would be great to have a comment right there indicating all the possibilities that exist. 

I chose to minimize the number of different attempts because that seemed like the principled thing to do. I even tried to remove the attempt on line 372 but some of the tests started failing.

PR #656, opened 2022-11-25
Release version 2.15.10

MikaelMayer:


robin-aws (changes requested):
Based on a quick glance at the commits since 2.15.9, I'm pretty darn sure there's at least one new feature in there, so this should probably be 2.16.0 instead.

Source/Directory.Build.props
```diff
@@ -2,7 +2,7 @@
 
   <!-- Target framework and package configuration -->
   <PropertyGroup>
-    <Version>2.15.9</Version>
+    <Version>2.15.10</Version>
```
robin-aws, line 5:
```suggestion
    <Version>2.16.0</Version>
```

PR #661, opened 2022-11-28
Add  field update expression

shazqadeer:
This PR adds a field update expression to Boogie. The syntax is inspired by Dafny. The expression 
```
e1->(f := e2)
```
updates field f of the datatype value computed by e1 with the result from e2. The type of the expression is the same as the type of e1.


shazqadeer:
@RustanLeino : I am going to merge this PR. But please leave your feedback here. I will make sure to address it in subsequent PRs separately. I consider datatype support for Boogie ongoing work and there will be several PRs before it gets done.

shazqadeer:
@bkragl : FYI

Issue #659, opened 2022-11-26
Boogie generates cvc5 incompatible smt2 syntax without monomorphize


ChuyueSun:
After I modified [this part](https://github.com/ChuyueSun/dafny/blob/55212773b2fc7d86b3f5eda77f2b31d750f6b71a/Source/DafnyCore/DafnyPrelude.bpl#L961-L1184) of the dafny prelude to invoke the built-in sequence theory in cvc5. As a result, [this dfy file](https://github.com/ChuyueSun/dafny/blob/master/Test/mytest/seq.dfy) will generate [this bpl file](https://github.com/ChuyueSun/dafny/blob/55212773b2fc7d86b3f5eda77f2b31d750f6b71a/Test/mytest/seq_generated_reference.bpl#L1165-L1194). Unfortunately, boogie cannot directly monomorphize this program due to limitations created by the other parts of the prelude. 

If I run `boogie/Source/BoogieDriver/bin/Debug/net6.0/BoogieDriver  /proverOpt:PROVER_PATH="cvc5_path" /proverOpt:C:"--strings-exp" /proverLog:Test/mytest/seq_generated_smt_by_reference.smt2 Test/mytest/seq_generated_reference.bpl` on the bpl file I mentioned above, then it will give 

```
Test/mytest/seq_generated_reference.bpl(233,9): Warning: type parameter T is ambiguous, instantiating to int
Test/mytest/seq_generated_reference.bpl(233,35): Warning: type parameter T is ambiguous, instantiating to int
Advisory: m (correctness) SKIPPED due to I/O exception: Pipe is broken.
Test/mytest/seq_generated_reference.bpl(2592,49): Verification encountered solver exception (m (correctness))

Boogie program verifier finished with 0 verified, 0 errors, 1 solver exceptions
```

I should also note that the current boogie on master does not recognize my solver to be CVC5 when I use the `PROVER_PATH` option. I made manual changes on my local branch to fix it.

The generated `Test/mytest/seq_generated_smt_by_reference.smt2` smt2 file is [here](https://github.com/ChuyueSun/dafny/blob/master/Test/mytest/seq_generated_smt_by_reference.smt2).

If I run the following command on [this bpl file](https://github.com/ChuyueSun/dafny/blob/master/Test/mytest/seq_generated.bpl) which only contains sequences related parts, then it is able to verify only with `\monomorphize`.
```
[...]

PR #664, opened 2022-12-09
Emit axioms only for types used in program

### LINKED ISSUE #660: Emit axioms only for theories used by queries
Currently, the SMT-Lib generated from Boogie includes background axioms for the built-in data types, even when they aren't used. In particular, this includes the `RegEx` type, which CVC5 doesn't support. By emitting axioms only for used theories, we could improve the ability to send a single SMT query to multiple solvers.

This would solve part of the issue encountered in the discussion of #659.

atomb:
Previously, Boogie would emit axioms for all types it supports, even if not used in the input program. This PR causes it to instead emit axioms only for the types that are used. This results in more compatible SMT-Lib files, usable by solvers that don't support theories such as regular expressions.

Fixes #660. Also addresses part of #659.


keyboardDrummer (approved):

MikaelMayer (approved):
Thanks. Although I'm not sure it will improve efficiency (which you don't claim anyway), having it to work with other SMT solvers is priceless.

shazqadeer:
@atomb:

When I try to run the test UsedTypes.bpl on my machine (Mac), I get this error message:

```
$ "!" "grep" "RegEx" "/Users/shaz/Source/boogie/Test/prover/Output/UsedTypes.bpl.tmp.smt2"
# command stderr:
'!': command not found
error: command failed with exit status: 127
```

Do I have to install some other tool ?

atomb:
Oh, interesting. It works correctly on my Mac. I'll look into whether there's a better way to require the absence of a match in a file.

keyboardDrummer:
> Oh, interesting. It works correctly on my Mac. I'll look into whether there's a better way to require the absence of a match in a file.

Dafny has this line to introduce `!`: https://github.com/dafny-lang/dafny/blob/master/Test/lit.site.cfg#L178

But that's not in https://github.com/boogie-org/boogie/blob/master/Test/lit.site.cfg

atomb:
> > Oh, interesting. It works correctly on my Mac. I'll look into whether there's a better way to require the absence of a match in a file.
> 
> Dafny has this line to introduce `!`: https://github.com/dafny-lang/dafny/blob/master/Test/lit.site.cfg#L178
> 
> But that's not in https://github.com/boogie-org/boogie/blob/master/Test/lit.site.cfg

Strange, given that I wasn't using Dafny's configuration to run this test. Regardless, the test uses a more standard approach now.

PR #667, opened 2022-12-17
Remove <:, extends, complete

shazqadeer:
Boogie has a polymorphic partial order operator ```<:```.  To support specifications related to this operator, there are two other keywords---extends and complete. As far as I know, these features are not being used in any client of Boogie. Moreover, Boogie already supports polymorphic functions (and relations) and axioms, so it is already capable of expressing the constraints related to ```<:```. The handling of partial orders inside Boogie is nothing special---a bunch of axioms are generated and introduced into the VC. These axioms could be generated by clients themselves and included in the generated Boogie program. 

Overall, the partial order feature does not seem to offer much benefit to use cases of Boogie as an IVL. In the interest of simplifying Boogie code base and making it more robust, I decided to drop this feature.


keyboardDrummer (approved):
Looks good, thanks

crhf:
it's used in https://github.com/martinschaef/jar2bpl, the only java front end I can find...

Issue #560, opened 2022-03-29
Prune is nondeterministic with -vcsCores:2


RustanLeino:
Running Boogie repeatedly on the same input and with `-vcsCores:2 -prune` produces different output (for example, sometimes reporting 2 verification errors and other times reporting 4 verification errors). Furthermore, using `-printPruned` shows that there are differences even in the cases when the same number of verification errors is reported. For the test file I'm enclosing, it is also the case that running without pruning gives 0 verification errors (instead of 2 or 4).

## Repro

I'm calling `boogie` with the SMT settings that Dafny uses, namely

    /proverOpt:O:smt.arith.solver=2 /proverOpt:O:auto_config=false /proverOpt:O:type_check=true /proverOpt:O:smt.case_split=3 /proverOpt:O:smt.qi.eager_threshold=100 /proverOpt:O:smt.delay_units=true

and I'm using Z3 version 4.8.5 on MacOS. In the repros below, I'm also using `-errorTrace:0` for cleaner output, but this flag makes no difference for the issue I'm reporting. The input file I used is [d.bpl](https://github.com/boogie-org/boogie/files/8375152/d.bpl.txt).

### Pruning and using 2 cores

Runnng `boogie d.bpl $(SMT_OPTIONS_ABOVE) -errorTrace:0 -vcsCores:2 -prune` sometimes gives

```
d.bpl(5595,11): Error: This loop invariant might not be maintained by the loop.
d.bpl(5601,11): Error: This loop invariant might not be maintained by the loop.

Boogie program verifier finished with 12 verified, 2 errors
```

and sometimes gives

```
d.bpl(5595,11): Error: This loop invariant might not be maintained by the loop.
d.bpl(5601,11): Error: This loop invariant might not be maintained by the loop.
d.bpl(6715,1): Error: A postcondition might not hold on this return path.
d.bpl(6556,3): Related location: This is the postcondition that might not hold.
d.bpl(6715,1): Error: A postcondition might not hold on this return path.
d.bpl(6571,3): Related location: This is the postcondition that might not hold.

Boogie program verifier finished with 11 verified, 4 errors
```

If I additionally use `-printPruned` to print the pruned files, then I get differences even when the number of errors is generated to be the same. This suggests that the nondeterministic pruning happens more often than the verification output shows. For example, doing three consecutive runs of `boogie` with the flags above and `-printPruned` gives the following differences:

```
% diff tmp0.out tmp1.out
26282,26284d26281
[...]

PR #670, opened 2022-12-21
Fix graph search used in declaration pruning

shazqadeer:
This PR fixes an error in the graph search used in declaration pruning by properly accounting for the meaning of merge nodes in the fixpoint calculation.

I expect this PR to fix the nondeterminism issue reported in #560 . However, the cause of incompleteness when using /prune remains a mystery.


keyboardDrummer (commented):
Could you add a test on `FindReachableNodesInGraphWithMergeNodes` that fails without this change?

shazqadeer:
The original code attempted to traverse the outgoing edges of a merge node as soon as it was discovered, i.e., before all reachable nodes are known. This would not be correct based on my understanding of what this code is trying to accomplish. It is possible that a merge node is discovered and processed and its outgoing edges are ignored because some member of the merge node is not known to be in the current visited set. But if the processing had been delayed, we could have discovered that all members of the merge node have been visited and therefore the outgoing edges need to be processed. 

If my understanding of visited set is correct, the older code could result in reporting a subset of the real visited set.

shazqadeer:
> Could you add a test on FindReachableNodesInGraphWithMergeNodes that fails without this change?

I have only written end-to-end tests before that are executed from the command line using lit. You are suggesting something else here?

shazqadeer:
I suspect there is another problem with the code but I have not attempted to fix it yet. The code in Prune.ComputeDeclarationDependencies calculates the dependency graph whose nodes are either a Declaration or a Declaration[]. It is possible that there are two instances of Declaration[] which contain the same Declaration's. Based on my understanding of what this code is trying to accomplish, those two instances must be unified in the dependency graph. In other words, we want the nodes in the dependency graph to be either Declaration or HashSet<Declaration> and we want the key comparison in the dictionary representing the dependency graph to be done on set semantics rather than memory address semantics.

shazqadeer:
I have also attempted a fix to the second bug I described above.

keyboardDrummer:
> The original code attempted to traverse the outgoing edges of a merge node as soon as it was discovered, i.e., before all reachable nodes are known. This would not be correct based on my understanding of what this code is trying to accomplish. It is possible that a merge node is discovered and processed and its outgoing edges are ignored because some member of the merge node is not known to be in the current visited set. But if the processing had been delayed, we could have discovered that all members of the merge node have been visited and therefore the outgoing edges need to be processed. If my understanding of the meaning of merge nodes is correct, the older code could result in reporting a subset of the real visited set.

But if the merge node was visited and not all incoming edges had been visited, then the merge node was not added to the visited set, so it would be visited again when another incoming edge would be visited, ensuring that if ever all the incoming edges of the merge node would be visited, then the outgoing edges would be as well.

```c#
        var node = todo.Pop();
        if (visitedEdges.Contains(node)) {
          continue;
        }

        if (node is IEnumerable<object> objects) {
          if (!visitedEdges.IsSupersetOf(objects)) {
            continue;
          }
        }
        // note that the add is skipped if the above continue is hit.
        visitedEdges.Add(node);
```

> I suspect there is another problem with the code but I have not attempted to fix it yet. The code in Prune.ComputeDeclarationDependencies calculates the dependency graph whose nodes are either a Declaration or a Declaration[]. It is possible that there are two instances of Declaration[] which contain the same Declaration's. Based on my understanding of what this code is trying to accomplish, those two instances must be unified in the dependency graph. In other words, we want the nodes in the dependency graph to be either Declaration or HashSet and we want the key comparison in the dictionary representing the dependency graph to be done on set semantics rather than memory address semantics.

The default equality/hashcode implementation of `HashSet` looks at the elements in the set, not the pointer of the `HashSet`:
https://dotnetfiddle.net/hl2GbS

shazqadeer:
>The default equality/hashcode implementation of HashSet looks at the elements in the set, not the pointer of the HashSet: https://dotnetfiddle.net/hl2GbS

> Does that alleviate your concern?

No. The problem I had in mind was something like this: https://dotnetfiddle.net/FxobuY. I would have liked to see the number of members in collectionOfHashSets to be 2 rather than 3.

shazqadeer:
After your explanation of the original code, I agree that the first problem I pointed out will not happen.

shazqadeer:
Upon reading the code in AxiomVisitor.VisitExpr, I detected some other potential problems:

1. It looks like the field pos to track polarity of an expression was added to Expr class just for this use case. This is inappropriate because pos is a temporary piece of tracking state. It should not be added to a core class such as Expr. Rather, we should attempt to track this state in AxiomVisitor itself.

2. The propagation of pos looks erroneous to me. For example, I see this code
```
n.Args[0].pos = Expr.NegatePosition(n.Args[0].pos);
```
to update pos of the child when the parent n is a Not expression. I suspect the intention was to do
```
n.Args[0].pos = Expr.NegatePosition(n.pos);
```
Otherwise, nested negation will not be handled correctly. There are similar problems in handling of other expressions also.

shazqadeer:
I stumbled upon the helpful comment in CommandLineOptions.cs that explains the design of the pruning feature. Here are a few thoughts on the design.

The uses declaration seems odd to me for two reasons. First, it may happen that an axiom constrains several functions in which case I would like to add an edge from each of those functions to the axiom. With the current design, we would have to copy the axiom in the uses clause of each of the functions. Second, just as a matter of taste (and Boogie tradition), it is better to use attributes to introduce experimental features into Boogie. Later, when the feature is well understood, we can introduce custom ergonomic syntax for it. The uses clause could have been simulated by adding an attribute on axioms comprising names of functions/constants from which edges to the axiom must be added.

The design of the :include_dep attribute given the rationale of easy migration did not make sense to me. Migration concerns would suggest that we preserve existing behavior by default, which would be to do as little pruning as possible. With the current semantics of :include_dep, an unannotated axiom is likely to be dropped because of lack of incoming edges resulting in excessive pruning which would then have to be turned off by suitably adding :include_dep. The alternative of adding references from declarations mentioned in the axiom by default and changing the behavior based on suitable attribute on the axiom would be an alternative design.

keyboardDrummer:
> > The default equality/hashcode implementation of HashSet looks at the elements in the set, not the pointer of the HashSet: https://dotnetfiddle.net/hl2GbS
> 
> > Does that alleviate your concern?
> 
> No. The problem I had in mind was something like this: https://dotnetfiddle.net/FxobuY. I would have liked to see the number of members in collectionOfHashSets to be 2 rather than 3.

Err.. You're right, thanks.

> Because of the invariant above and your explanation of the code, the search does not depend on logical equality over nodes of type Set; reference equality suffices.

I think it would be OK to use logical set quality for the `Declaration[]` objects when putting them in `edges` in `ComputeDeclarationDependencies`, but indeed it's not required.

> The algorithm and implementation seems to be highly customized to the specifics of pruning code. 

`GraphAlgorithms.FindReachableNodesInGraphWithMergeNodes` still seems generic to me, that's why it's in that location.

> It would be nice to document all these specific details.

I think mentioning in `ComputeDeclarationDependencies ` that it's OK if there are multiple `Declaration[]`s with the same set representation in the `edges` graph, would be good.

shazqadeer:
I will not land this PR. The discussion is useful for anybody who wants to refactor the pruning code. In addition to the discussion above, I also want to point out that the untyped API of the graph search method is not desirable. A cleaner formalization ought to be possible.

PR #676, opened 2023-01-06
Fix: Error wording closer to semantics

MikaelMayer:
This PR is the companion of https://github.com/dafny-lang/dafny/pull/3324
We established numerous times this year that the wording "might not hold" still incorrectly provides the feeling that the postcondition is incorrect, when the real problem is often that some proof hints are missings, which is rather what we want our users to focus on.
So, rather than focusing on whether assertions hold or not, this PR fixes the message by focusing on the fact that the verifier was not able to prove the assertion. Proof now becomes a first-class goal.

There was the choice of starting all errors by "Could not prove", but then the first 3 words of every error message are never informative, which is not very friendly. The passive voice solves this problem and is less of a disruption in the errors messages.


Source/Core/ProofObligationDescription.cs
```diff
@@ -32,75 +32,75 @@ public abstract class ProofObligationDescription {
 
 public class AssertionDescription : ProofObligationDescription
 {
-  public override string SuccessDescription => "This assertion holds.";
+  public override string SuccessDescription => "this assertion holds";
```
rakamaric, line 35: If we are going after "proven" to stress out this is about proofs, then why not also "proven" instead of "holds"?
Like "Boogie proved this assertion".

rakamaric: Same goes for other such patterns.

Source/UnitTests/ExecutionEngineTests/ExecutionEngineTest.cs
```diff
@@ -70,10 +70,10 @@ procedure Bad2(y: int)
     options.VcsCores = 4;
     var engine = ExecutionEngine.CreateWithoutSharedCache(options);
 
-    var expected = @"fakeFilename1(3,3): Error: This assertion might not hold.
+    var expected = @"fakeFilename1(3,3): Error: this assertion could not be proven
 Execution trace:
     fakeFilename1(3,3): anon0
-fakeFilename1(8,3): Error: This assertion might not hold.
+fakeFilename1(8,3): Error: this assertion could not be proven
```
rakamaric, line 76: I am also wondering if active voice would be nicer in all of these. So something like: "Boogie could not prove this assertion". Personally, I would prefer such active voice to current passive voice.

atomb:
This looks good overall, but it looks like the tests still need to be updated to account for the output changes.

rakamaric:
As a side note, this will be a breaking change for (some) downstream tools. For example, SMACK relies on these messages to figure out the verification status. Not that we should not make this update, but just wanted to mention this.

Source/UnitTests/ExecutionEngineTests/ExecutionEngineTest.cs
MikaelMayer: Me too, but after looking at the error messages, the first 4 words of "Boogie could not prove " are useless to know what the error is about, whereas by keeping this assertion as the subject, it's much clearer what's in the focus there.

Source/Core/ProofObligationDescription.cs
MikaelMayer: That's a good point. Note that the success message is for now only displayed in rare occasions, in some toolings and while hovering Dafny programs in VSCode. Since there is no call to action, "holds" is the same as "was proven", but it's shorter.

MikaelMayer:
> As a side note, this will be a breaking change for (some) downstream tools. For example, SMACK relies on these messages to figure out the verification status. Not that we should not make this update, but just wanted to mention this.

Indeed, this is also a breaking change for Dafny. The only reason I'm pushing for this is that last year we went from "assertion violation" to "assertion might not hold", which was already a plus (and we did not get push back), yet, new users are still puzzled and first try to figure out why their assertion is wrong when all they need is figure out how to actually prove it.
So I think this change is going to be extremely useful to avoid this problem.

shazqadeer:
@MikaelMayer : I don't quite see how changing the words from "does not hold" to "could not prove" or other such variations will actually help a Boogie or Dafny user decide on the next step. I do agree that "could not prove" is more accurate so I do not object to making this change.

bkragl:
Bikeshedding here, but I think that "proved" is better than "proven" (although both are considered correct). We could consult a particular style guide, but let's use metrics instead: one less syllable to pronounce and more Google hits :)

MikaelMayer:
> In the last commit, I'm fixing the last two CI errors. One is puzzling to me. [9eb5120](https://github.com/boogie-org/boogie/commit/9eb512099516fa62d1acb801086751429a2702c1) An error message is displayed 10 lines earlier on another assertion. I don't understand how this is possible because I only changed error messages. I'm investigating though.

Ok it looks like the test is a parallel test, so the error isn't one, it's more a stability one. It looks like  it's deterministic though so I think changing the test output makes sense.

Source/Core/ProofObligationDescription.cs
```diff
@@ -32,75 +32,75 @@ public abstract class ProofObligationDescription {
...
 
 public class InvariantMaintainedDescription : AssertionDescription
 {
   public override string SuccessDescription =>
-    "This loop invariant is maintained by the loop.";
+    "this loop invariant is maintained by the loop";
 
   public override string FailureDescription =>
-    "This loop invariant might not be maintained by the loop.";
+    "this loop invariant might not be maintained by the loop";
```
atomb, line 103: This message still contains the "might not" phrasing, as opposed to "could not be proved". Unfortunately, using that wording makes it pretty awkward: "this loop invariant could not be proved to be maintained by the loop".

atomb:
It looks like there are two test files remaining to update.

Source/Core/ProofObligationDescription.cs
MikaelMayer: What about: "this invariant could not be proved within the loop"?
Or, a shorter version of yours, "this invariant could not be proved to be maintained by the loop"

MikaelMayer: In the current PR, I used the second one.

atomb: Yeah, I like that one, too.

atomb (approved):
Looks good!

keyboardDrummer:
Looks good to me too but let's get a few more approvals on this

shazqadeer (approved):

MikaelMayer:
> @MikaelMayer : I don't quite see how changing the words from "does not hold" to "could not prove" or other such variations will actually help a Boogie or Dafny user decide on the next step. I do agree that "could not prove" is more accurate so I do not object to making this change.

I'm not sure a lot about Boogie users, but I can tell you more about the Dafny user experience. As you outlined "could not prove" is more accurate, and that's already the first step.
What happened with new Dafny users is that, as soon as they encountered "might not hold", they immediately understood "is probably wrong". This has several possible causes:
- Users without verification background usually don't know what "false negative" is.
- In most compiler tools used by software engineers, if an error is reported, it is accepted that the user needs to fix something, that there is something intrinsically wrong with their code, they read the doc and the error, and they fix it. Their code was broken, now it can be compiled.
- Even type annotations are perceived as a necessity in most languages that do not have good type inference, and error messages are clear "X is used as Y but its inferred type was Z.", so users know that they need to add type annotation and even fix their code, no big deal.

Proofs are very much like type annotations, except that the type-checking system is incomplete.
So imagine, if users were presented an error on a type annotation (expression: X) like this

> "The provided type X of expression might be wrong" 

The first idea users would have is that, since the compiler says so, `: X` is the wrong annotation and the type of the expression is different. But it wouldn't come to mind that "might" indicates that it's actually correct and it requires hints, because adding hints is not something users do for type checking besides obvious places.

So when we provide an error message like "might not hold", we face the same problem. As users aren't used to provide hints, they might erroneously adhere to the belief that the compiler is clever and just being nice and that something is indeed wrong.

With "could not prove", not only it is more accurate, but the subject of the "prove" is Dafny itself. It says nothing about the proposition being false, only that it could not find a proof. This is the implicit invitation we want Dafny users to receive.

Hope it helps ! Thanks for the review.

shazqadeer:
@MikaelMayer : I appreciate the detailed explanation. Thanks.

PR #680, opened 2023-01-12
add more tests for monomorphization

gauravpartha:
This pull requests adds some more tests for the monomorphization of polymorphic maps and universal type quantification. One of the tests is regarding the liberal typing rule for equality (`a == b` type checks in Boogie even in cases where `a` and `b` may not have the same identical type as long as there is some instantiation of the free type variables that result in the same type). One test is on polymorphic lambdas. One test is on polymorphic maps.


shazqadeer (approved):
Thanks for the examples.

PR #682, opened 2023-01-18
Add a Dafny-derived test

atomb:
The goal here is to more regularly exercise features used by Dafny during Boogie development, to more quickly discover any changes that may have a negative effect on Dafny verification. This doesn't replace running the full Dafny test suite on prospective Boogie changes, but hopefully will allow certain issues to be detected with less effort.


shazqadeer:
In one of the test runs, one of the problems in Seq.bpl is failing. This suggests the test is flaky. 

@atomb : Please try the wonderful randomization feature you added to Boogie with your new test and provide here some estimate of the flakiness you are seeing.

atomb:
Alright, now it goes through. For anyone curious, dafny-lang/libraries#65 includes the changes needed to make the proofs less flaky.

shazqadeer:
Great! For completeness, please also report on how many randomizeVcIterations you ran on the revised test and approximately how much variability in running time you saw. This data will bolster confidence that the test will continue to not be flaky.

atomb:
Running 10 iterations through Dafny (so that I can get output to use with [`dafny-reportgenerator`](https://github.com/dafny-lang/dafny-reportgenerator/)), all of the proofs in this file complete with a normalized standard deviation in solver resource user of under 65% and in duration of under 55%. Very few of the proofs have normalized standard deviation of either metric of more than 30%. If this test gives us trouble in the future, it should be possible to improve those few more widely-variable proofs.

It's perhaps worth mentioning that I intended this test to be at least a little difficult (and perhaps flaky, with the current implementation) as a guide toward improvements to Boogie itself that may reduce that problem. If it ends up causing more trouble than it's worth at any point, I'm quite happy to disable or improve it.

MikaelMayer (approved):
Oh that's a great idea. Thanks for including all the details needed to regenerate the file as well.

shazqadeer:
@atomb : Thanks for the data about how you estimated the "flakiness". Good work! Let us land this PR.

shazqadeer (approved):

PR #689, opened 2023-02-07
Set prover options outside (push) in batch mode

atomb:
This makes files generated for Z3 slightly more compatible with CVC5. Although CVC5 is supported in a separate mode, it can be useful for SMT files generated when running with Z3 to be as compatible as possible with other solvers.

### DIFF
--- a/Source/Provers/SMTLib/SMTLibBatchTheoremProver.cs
+++ b/Source/Provers/SMTLib/SMTLibBatchTheoremProver.cs
@@ -78,6 +78,7 @@ namespace Microsoft.Boogie.SMTLib
           await currentLogFile.WriteAsync(common.ToString());
         }
 
+        SendVCOptions();
         PrepareCommon();
         FlushAxioms();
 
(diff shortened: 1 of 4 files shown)


Source/Provers/SMTLib/SMTLibBatchTheoremProver.cs
```diff
@@ -87,7 +88,10 @@ public override int FlushAxiomsToTheoremProver()
         FlushAxioms();
 
         Push();
-        SendVCAndOptions(descriptiveName, vcString);
+        if (this.libOptions.EmitDebugInformation) {
```
keyboardDrummer, line 91: Can we deduplicate this `if (this.libOptions.EmitDebugInformation) {`

atomb: Probably so. I'll look into it.

keyboardDrummer (approved):

PR #692, opened 2023-02-14
Do not let assertion batch completion notification be fired before a running notification

keyboardDrummer:
### Changes 

Do not let assertion batch completion notification be fired before a running notification

### Testing

- Updated tests to use the new API


Source/VCGeneration/SplitAndVerifyWorker.cs
```diff
@@ -12,6 +13,9 @@ namespace VC
 {
   class SplitAndVerifyWorker
   {
+    public IObservable<(Split split, VCResult vcResult)> BatchCompletions => batchCompletions;
```
atomb, line 16: This is a very nice improvement. Thanks!

atomb (changes requested):
This seems like a very nice improvement to me. It looks like `test0/Irreducible.bpl` is timing out now, though. Locally, on `master` that test passes very quickly for me, so I wonder whether the changes introduced a deadlock or something like that.

atomb (approved):
Looks good!

PR #693, opened 2023-02-15
Do not wait for the solver process to exit before continuing

keyboardDrummer:
This improves speed when solver runs are being cancelled. The test `ManualRunCancelCancelRunRun` in https://github.com/dafny-lang/dafny/blob/master/Source/DafnyLanguageServer.Test/Synchronization/VerificationStatusTest.cs#L244 goes from >7s seconds to ~4 second of run time.


atomb (approved):
Looks reasonable to me.

PR #707, opened 2023-03-17
Remove many static console write usages

keyboardDrummer:
This makes the code more reusable. One use-case is running the code concurrently while keeping the output streams separate.


Source/ExecutionEngine/CommandLineOptionEngine.cs
```diff
@@ -0,0 +1,266 @@
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.Contracts;
+using System.IO;
+
+namespace Microsoft.Boogie;
+
+public class CommandLineOptionEngine
```
keyboardDrummer, line 8: Moved out of `CommandLineOptions.cs`

Source/Core/AST/Program.cs
```diff
@@ -0,0 +1,587 @@
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.Contracts;
+using System.Linq;
+using Microsoft.Boogie.GraphUtil;
+
+namespace Microsoft.Boogie;
+
+public class Program : Absy
```
keyboardDrummer, line 9: Moved out of `Absy.cs`. Updated one line to `rc.Options.OutputWriter.WriteLine("Warning: Ignoring implementation {0} because of translation resolution errors",`

Source/Core/TokenTextWriter.cs
```diff
@@ -493,7 +496,10 @@ public void WriteLine(Absy node, int level, string text, params object[] args)
 
     public void Close()
     {
-      this.writer.Close();
+      if (disposeWriter) {
+        this.writer.Close();
+
```
atomb, line 501: nit: I'd leave out this blank line.

Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -53,24 +53,17 @@ public class ExecutionEngine : IDisposable
...
-
     public ExecutionEngine(ExecutionEngineOptions options, VerificationResultCache cache) {
       Options = options;
       Cache = cache;
       checkerPool = new CheckerPool(options);
       verifyImplementationSemaphore = new SemaphoreSlim(Options.VcsCores);
       
       var largeThreadScheduler = CustomStackSizePoolTaskScheduler.Create(16 * 1024 * 1024, Options.VcsCores);
-      largeThreadTaskFactory = new(CancellationToken.None, TaskCreationOptions.DenyChildAttach, TaskContinuationOptions.None, largeThreadScheduler);
+      largeThreadTaskFactory = new(CancellationToken.None, TaskCreationOptions.None, TaskContinuationOptions.None, largeThreadScheduler);
```
atomb, line 66: Do you know why it was `DenyChildAttach` before?

atomb (approved):
Looks good! I left a couple of minor comments, but they don't necessarily need to lead to changes.

Source/ExecutionEngine/ExecutionEngine.cs
keyboardDrummer: I don't think this change has any effect. By default tasks are not attached. A child must specify `TaskCreationOptions.AttachedToParent` before it attaches, which isn't done in this codebase.

Source/Core/TokenTextWriter.cs
keyboardDrummer: Time for a `dotnet format` CI check ^,^

PR #709, opened 2023-03-29
Dynamic stack

### LINKED ISSUE #711: Replace call-stack based AST traversal code with heap-stack based traversal
We current use threads with very large stack sizes to prevent getting a stack overflow when traversing deep ASTs. However, it would be better to change these traversals so they don't use the call-stack for going into nested code.

PR #709 is a WIP attempt at doing so

keyboardDrummer:
Fixes https://github.com/boogie-org/boogie/issues/711


shazqadeer:
@keyboardDrummer : Are you interested in pursuing this PR further? If not, I will close it.

PR #712, opened 2023-04-05
[Corral] Make ControlFlowIdMap into a field

Dargones:
This PR is part of the effort to update [Corral](https://github.com/Dargones/corral) so that it works with the latest version of Boogie. The idea is to incrementally update this branch of Boogie alongside Corral until a PR could be made to Boogie's main branch.

This particular PR makes to changes:

- Make Boogie 2.9.5 compilable with dotnet6.
- Change a local variable that stores the mapping from AST nodes to Ids into a field of StatifiedInliningInfo, so that Corral can reuse that mapping later on during error-trace parsing.

### DIFF
--- a/Source/Directory.Build.props
+++ b/Source/Directory.Build.props
@@ -3,7 +3,7 @@
   <!-- Target framework and package configuration -->
   <PropertyGroup>
     <Version>2.9.5</Version>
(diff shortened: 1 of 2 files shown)


rakamaric (approved):
LGTM

PR #720, opened 2023-05-01
Remove usage of ! for older lit versions

keyboardDrummer:


davidcok (commented):
I'm curious why this is needed. The change is not equivalent: the ! checks that there is a non-zero exit code, whereas the change ignores the exit code.

shazqadeer:
> I'm curious why this is needed. The change is not equivalent: the ! checks that there is a non-zero exit code, whereas the change ignores the exit code.

@davidcok : Could you point me to the documentation for the ! operator?


I see the following on my end:

```
shaz@shaz-mbp commandline % lit --version
lit 0.11.1
```

What do you have on your end?

shazqadeer:
I updated lit on my machine:

```
shaz@shaz-mbp commandline % lit --version
lit 16.0.2
```

Now the test works on my end. It seems fine to continue to use the ! operator. A note about the lit version number on the main page where we mention the lit dependency might be helpful.

davidcok:
I also have 0.11.1 - and ! works fine.
I don't see mention of ! in the online llvm-lit documentation, but that documentation is very brief.
But I should not speak for lit in the boogie environment.
In dafny testing we have implemented a number of extensions to lit, such as additional % macros. Perhaps ! is an extension as well. In any case, it and %exits-with are very handy for testing expected-to-fail RUN commands.

PR #731, opened 2023-05-17
Clean up threads from custom thread pool when disposing ExecutionEngine

keyboardDrummer:
Clean up threads from custom thread pool when disposing ExecutionEngine to enable freeing up resources


atomb (approved):
Even though this may occasionally fail to clean up all threads, at least it frequently does, which is still an improvement.

PR #734, opened 2023-05-24
Revert "Clean up threads from custom thread pool when disposing ExecutionEngine (#731)"

atomb:
This reverts commit fcf0dde0f9051e5271ed937908531119825a4b64.

The change was leading to stack overflows in Dafny. I'd like to be able to track down what's causing the stack overflows, even if it's best fixed on the Dafny side, before committing to this change.


keyboardDrummer (changes requested):
Let's see if the issue is resolved with Boogie 2.16.8

PR #739, opened 2023-05-31
Make task scheduler code easier to understand

keyboardDrummer:


atomb (approved):
This seems reasonable to me.

PR #748, opened 2023-06-07
Remove obsolete methods from LinqExtender and do further cleanup

keyboardDrummer:
I got an error in Dafny because because `(a as IEnumerable<T>).Zip` resolves to both `Enumerable.Zip` and `LinqExtender.Zip`, two methods that behave exactly the same, and this triggered me to cleanup `LinqExtender`

### Changes

- Use the built-in method `List.ForEach` instead of `LinqExtenter.Iter` where possible, because it's faster and it's built-in.
- Remove LinqExtender's Concat, MapConcat, SkipEnd and Zip methods, since equivalents are also available in the standard library.
- Remove LinqExtender's Concat1 method since it was unused.
- Turn `LinqExtender.Map` into an extension method and rename it to `MapTo`
- Rename `LinqExtender` to `CollectionExtensions`
- Rename `Iter` to `ForEach` (this is in a separate commit, please review separately for efficiency)


shazqadeer (approved):
Looks good to me.

Issue #735, opened 2023-05-26
Inconsistency in predicate-based map axioms when using MBQI


atomb:
While investigating [this Dafny issue](https://github.com/dafny-lang/dafny/issues/4053), I discovered that Boogie seems to be generating map axioms when using the predicate-based encoding for polymorphic types that are inconsistent when MBQI (which is disabled by default) is enabled in the solver. Take the following Boogie program, distilled from the original Dafny example.

```
type ref;
type Field _;
type Heap = [ref]<alpha>[Field alpha]alpha;
function $IsGoodHeap(Heap) : bool;
function $IsHeapAnchor(Heap) : bool;
var $Heap: Heap where $IsGoodHeap($Heap) && $IsHeapAnchor($Heap);

procedure P()
{
    assert false;
}
```

When running with the various type encoding flags, and MBQI enabled, I get the following.

```
$ boogie issue-4053-pruned-prefix.bpl /proverOpt:O:smt.mbqi=true /typeEncoding:p

Boogie program verifier finished with 1 verified, 0 errors
$ boogie issue-4053-pruned-prefix.bpl /proverOpt:O:smt.mbqi=true /typeEncoding:a
issue-4053-pruned-prefix.bpl(10,5): Error: this assertion could not be proved
Execution trace:
    issue-4053-pruned-prefix.bpl(10,5): anon0

Boogie program verifier finished with 0 verified, 1 error
$ boogie issue-4053-pruned-prefix.bpl /proverOpt:O:smt.mbqi=true /typeEncoding:m
issue-4053-pruned-prefix.bpl(10,5): Error: this assertion could not be proved
Execution trace:
    issue-4053-pruned-prefix.bpl(10,5): anon0

Boogie program verifier finished with 0 verified, 1 error
$
```

Is there some reason it should be expected that these axioms are incompatible with MBQI? And is that why it's disabled by defalut?

Running various solvers on the generated SMT-Lib, several versions of Z3, CVC5, and Vampire all report `unsat`. This happens when enabling MBQI in Z3 or CVC5, and when using Vampire with its default flags. Using Z3 or CVC5 with MBQI disabled results in `unknown`.

atomb:
The comment from @RustanLeino [here](https://github.com/dafny-lang/dafny/issues/4053#issuecomment-1563731418) suggests how to fix the issue.

shazqadeer:
> Is there some reason it should be expected that these axioms are incompatible with MBQI? 
[...]

PR #749, opened 2023-06-09
Add missing antecedents to function axioms

atomb:
Add type-constraining antecedents for each argument of a function when generating the type axiom for that function.

These seem to have been intentionally but incorrectly left out in the original implementation.

This fixes the missing antecedent issue in #735 and dafny-lang/dafny#4053 but does not yet fix the missing type constructor parameters. I have yet to track down exactly why those are left out.

### DIFF
--- a/Test/test21/issue-735.bpl
+++ b/Test/test21/issue-735.bpl
@@ -1,4 +1,4 @@
-// RUN: %parallel-boogie /proverOpt:O:smt.mbqi=true /typeEncoding:p "%s" > "%t"
+// RUN: %parallel-boogie /proverOpt:O:smt.mbqi=true /timeLimit:5 /typeEncoding:p "%s" > "%t"
 // RUN: %diff "%s.expect" "%t"
 // RUN: %parallel-boogie /proverOpt:O:smt.mbqi=true /typeEncoding:a "%s" > "%t"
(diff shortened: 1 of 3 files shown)


Test/test21/issue-735.bpl
```diff
@@ -0,0 +1,18 @@
...
+// RUN: %diff "%s.expect" "%t"
+// RUN: %parallel-boogie /proverOpt:O:smt.mbqi=true /typeEncoding:m "%s" > "%t"
+// RUN: %diff "%s.expect" "%t"
+
+type ref;
+type Field _;
+type Heap = [ref]<alpha>[Field alpha]alpha;
+function $IsGoodHeap(Heap) : bool;
+function $IsHeapAnchor(Heap) : bool;
+var $Heap: Heap where $IsGoodHeap($Heap) && $IsHeapAnchor($Heap);
```
keyboardDrummer, line 13: Is this variable necessary for the reproduction? Also I'm guessing we only need one of the two functions.

keyboardDrummer (approved):

Test/test21/issue-735.bpl
atomb: Yeah, I suspect it could be simplified further. But it's relatively small and understandable now.

PR #753, opened 2023-06-20
Fix inlining of type constraints

Dargones:
This PR fixes a bug in which the `where` type constraint on a variable's declaration is not run through `BoundVarAndReplacingOldSubstituter` during inlining. Specifically, consider the case of inlining the following procedure:

```Boogie
procedure {:inline 1} b() returns () {
    var v:int where v > 3;
    havoc v;
}
```

Boogie would previously introduce a new variable into the caller declared as `var inline$b$0$v: int where v > 3;`, which causes resolution errors upon trying to run Boogie on that procedure, since `v` is not in scope. Instead, the variable should be declared as `var inline$b$0$v: int where inline$b$0$v > 3;`. I add this example as a new test case.

As far as I can tell, this bug does not manifest itself during verification phase but it does cause incorrect code being printed, if you use the `/printInlined` attribute.


Test/inline/testWhereExpr.bpl.expect
```diff
@@ -0,0 +1,59 @@
...
+implementation a()
+{
+  var inline$b$0$v: int where inline$b$0$v > 3;
+
+  anon0:
+    goto inline$b$0$Entry;
+
+  inline$b$0$Entry:
+    havoc inline$b$0$v /* where inline$b$0$v > 3 */;
+    assume inline$b$0$v > 3;
```
shazqadeer, line 43: It seems that this assume is injected automatically by the inlining implementation. Is that correct?

Test/inline/testWhereExpr.bpl.expect
```diff
@@ -0,0 +1,59 @@
...
+    goto inline$b$0$anon0;
+
+  inline$b$0$anon0:
+    havoc inline$b$0$v /* where inline$b$0$v > 3 */;
+    goto inline$b$0$Return;
```
shazqadeer, line 48: Strange! Why is an assume not injected here?

Test/inline/testWhereExpr.bpl.expect
Dargones: Yes, I think [these lines](https://github.com/boogie-org/boogie/blob/1969ad0f9b8c8d5de4aa5f8021d2364e4fcb366a/Source/Core/Inline.cs#L657-L688) are responsible for doing this. The justification for adding `havoc` is that the procedure might be invoked in a loop and so the values of its local variables would have to be reset on each iteration. There are `FIXME` comments surrounding the  added `assume` and I think they might be technically redundant but I am not confident enough in the semantics.

Test/inline/testWhereExpr.bpl.expect
Dargones: I am not sure. My suspicion is that an assumption would be redundant both here and above but I have not fully convinced myself yet. Empirically, it seems that `havoc` must respect the type constraint by default, e.g. the code below verifies. I will make another PR if I figure it out.

```Boogie
procedure b() {
    var d:int where d > 3;
    havoc d;
    assert d > 3;
}
```

shazqadeer: Here is a conjecture. 

- It is indeed redundant to put the assumes after the havocs in the inlining code because some processing later on (possibly VC generation) takes care of handling the where clauses after havoc appropriately.
- The bug in the inlining code you uncovered turned off the treatment above because the where clause was not appropriately substituted.
- The bug was discovered by somebody who did not understand the root cause and chose to fix the problem by putting the assumes in the inlining code itself.

If my conjecture is correct, then after your fix has landed, it ought to be possible to delete the assume injection. Of course, we have to make sure that the regression tests are adequate to check the behavior. You might need to add more tests.

I can approve your PR if you are ready to land it.

Dargones: Thank you, @shazqadeer! I think you are right, let me do some testing to make sure this is the case and then make another PR about the `assume`. I am ready to land this one!

shazqadeer (dismissed):

rakamaric (dismissed):
LGTM

rakamaric (approved):
LGTM

PR #763, opened 2023-07-17
Add sample for building mutex over a futex

shazqadeer:
This PR adds a sample to implement a mutex over a futex. In the processing of writing this sample, I discovered that the field YieldingLoops which is now part of the AST is not being duplicated properly. The discovery of this bug led to cleanup in the code for the new AST nodes in StandardVisitor and Duplicator.

YieldingLoops rightfully belongs to the class Implementation. At the moment, it is left in the class YieldProcedureDecl because Implementation already has a lot of clutter in it related to processing of VCs and DoomedVCs. It would be good to separate out this processing from the stuff that is related to the AST aspect of the Implementation class.

### DIFF
--- a/Source/Concurrency/LinearTypeChecker.cs
+++ b/Source/Concurrency/LinearTypeChecker.cs
@@ -84,7 +84,11 @@ namespace Microsoft.Boogie
       })));
 
       var oldErrorCount = checkingContext.ErrorCount;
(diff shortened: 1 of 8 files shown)


shazqadeer:
@afd : I am going to land this PR. You are welcome to provide feedback and ask questions about the new sample any time.

PR #764, opened 2023-07-18
Allow absolute paths to prover plugins

atomb:
Previously, the `/proverDll` option allowed only relative paths, since it looked for `/` or `\` _within_ a name, but not at the beginning, to determine whether the given name was a filename or a portion of an internal DLL name. This changes it to also interpret a name as a filename when either of those characters is the first character.


shazqadeer:
I understand that on a Linux/MacOS machine, a full path starts with `/`. What is the appropriate way to identify a full path on Windows?

atomb:
Most full paths on Windows will start with drive names (e.g., `c:\dir1\dir2\plugin.dll`), and those would already be allowed before this change. However, I believe that Windows paths starting with a backslash are valid, and that network paths are usually formatted that way, so allowing an initial backslash seems worthwhile.

shazqadeer (approved):

PR #767, opened 2023-08-03
Fixes lost dependencies for pruning during monomorphization.

zafer-esen:
This PR attempts to fix more axioms being pruned with the `/prune` flag than intended with monomorphization. 

Monomorphization introduces new axioms that are monomorphized, and tries splitting all axioms (polymorphic or not) into their conjuncts. This PR adds logic to update functions and constants to point to the new axioms as dependencies.

### DIFF
--- a/Source/Core/AST/Absy.cs
+++ b/Source/Core/AST/Absy.cs
@@ -1429,23 +1429,7 @@ namespace Microsoft.Boogie
     // from all other constants.
     public readonly bool Unique;
 
-    public IList<Axiom> DefinitionAxioms => definitionAxioms;
-
-    private IList<Axiom> definitionAxioms = new List<Axiom>();
-
-    public bool RemoveDefinitionAxiom(Axiom axiom)
-    {
-      Contract.Requires(axiom != null);
-
-      return definitionAxioms.Remove(axiom);
-    }
-
-    public void AddDefinitionAxiom(Axiom axiom)
-    {
-      Contract.Requires(axiom != null);
-
-      DefinitionAxioms.Add(axiom);
-    }
+    public IList<Axiom> DefinitionAxioms { get; set; }
 
     public Constant(IToken /*!*/ tok, TypedIdent /*!*/ typedIdent)
       : this(tok, typedIdent, true)
@@ -1468,7 +1452,7 @@ namespace Microsoft.Boogie
       Contract.Requires(typedIdent.Name != null && typedIdent.Name.Length > 0);
       Contract.Requires(typedIdent.WhereExpr == null);
       this.Unique = unique;
-      this.definitionAxioms = definitionAxioms ?? new List<Axiom>();
+      this.DefinitionAxioms = definitionAxioms ?? new List<Axiom>();
     }
 
     public override bool IsMutable => false;
--- a/Source/Core/AST/Absy.cs
+++ b/Source/Core/AST/Absy.cs
@@ -1429,7 +1429,23 @@ namespace Microsoft.Boogie
     // from all other constants.
     public readonly bool Unique;
 
-    public IList<Axiom> DefinitionAxioms { get; }
+    public IList<Axiom> DefinitionAxioms => definitionAxioms;
+
+    private IList<Axiom> definitionAxioms = new List<Axiom>();
+
+    public bool RemoveDefinitionAxiom(Axiom axiom)
+    {
+      Contract.Requires(axiom != null);
+
+      return definitionAxioms.Remove(axiom);
+    }
+
+    public void AddDefinitionAxiom(Axiom axiom)
+    {
+      Contract.Requires(axiom != null);
+
+      DefinitionAxioms.Add(axiom);
+    }
 
     public Constant(IToken /*!*/ tok, TypedIdent /*!*/ typedIdent)
       : this(tok, typedIdent, true)
@@ -1452,7 +1468,7 @@ namespace Microsoft.Boogie
       Contract.Requires(typedIdent.Name != null && typedIdent.Name.Length > 0);
       Contract.Requires(typedIdent.WhereExpr == null);
       this.Unique = unique;
-      this.DefinitionAxioms = definitionAxioms ?? new List<Axiom>();
+      this.definitionAxioms = definitionAxioms ?? new List<Axiom>();
     }
 
     public override bool IsMutable => false;
@@ -2056,6 +2072,13 @@ namespace Microsoft.Boogie
       Contract.Requires(axiom != null);
 
       otherDefinitionAxioms.Add(axiom);
+    }
+
+    public bool RemoveOtherDefinitionAxiom(Axiom axiom)
+    {
+      Contract.Requires(axiom != null);
+
+      return otherDefinitionAxioms.Remove(axiom);
     }
 
     private bool neverTrigger;
--- a/Source/Core/AST/Absy.cs
+++ b/Source/Core/AST/Absy.cs
@@ -2045,25 +2045,9 @@ namespace Microsoft.Boogie
     public NAryExpr DefinitionBody; // Only set if the function is declared with {:define}
     public Axiom DefinitionAxiom;
 
-    public IList<Axiom> otherDefinitionAxioms = new List<Axiom>();
+    public IList<Axiom> OtherDefinitionAxioms = new List<Axiom>();
     public IEnumerable<Axiom> DefinitionAxioms => 
-      (DefinitionAxiom == null ? Enumerable.Empty<Axiom>() : new[]{ DefinitionAxiom }).Concat(otherDefinitionAxioms);
-
-    public IEnumerable<Axiom> OtherDefinitionAxioms => otherDefinitionAxioms;
-
-    public void AddOtherDefinitionAxiom(Axiom axiom)
-    {
-      Contract.Requires(axiom != null);
-
-      otherDefinitionAxioms.Add(axiom);
-    }
-
-    public bool RemoveOtherDefinitionAxiom(Axiom axiom)
-    {
-      Contract.Requires(axiom != null);
-
-      return otherDefinitionAxioms.Remove(axiom);
-    }
+      (DefinitionAxiom == null ? Enumerable.Empty<Axiom>() : new[]{ DefinitionAxiom }).Concat(OtherDefinitionAxioms);
 
     private bool neverTrigger;
     private bool neverTriggerComputed;
--- a/Source/Core/Monomorphization.cs
+++ b/Source/Core/Monomorphization.cs
@@ -650,8 +650,8 @@ namespace Microsoft.Boogie
         else
         {
           var newAxiom = new Axiom(Token.NoToken, expr);
-          if(monomorphizationVisitor.oldAxiomsToSplitAxioms.ContainsKey(axiom)) {
-            monomorphizationVisitor.oldAxiomsToSplitAxioms[axiom].Add(newAxiom);
+          if (monomorphizationVisitor.originalAxiomToSplitAxioms.ContainsKey(axiom)) {
+            monomorphizationVisitor.originalAxiomToSplitAxioms[axiom].Add(newAxiom);
           }
           splitAxioms.Add(newAxiom);
         }
@@ -1299,9 +1299,9 @@ namespace Microsoft.Boogie
     private Dictionary<Procedure, Implementation> procToImpl;
     public readonly HashSet<Function> originalFunctions;
     public readonly HashSet<Constant> originalConstants;
-    // Note that old axioms refer to axioms of the original program which might have been updated in-place during monomorphization.
-    public readonly Dictionary<Axiom, HashSet<Axiom>> oldAxiomsToSplitAxioms;
-    public readonly Dictionary<Axiom, Axiom> oldAxiomsToOriginalCopies;
+    // Note that original axioms refer to axioms of the original program which might have been updated in-place during monomorphization.
+    public readonly Dictionary<Axiom, HashSet<Axiom>> originalAxiomToSplitAxioms;
+    public readonly Dictionary<Axiom, Axiom> originalAxiomToUninstantiatedCopies;
 
     private MonomorphizationVisitor(CoreOptions options, Program program, HashSet<Axiom> polymorphicFunctionAxioms)
     {
@@ -1309,8 +1309,8 @@ namespace Microsoft.Boogie
       this.program = program;
       originalFunctions = program.Functions.ToHashSet();
       originalConstants = program.Constants.ToHashSet();
-      oldAxiomsToSplitAxioms = program.Axioms.ToDictionary(ax => ax, _ => new HashSet<Axiom>());
-      oldAxiomsToOriginalCopies = program.Axioms.ToDictionary(ax => ax, ax => (Axiom) ax.Clone() );
+      originalAxiomToSplitAxioms = program.Axioms.ToDictionary(ax => ax, _ => new HashSet<Axiom>());
+      originalAxiomToUninstantiatedCopies = program.Axioms.ToDictionary(ax => ax, ax => (Axiom) ax.Clone() );
       implInstantiations = new Dictionary<Implementation, Dictionary<List<Type>, Implementation>>();
       nameToImplementation = new Dictionary<string, Implementation>();
       program.TopLevelDeclarations.OfType<Implementation>().Where(impl => impl.TypeParameters.Count > 0).ForEach(
@@ -1406,10 +1406,10 @@ namespace Microsoft.Boogie
       // those original axioms as dependencies need their references updated to the new axioms.
       // Axioms / functions could both have been instantiated, which adds some complexity.
 
-      foreach (var (oldAxiom, newAxioms) in monomorphizationVisitor.oldAxiomsToSplitAxioms)
+      foreach (var (oldAxiom, newAxioms) in monomorphizationVisitor.originalAxiomToSplitAxioms)
       {
         var oldAxiomFC = new FunctionAndConstantCollector();
-        var unmodifiedOldAxiom = monomorphizationVisitor.oldAxiomsToOriginalCopies[oldAxiom];
+        var unmodifiedOldAxiom = monomorphizationVisitor.originalAxiomToUninstantiatedCopies[oldAxiom];
         oldAxiomFC.Visit(unmodifiedOldAxiom);
         var oldAxiomFunctions = oldAxiomFC.Functions;
 
@@ -1616,8 +1616,8 @@ namespace Microsoft.Boogie
     {
       var axiomExpr = InstantiateBinderExpr((BinderExpr)axiom.Expr, actualTypeParams);
       var instantiatedAxiom = new Axiom(axiom.tok, axiomExpr, axiom.Comment, axiom.Attributes);
-      if(oldAxiomsToSplitAxioms.ContainsKey(axiom)) {
-        oldAxiomsToSplitAxioms[axiom].Add(instantiatedAxiom);
+      if (originalAxiomToSplitAxioms.ContainsKey(axiom)) {
+        originalAxiomToSplitAxioms[axiom].Add(instantiatedAxiom);
       }
       newInstantiatedDeclarations.Add(instantiatedAxiom);
       return instantiatedAxiom;
--- a/Source/Core/Monomorphization.cs
+++ b/Source/Core/Monomorphization.cs
@@ -343,7 +343,7 @@ namespace Microsoft.Boogie
      */
 
     protected MonomorphizationVisitor monomorphizationVisitor;
-    protected Dictionary<List<Type>, Expr> instanceExprs;
+    public Dictionary<List<Type>, Expr> instanceExprs { get; }
 
     public BinderExprMonomorphizer(MonomorphizationVisitor monomorphizationVisitor)
     {
@@ -615,6 +615,23 @@ namespace Microsoft.Boogie
     // be applied.
     private List<Axiom> splitAxioms;
 
+    private Dictionary<Expr, List<Type>> binderExprToTypes = null;
+    public List<Type> lookupTypeForBinderExpr(Expr binderExpr)
+    {
+      if (binderExprToTypes == null)
+      {
+        binderExprToTypes = new Dictionary<Expr, List<Type>>();
+        foreach (var binderMonomorphizer in binderExprMonomorphizers.Values)
+        {
+          foreach ((var types, var expr) in binderMonomorphizer.instanceExprs)
+          {
+            binderExprToTypes.TryAdd(expr, types);
+          }
+        }
+      }
+      return binderExprToTypes.TryGetValue(binderExpr, out var res) ? res : new List<Type>();
+    }
+
     public PolymorphicMapAndBinderSubstituter(MonomorphizationVisitor monomorphizationVisitor)
     {
       this.monomorphizationVisitor = monomorphizationVisitor;
@@ -646,6 +663,10 @@ namespace Microsoft.Boogie
         {
           stack.Push(nAryExpr.Args[0]);
           stack.Push(nAryExpr.Args[1]);
+        }
+        else if (expr == axiom.Expr)
+        {
+          splitAxioms.Add(axiom);
         }
         else
         {
@@ -1379,6 +1400,28 @@ namespace Microsoft.Boogie
           polymorphicMapInfo.CreateDatatypeTypeCtorDecl(polymorphicMapAndBinderSubstituter)).ToList();
       var splitAxioms = polymorphicMapAndBinderSubstituter.Substitute(program);
       program.RemoveTopLevelDeclarations(decl => decl is Axiom);
+      foreach (var axiom in splitAxioms)
+      {
+        if (axiom.FunctionDependencies != null)
+        {
+          var functionDependenciesCopy = new List<Function>(axiom.FunctionDependencies);
+          foreach (var function in functionDependenciesCopy)
+          {
+            if (monomorphizationVisitor.functionInstantiations.TryGetValue(function, out var instFuncsMap))
+            {
+              var types = polymorphicMapAndBinderSubstituter.lookupTypeForBinderExpr(axiom.Expr);
+              if (types.Any() && instFuncsMap.ContainsKey(types))
+              {
+                var newFunction = instFuncsMap[types];
+                if (!newFunction.DefinitionAxioms.Contains(axiom))
+                {
+                  newFunction.AddOtherDefinitionAxiom(axiom);
+                }
+              }
+            }
+          }
+        }
+      }
       program.AddTopLevelDeclarations(splitAxioms);
       program.AddTopLevelDeclarations(polymorphicMapDatatypeCtorDecls);
       Contract.Assert(MonomorphismChecker.IsMonomorphic(program));
--- a/Source/Core/Monomorphization.cs
+++ b/Source/Core/Monomorphization.cs
@@ -7,7 +7,7 @@ using Microsoft.Boogie.GraphUtil;
 
 namespace Microsoft.Boogie
 {
-  public class MonomorphismChecker : ReadOnlyVisitor
+  sealed public class MonomorphismChecker : ReadOnlyVisitor
   {
     // This visitor checks if the program is already monomorphic.
 
@@ -131,11 +131,13 @@ namespace Microsoft.Boogie
      * This visitor checks if the program is monomorphizable. It calculates one status value from
      * the enum MonomorphizableStatus for the program.
      */
-    public static MonomorphizableStatus IsMonomorphizable(Program program, out HashSet<Axiom> polymorphicFunctionAxioms)
+    public static MonomorphizableStatus IsMonomorphizable(Program program, out HashSet<Axiom> polymorphicFunctionAxioms,
+      out Dictionary<Axiom, (HashSet<Function>, HashSet<Constant>)> axiomsThatWillLoseQuantifiers)
     {
       var checker = new MonomorphizableChecker(program);
       checker.VisitProgram(program);
       polymorphicFunctionAxioms = checker.polymorphicFunctionAxioms;
+      axiomsThatWillLoseQuantifiers = checker.axiomsThatWillLoseQuantifiers;
       if (!checker.isMonomorphizable)
       {
         return MonomorphizableStatus.UnhandledPolymorphism;
@@ -148,7 +150,9 @@ namespace Microsoft.Boogie
     }
     
     private bool isMonomorphizable;
+    private Axiom curAxiom;
     private HashSet<Axiom> polymorphicFunctionAxioms;
+    private Dictionary<Axiom, (HashSet<Function>, HashSet<Constant>)> axiomsThatWillLoseQuantifiers;
     private Graph<TypeVariable> typeVariableDependencyGraph; // (T,U) in this graph iff T flows to U
     private HashSet<Tuple<TypeVariable, TypeVariable>> strongDependencyEdges; // (T,U) in this set iff a type constructed from T flows into U
 
@@ -158,6 +162,8 @@ namespace Microsoft.Boogie
       polymorphicFunctionAxioms = program.TopLevelDeclarations.OfType<Function>()
         .Where(f => f.TypeParameters.Count > 0 && f.DefinitionAxiom != null)
         .Select(f => f.DefinitionAxiom).ToHashSet();
+      axiomsThatWillLoseQuantifiers = new Dictionary<Axiom, (HashSet<Function>, HashSet<Constant>)>();
+      curAxiom = null;
       typeVariableDependencyGraph = new Graph<TypeVariable>();
       strongDependencyEdges = new HashSet<Tuple<TypeVariable, TypeVariable>>();
     }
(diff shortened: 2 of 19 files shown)


shazqadeer:
I am afraid we have to hit the brakes here. As I have indicated in the discussion on issue #560 , I don't understand parts of the implementation of pruning. This PR will leak some of that poorly-understood complexity into the implementation of monomorphization as well.

I have a question and a suggestion for your consideration.

First, the question:

Suppose an edge is explicitly indicated from a function to an axiom via `uses`. Suppose also that the function is polymorphic and the axiom has polymorphic quantifiers in it. Then, both the function and the axiom will be instantiated. What is the appropriate way of defining edges between the instantiations?

Second, the suggestion:

The most difficult to understand part of the current implementation of pruning is the automatic generation of incoming edges into axioms through the analysis of triggers. I suggest we drop this feature, thereby requiring the user of Boogie to do some analysis itself and put the appropriate annotations explicitly in Boogie code.

keyboardDrummer:
> There were two main issues: Monomorphization did not preserve uses clauses of constants and functions.

Are you saying it removes uses clauses from functions and constants? That seems unlikely to me. Can you clarify what you meant?

> In quantified axioms outside of uses clauses, pruning depends on triggers for computing incoming edges. Since
expressions of the form forall <T> :: {...} and exists <T> :: {...} become unquantified after monomorphization, they were being pruned.

Can you provide an example that shows which uses clauses are missing in the monomorphized code?

The way I understood the problem is that monomorphization creates new axioms based on existing ones, and does not copy over the uses clauses for the created ones.

zafer-esen:
Thank you both for the reviews @shazqadeer and @keyboardDrummer! I was hoping the PR would initiate exactly this kind of discussion.

> I don't understand parts of the implementation of pruning. This PR will leak some of that poorly-understood complexity into the implementation of monomorphization as well.

I agree, there seem to be some assumptions the pruning algorithm makes, and these seem to have changed over time since its initial implementation.

> Suppose an edge is explicitly indicated from a function to an axiom via `uses`. Suppose also that the function is polymorphic and the axiom has polymorphic quantifiers in it. Then, both the function and the axiom will be instantiated. What is the appropriate way of defining edges between the instantiations?

This was something I was hoping to get your opinions on. Let's call the polymorphic function `f`, its instantiated instances `f_0`...`f_m`, the original axiom `a`, and its instantiated instances `a_0`...`a_n` with `n >= m`. Currently, the edges are computed as follows:

- if `f` **appears** as a symbol anywhere in `a`, then an edge from `f_i` (`i = 0...m`) to `a_j` (`j = 0...n`) is only added if `f_i` appears as a symbol anywhere in `a_j`.
- if `f` **does not appear** as a symbol in `a`, then edges from `f_i` (`i = 0...m`) to all `a_0`...`a_n` are added.

There are also axioms that do not contain any polymorphic quantifiers in it, but these axioms are still split into its conjuncts. The same algorithm is applied here

> The most difficult to understand part of the current implementation of pruning is the automatic generation of incoming edges into axioms through the analysis of triggers. I suggest we drop this feature, thereby requiring the user of Boogie to do some analysis itself and put the appropriate annotations explicitly in Boogie code.

I agree that this would be a good solution, and make the fix for the second issue mentioned in the PR unnecessary. It would also prevent the leak from the pruning algorithm into the monomorphization code, since I think this fix is the only part that depends on assumptions from the pruning algorithm (i.e., symbols inside triggers are considered automatically as incoming edges to axioms). However, I did not want to change anything in the pruning code since that would require changes in front-end tools that use pruning (only Dafny?).

To summarize, this PR attempts to: 

- Re-establish `uses` edges that are lost during monomorphization, I think this should be done in either case. Your feedback would be appreciated on how to best do this in the presence of instantiated functions and axioms!
- Re-establish edges that were automatically added by the pruning algorithm, namely the incoming edges of axioms due to triggers. This can be dropped, but I am not sure of all its implications in the front-end tools that use pruning.

zafer-esen:
> Are you saying it removes uses clauses from functions and constants? That seems unlikely to me. Can you clarify what you meant?

What I meant was `uses` clauses were not preserved by the monomorphization code due to new axioms being generated ([here](https://github.com/boogie-org/boogie/blob/0b1e6843c94480df8e05790f43ae4182e52e3204/Source/Core/Monomorphization.cs#L652) and [here](https://github.com/boogie-org/boogie/blob/0b1e6843c94480df8e05790f43ae4182e52e3204/Source/Core/Monomorphization.cs#L1494), although I do not think the latter ever triggers since `uses` clauses add axioms to `OtherDefinitionAxioms` of functions).

> Can you provide an example that shows which uses clauses are missing in the monomorphized code?
[Here](https://github.com/dafny-lang/dafny/blob/df505adb6dbf893e9a546ae7dadda803bb16ba13/Source/DafnyCore/DafnyPrelude.bpl#L971) is an example from Dafny:
`axiom (forall<T> :: { Seq#Empty(): Seq T } Seq#Length(Seq#Empty(): Seq T) == 0);`

Putting this inside the `uses` clause of `function Seq#Empty<T>(): Seq T;` would fix it in this case. I am not aware of any others in Dafny code. Still, I wasn't sure if such axioms would show up in other places, and wanted to have a fix for it in line with the current assumptions of the pruning code (i.e., that symbols from triggers are automatically considered as incoming edges).


> The way I understood the problem is that monomorphization creates new axioms based on existing ones, and does not copy over the uses clauses for the created ones.

Yes, this is the first issue mentioned in this PR. It is a bit tricky as both functions and axioms can be instantiated, and even non-instantiated axioms are split into their conjuncts during monomorphization. This is independent from above discussion in this comment however.

Source/Core/AST/Absy.cs
```diff
@@ -1429,7 +1429,23 @@ public class Constant : Variable
     // from all other constants.
     public readonly bool Unique;
 
-    public IList<Axiom> DefinitionAxioms { get; }
+    public IList<Axiom> DefinitionAxioms => definitionAxioms;
+
+    private IList<Axiom> definitionAxioms = new List<Axiom>();
+
+    public bool RemoveDefinitionAxiom(Axiom axiom)
```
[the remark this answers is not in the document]
zafer-esen: Good point, I can change to use those instead.

Source/Core/AST/Absy.cs
```diff
@@ -1429,7 +1429,23 @@ public class Constant : Variable
     // from all other constants.
     public readonly bool Unique;
 
-    public IList<Axiom> DefinitionAxioms { get; }
+    public IList<Axiom> DefinitionAxioms => definitionAxioms;
+
+    private IList<Axiom> definitionAxioms = new List<Axiom>();
```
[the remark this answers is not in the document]
zafer-esen: Good point, will add a `set` to the auto property instead.

shazqadeer:
@zafer-esen : 

Your method for adding edges between instantiated function and instantiated axioms seems fine to me. Please make sure you document it somewhere in the Boogie code.

Please reduce the scope of this PR so that it addresses only issue #1 from your PR description.

Dafny is the only user of the `/prune` flag in Boogie. Please examine the feasibility of Dafny not using the edge inference due to triggers. I strongly recommend deleting this aspect of the prune implementation in Boogie and supporting only explicit edges. My hope is that we would not have to address issue #2.

zafer-esen:
@shazqadeer Thank you for the feedback, I will add some documentation to the code for (1) and reduce the scope of the PR.

I discussed this with @atomb and will also look into the feasibility of removing edge inference using triggers. I think something like a `used_by` attribute rather than a `uses` clause is needed for this to work. This is because an axiom can have multiple triggers, with each trigger possibly referring to multiple symbols (example and documentation [here](https://github.com/boogie-org/boogie/blob/0b1e6843c94480df8e05790f43ae4182e52e3204/Source/VCGeneration/Prune/DependencyEvaluator.cs#L7-L20)).  If an axiom has the two triggers `{P, Q} {R}` as in the linked example, an attribute `{:used_by "P, Q; R"}` could be added.

One complication is that both symbol names and axioms can change during monomorphization, but I think this can be handled by updating the new axiom attributes using aforementioned method of adding edges between instantiated functions and axioms, possibly inside the proposed `UpdateDependencies` method in this PR.

shazqadeer:
I didn't understand why both `uses` and `used_by` would be needed. Can you provide a concrete example?

zafer-esen:
Each declaration can have both incoming and outgoing declarations. I was basing my understanding on [this comment](https://github.com/boogie-org/boogie/blob/0b1e6843c94480df8e05790f43ae4182e52e3204/Source/VCGeneration/Prune/DependencyEvaluator.cs#L19-L20):
> a declaration A depends on B, if the outgoing dependents of A match with some incoming dependent of B (see method depends).

This comment is out of date. The method `Depends` was deleted in PR #442, and the semantics seems to have changed. In the current pruning algorithm, if I am interpreting this correctly, an edge between two nodes is established if one of the two nodes has the other one in its dependencies. In this semantics, `used_by` is not needed for axioms inside `uses` clauses. Since this is the case, `uses` could be completely removed and subsumed by `used_by`, which is more expressive. I make a case for this below.

Here are some examples to explain the current semantics:
**Example 1:**
```
function foo() : int;
axiom foo() == 42; // label: a
procedure p()
{
  assert foo() == 42;
}
```
Here the procedure `p` is at the root, and has `foo` is an _outgoing_ dependency. Let's denote this `p -> f` (`p` depends on `f`). `foo` has no dependencies. Axiom `a`, having no triggers to establish incoming dependencies, has only `foo` as its _outgoing_ dependency. So we have the following set of dependencies:
`{p -> foo, a -> foo}`
This implies `a` gets pruned away, because it is not reachable from the root node `p`.

If we add axiom `a` inside the `uses` clause of `foo`, the dependencies become:
`{p -> foo -> a, a -> foo}`

Since `a` is now reachable, it is not pruned away.

**Example 2**
```
function foo(x: int) : int;
axiom (forall x: int :: {foo(x)} foo(x) == 42); // label: a
procedure p()
{
  assert foo(0) == 42;
}
```
Now the axiom is quantified and has a trigger, so automatic edge inference is applied. This builds the dependencies as:
`{p -> foo, foo -> a -> foo}`
I.e., `foo` is added as an _incoming_ edge to `a`. In the current semantics, the edges computed from this set is equivalent to `{p -> foo -> a, a -> foo}`.

Still, triggers are more expressive than `uses` clauses, so a `uses` clause is not sufficient by itself to replace the edge inference. I attempt to explain this in the next example.

**Example 3:**
```
function bar(x: int) : int;
function foo(x: int) : int;

axiom (forall x: int :: {foo(bar(x))} foo(bar(x)) == 42); // label: a
axiom (forall x: int :: {bar(x)} bar(x) == 0); // label: b

procedure p()
{
  assert foo(0) == 42;
}
```
Here the trigger of axiom `a` contains two function symbols. If the trigger contains more than one function or constant symbol, its dependencies are tuples, and all symbols inside the tuple must be reachable from the root for the axiom to be preserved. In this case the dependencies are:
```
{p -> foo, (foo, bar) -> a -> foo, b -> bar}
                           -> bar
```
Here `bar` is unreachable from the root `p`, hence `a` is pruned away. If we only add a call to `bar` inside `p`, the graph would change to
```
{p -> foo, (foo, bar) -> a -> foo, b -> bar}
   -> bar                  -> bar
```
and `a` and `b` would be both kept, and the assertion would verify.

One might at this point wonder what this expressiveness brings; one could just put both axioms inside a `uses` clause, and they both would be preserved anyway. Consider the following [axioms from the Dafny prelude](https://github.com/dafny-lang/dafny/blob/c31932b4b4f77a38e5da9c9f6a7689d8f57346bf/Source/DafnyCore/DafnyPrelude.bpl#L230-L290), which does exactly this:
```
function $Is<T>(T,Ty): bool uses {           // no heap for now
    axiom(forall v : int  :: { $Is(v,TInt) }  $Is(v,TInt));
    axiom(forall v : real :: { $Is(v,TReal) } $Is(v,TReal));
    axiom(forall v : bool :: { $Is(v,TBool) } $Is(v,TBool));
    axiom(forall v : char :: { $Is(v,TChar) } $Is(v,TChar));
    axiom(forall v : ORDINAL :: { $Is(v,TORDINAL) } $Is(v,TORDINAL));
    [...] // more axioms for every type
 }
```
Here `$Is` is a function symbol, and `TInt`, `TReal` etc. are constants.
As soon as `$Is` is reachable from any root node, none of these axioms can be pruned away. Furthermore, all symbols appearing in their bodies are also included in the dependency graph, so all axioms inside the `uses` clauses of those symbols are transitively included.

I tried moving these axioms out of the `uses` clause, and in my limited experiments I saw improvements in solver resource counts. (Takeaway: in the current semantics, `uses` should not be used for any axiom with more than one symbol in any of its triggers.) This improvement is achieved by the current edge inference, so it would be good to keep its expressiveness around via an attribute like `used_by` that would allow tuple dependencies. 

Since `used_by` subsumes the expressiveness of `uses` and can be applied to any axiom with or without triggers, I propose removing `uses` completely, as its incorrect usage leads to more axioms being preserved than necessary.

zafer-esen:
I had the same idea with Examples 1 & 2. I had a slightly different idea in Example 3, but maybe this was not the best example: it cannot be verified even without pruning as the axioms don't trigger without both `foo` and `bar`! However, I think it still serves to illustrate the difference between `uses` and `used_by`.

If `used_by` is to exactly have the current edge inference semantics (which seems to be in line with how triggers work), Example 3 would be as follows:
```
function bar(x: int) : int;
function foo(x: int) : int;

axiom (forall x: int :: {foo(bar(x))} foo(bar(x)) == 42) used_by (foo, bar); // label: a
axiom (forall x: int :: {bar(x)} bar(x) == 0) used_by bar; // label: b

procedure p()
{
  assert foo(0) == 42;
}
```
i.e., make all symbols in the trigger `used_by` (added parentheses to denote a tuple as there can be multiple triggers). This still prunes away both axioms, but they would not be used anyway due to axioms not triggering. 

Contrast this to using a `uses` clause:
```
function bar(x: int) : int;
function foo(x: int) : int uses {
  axiom (forall x: int :: {foo(bar(x))} foo(bar(x)) == 42);
  axiom (forall x: int :: {bar(x)} bar(x) == 0); // label: b
}

procedure p()
{
  assert foo(0) == 42;
}
```
In this case no pruning happens, but the program also does not verify, because `a` and `b` are still useless.

> Is it also correct that when we move to used_by, the algorithm for determining the set of included declarations will be vanilla graph reachability? I want to note here that the current algorithm is not simple reachability.

I am not sure about this, I think the algorithm remains similar? Reachability would still be computed from the roots, but with `uses_by`, axioms would be connected to the graph only through their incoming edges. Removing `uses` has the effect of removing outgoing edges from functions and constants. Maybe @keyboardDrummer can chime in.

> I suppose the used_by clause qualifying an axiom may contain function and constant names. Could it also contain procedure names?

I think it could contain any named declaration.

> Can we also get rid of :include_dep?

Yes, in axioms without triggers, `include_dep` was adding all body symbols as incoming. The same could be achieved by adding a `used_by` clause and listing all symbols. In axioms with triggers, I believe it had no effect.

zafer-esen:
I reduced the scope of this PR as requested. If you think adding `used_by` would make sense, it can perhaps be another PR.

Source/Core/Monomorphization.cs
```diff
@@ -1293,11 +1297,20 @@ class MonomorphizationVisitor : VarDeclOnceStandardVisitor
     private HashSet<TypeCtorDecl> visitedTypeCtorDecls;
     private HashSet<Function> visitedFunctions;
     private Dictionary<Procedure, Implementation> procToImpl;
+    public readonly HashSet<Function> originalFunctions;
+    public readonly HashSet<Constant> originalConstants;
+    // Note that old axioms refer to axioms of the original program which might have been updated in-place during monomorphization.
+    public readonly Dictionary<Axiom, HashSet<Axiom>> oldAxiomsToSplitAxioms;
```
shazqadeer, line 1303: Rename to originalAxiomToSplitAxioms

keyboardDrummer:
Without having read all of the above, 

I think adding `used_by`, and possibly removing `used`, can be useful, because then inferring dependencies based on triggers can be a compilation step where `used_by` declarations are added based on the triggers. If I understand correctly, this would prevent the need to complicate the monomorphization code.

> > Is it also correct that when we move to used_by, the algorithm for determining the set of included declarations will be vanilla graph reachability? I want to note here that the current algorithm is not simple reachability.

> I am not sure about this, I think the algorithm remains similar? Reachability would still be computed from the roots, but with uses_by, axioms would be connected to the graph only through their incoming edges. Removing uses has the effect of removing outgoing edges from functions and constants. Maybe @keyboardDrummer can chime in.

I don't think anything would change there.

Source/Core/Monomorphization.cs
```diff
@@ -1492,6 +1616,9 @@ private Axiom InstantiateAxiom(Axiom axiom, List<Type> actualTypeParams)
     {
       var axiomExpr = InstantiateBinderExpr((BinderExpr)axiom.Expr, actualTypeParams);
       var instantiatedAxiom = new Axiom(axiom.tok, axiomExpr, axiom.Comment, axiom.Attributes);
+      if(oldAxiomsToSplitAxioms.ContainsKey(axiom)) {
```
shazqadeer, line 1619: if (

Source/Core/Monomorphization.cs
```diff
@@ -649,7 +649,11 @@ public override Axiom VisitAxiom(Axiom node)
         }
         else
         {
-          splitAxioms.Add(new Axiom(Token.NoToken, expr));
+          var newAxiom = new Axiom(Token.NoToken, expr);
+          if(monomorphizationVisitor.oldAxiomsToSplitAxioms.ContainsKey(axiom)) {
```
shazqadeer, line 653: if (

Source/Core/Monomorphization.cs
```diff
@@ -1293,11 +1297,20 @@ class MonomorphizationVisitor : VarDeclOnceStandardVisitor
     private HashSet<TypeCtorDecl> visitedTypeCtorDecls;
     private HashSet<Function> visitedFunctions;
     private Dictionary<Procedure, Implementation> procToImpl;
+    public readonly HashSet<Function> originalFunctions;
+    public readonly HashSet<Constant> originalConstants;
+    // Note that old axioms refer to axioms of the original program which might have been updated in-place during monomorphization.
```
shazqadeer, line 1302: Are you sure old axioms are updated in-place? If I remember correctly, each axiom could be instantiated multiple times so the instantiation is done on a duplicate version of the axiom which is modified in-place. The original axiom stays as is to facilitate the next instantiation. Could you please check?

shazqadeer (changes requested):

shazqadeer:
@zafer-esen : Thanks for explaining the tuple notation for each member of the `used_by` clause. I agree that with the flexibility of having a tuple in addition to singletons, the current reachability algorithm remains unchanged. `used_by` makes sense. It would be great to see `uses` replaced by `used_by` in a separate PR.

Source/Core/Monomorphization.cs
zafer-esen: Yes, I checked now and old axioms get instantiated in-place after executing [this line](https://github.com/boogie-org/boogie/blob/0b1e6843c94480df8e05790f43ae4182e52e3204/Source/Core/Monomorphization.cs#L627). I think [this comment](https://github.com/boogie-org/boogie/blob/0b1e6843c94480df8e05790f43ae4182e52e3204/Source/Core/Monomorphization.cs#L1239-L1243) also confirms that the old axioms are instantiated in-place (with the help of `MonomorphizationDuplicator`).

Source/Core/Monomorphization.cs
zafer-esen: Done (I also renamed the other dictionary).

Source/Core/Monomorphization.cs
```diff
@@ -343,7 +343,7 @@ abstract class BinderExprMonomorphizer
      */
 
     protected MonomorphizationVisitor monomorphizationVisitor;
-    protected Dictionary<List<Type>, Expr> instanceExprs;
+    public Dictionary<List<Type>, Expr> instanceExprs { get; }
```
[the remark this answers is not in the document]
zafer-esen: I got such warnings before, but I am not getting any for this one... In any case, I will push a commit fixing this!

Source/Core/Monomorphization.cs
```diff
@@ -1378,13 +1392,123 @@ public static MonomorphizationVisitor Initialize(CoreOptions options, Program pr
...
+          newAxiomFC.Visit(newAxiom);
+          newAxiomFunctions.Add(newAxiom, newAxiomFC.Functions);
+          newAxiomConstants.Add(newAxiom, newAxiomFC.Constants);
+        }
+
+        void UpdateFunctionDependencies(HashSet<Function> functions)
+        {
+          foreach (var function in functions)
+          {
+            if (function.DefinitionAxioms.Contains(oldAxiom))
```
keyboardDrummer, line 1431: Consider inverting the condition so you can do an early exit using continue, and reduce nesting of the remaining code.

Source/Core/Monomorphization.cs
```diff
@@ -1378,13 +1392,123 @@ public static MonomorphizationVisitor Initialize(CoreOptions options, Program pr
...
+
+        foreach (var newAxiom in newAxioms)
+        {
+          var newAxiomFC = new FunctionAndConstantCollector();
+          newAxiomFC.Visit(newAxiom);
+          newAxiomFunctions.Add(newAxiom, newAxiomFC.Functions);
+          newAxiomConstants.Add(newAxiom, newAxiomFC.Constants);
+        }
+
+        void UpdateFunctionDependencies(HashSet<Function> functions)
```
keyboardDrummer, line 1427: Adding the parameter `HashSet<Function> functions` seems unnecessary, since it's also available through `monomorphizationVisitor.originalFunctions`

I'm not always sure how much inner functions help read the code. Here I have to inspect the outer scope to understand the meaning of some of the variables used.

Source/Core/Monomorphization.cs
```diff
@@ -1378,13 +1392,123 @@ public static MonomorphizationVisitor Initialize(CoreOptions options, Program pr
...
+        {
+          foreach (var function in functions)
```
keyboardDrummer, line 1429: `function` -> `oldFunction `, `functions` -> `oldFunctions`

Source/Core/Monomorphization.cs
```diff
@@ -1378,13 +1392,123 @@ public static MonomorphizationVisitor Initialize(CoreOptions options, Program pr
...
       program.RemoveTopLevelDeclarations(decl => decl is Axiom);
       program.AddTopLevelDeclarations(splitAxioms);
       program.AddTopLevelDeclarations(polymorphicMapDatatypeCtorDecls);
       Contract.Assert(MonomorphismChecker.IsMonomorphic(program));
       return monomorphizationVisitor;
     }
 
+    private static void UpdateDependencies(MonomorphizationVisitor monomorphizationVisitor)
+    {
+      // Original program axioms were replaced with new splitAxioms. Program constants and functions that had
```
keyboardDrummer, line 1405: Could you move the comment to above the method, and use `///` ?

Source/Core/Monomorphization.cs
```diff
@@ -1378,13 +1392,123 @@ public static MonomorphizationVisitor Initialize(CoreOptions options, Program pr
...
+      // Axioms / functions could both have been instantiated, which adds some complexity.
+
+      foreach (var (oldAxiom, newAxioms) in monomorphizationVisitor.originalAxiomToSplitAxioms)
+      {
+        var oldAxiomFC = new FunctionAndConstantCollector();
+        var unmodifiedOldAxiom = monomorphizationVisitor.originalAxiomToUninstantiatedCopies[oldAxiom];
+        oldAxiomFC.Visit(unmodifiedOldAxiom);
+        var oldAxiomFunctions = oldAxiomFC.Functions;
+
+        Dictionary<Axiom, HashSet<Function>> newAxiomFunctions = new Dictionary<Axiom, HashSet<Function>>();
```
keyboardDrummer, line 1416: Could you use `var` here? Does your IDE not suggest that?

keyboardDrummer (commented):
Could you add tests to this PR?

Source/Core/Monomorphization.cs
zafer-esen: Done.

Source/Core/Monomorphization.cs
zafer-esen: Removed inner functions. I also changed `old` to `original` to match related dictionary names.

Source/Core/Monomorphization.cs
zafer-esen: Updated. My IDE was apparently suggesting that, which I didn't notice.

Source/Core/AST/Absy.cs
```diff
@@ -2058,6 +2058,13 @@ public void AddOtherDefinitionAxiom(Axiom axiom)
       otherDefinitionAxioms.Add(axiom);
     }
 
+    public bool RemoveOtherDefinitionAxiom(Axiom axiom)
```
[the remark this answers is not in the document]
zafer-esen: Done.

Source/Core/Monomorphization.cs
```diff
@@ -1293,11 +1297,20 @@ class MonomorphizationVisitor : VarDeclOnceStandardVisitor
     private HashSet<TypeCtorDecl> visitedTypeCtorDecls;
     private HashSet<Function> visitedFunctions;
     private Dictionary<Procedure, Implementation> procToImpl;
+    public readonly HashSet<Function> originalFunctions;
+    public readonly HashSet<Constant> originalConstants;
+    // Note that original axioms refer to axioms of the original program which might have been updated in-place during monomorphization.
+    public readonly Dictionary<Axiom, HashSet<Axiom>> originalAxiomToSplitAxioms;
+    public readonly Dictionary<Axiom, Axiom> originalAxiomToUninstantiatedCopies;
```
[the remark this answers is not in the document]
zafer-esen: How about this:
We add something like the following to `Absy.Axiom`
```
    public Axiom OriginalAxiom;
    public Axiom OriginalAxiomUninstantiatedCopy;
```
and update them during monomorphization. This is in line with instantiated `DeclWithFormals` [in Absy.cs](https://github.com/boogie-org/boogie/blob/0b1e6843c94480df8e05790f43ae4182e52e3204/Source/Core/AST/Absy.cs#L1680-L1682).

We could also add to `Absy.Axiom`:
```
    public ISet<Function> Functions;
    public ISet<Function> Constants;
```
which could return body functions & symbols for an axiom. Then we would no longer need the additional dictionaries. The function and constant collector could also be moved out somewhere, maybe to `Absy.cs`.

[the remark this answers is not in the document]
zafer-esen: I implemented the version I proposed above, but I am not sure if it ended up being any better. In the new implementation members of `Axiom` need to be initialized in the right places during monomorphization, as opposed to adding entries to the dictionaries. The dictionaries are no longer needed, but similar collections need to be computed again from `splitAxioms` inside `UpdateFunctionDependencies`. Maybe additional members in `Axiom`, which are only used during monomorphization, are also undesirable. Overall I am tempted to just change the name for now. This can perhaps be considered again when adding `used_by`.

zafer-esen:
@keyboardDrummer 

>>if f does not appear as a symbol in a, then edges from f_i (i = 0...m) to all a_0...a_n are added.

>Does this occur in practice? I wonder whether we could disallow such uses/used-by clauses.

I do not know of any use cases, none in the `DafnyPrelude` I think.

> Could you add tests to this PR?

I will add a few tomorrow.

Source/Core/Monomorphization.cs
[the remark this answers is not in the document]
shazqadeer: When I made the suggestion, I had something very simple in mind. You have the following lines of code in your PR:

```
var originalAxiomFc = new FunctionAndConstantCollector();
var uninstantiatedOriginalAxiom = monomorphizationVisitor.originalAxiomToUninstantiatedCopy[originalAxiom];
originalAxiomFc.Visit(uninstantiatedOriginalAxiom);
```

If I understand this correctly, you retrieving the cloned axiom (prior to in-place updates) just so you can call the `originalFc.Visit` method. You could call this method up front and stash originalAxiomFc in the dictionary instead. Then you can look up the collected functions and constants later.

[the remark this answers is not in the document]
zafer-esen: That makes sense, and axioms don't need to be cloned anymore. I just pushed a [commit](https://github.com/boogie-org/boogie/pull/767/commits/7b28f0e4ad1a73dfacdf59f68f0719db8caea47d) doing this, thanks for the suggestion!

shazqadeer (approved):

shazqadeer (dismissed):

Test/pruning/MonomorphicSplit.bpl
```diff
@@ -0,0 +1,37 @@
+// RUN: %parallel-boogie /prune /errorTrace:0 /printPruned:"%t" "%s" > "%t"
+// RUN: %OutputCheck "%s" --file-to-check="%t-after-monomorphicSplit.bpl"
+
+// The following checks are a bit simplistic, but this is
+// on purpose to reduce brittleness. We assume there would now be two uses clauses
+// with one axiom each, and those axioms should not be conjunctions
```
shazqadeer, line 6: How can an axiom be a conjunction of monomorphized functions? The types do not match. Please rephrase.

The last sentence of this paragraph is equally mysterious.

Test/pruning/MonomorphicSplit.bpl
```diff
@@ -0,0 +1,37 @@
...
+// Both f1 and the axiom will be monomorphized into bool and int instances.
+// Automatic edge inference would already ensure that only the monomorphized instances
+// are incoming, however we want to test here how the new uses clauses are determined.
+//
+// After instantiation, each instance of f1 for some T should *only* have 
+// the T-instantiated axiom as outgoing. 
+
+procedure monomorphicSplit()
+  ensures f1(true) == 42;
+	ensures f1(3) == 42;
```
shazqadeer, line 35: Any particular reason to indent the second ensures past the first one?

Test/pruning/NonmonomorphicSplit.bpl
```diff
@@ -0,0 +1,37 @@
...
+function f2 (x: int) : int uses
+{
+  axiom f1(0) == 1 && f2(0) == 2;
+}
+
+// Above axiom will be split into two during monomorphization into
+// axiom f1(0) == 1 and axiom f2(0) == 2.
+// Current implementation ensures that f1(0) == 1 is kept as a dependency
+// of f1, and f2(0) is moved to be a dependency of f2. If there would be
+// other symbols in the axiom, the new split axioms would be added to those
```
shazqadeer, line 16: I sort of understand what the last sentence is saying. But I am confident its meaning can be made more transparent. It is fine to define symbols and use them later to clarify what is being said.

Test/pruning/MonomorphicSplit.bpl
zafer-esen: Ah, this was not visible on my end in both Emacs and VS Code, apparently the second line was added as a tab (new laptop). Should be fixed now.

Test/pruning/NonmonomorphicSplit.bpl
zafer-esen: Thanks for the feedback, they were written a bit hastily... I tried updating both explanations based on the comments.

zafer-esen:
@keyboardDrummer  About your question from earlier:
> > if f does not appear as a symbol in a, then edges from f_i (i = 0...m) to all a_0...a_n are added.

> Does this occur in practice? I wonder whether we could disallow such uses/used-by clauses.

I was having a look at [this paper](https://dspace.mit.edu/bitstream/handle/1721.1/103950/Pit-Claudel_Trigger%20Selection.pdf), and I believe this is what the paper refers to as _trigger sharing_, and is sometimes useful.

Consider Example 3 from before (ignoring pruning for now):
```
function bar(x: int) : int;
function foo(x: int) : int;

axiom (forall x: int :: {foo(bar(x))} foo(bar(x)) == 42);
axiom (forall x: int :: {bar(x)} bar(x) == 0); // label: b

procedure p()
{
  assert foo(0) == 42;
}
```
This does not verify, because there is no terms that can trigger any of the axioms. However, if we add `foo(x)` as another trigger to the second axiom, i.e.,
```
axiom (forall x: int :: {bar(x)} {foo(x)} bar(x) == 0); // label: b
```
This verifies.

When we add pruning and monomorphization, If `foo` is polymorphic and if axiom `b` is in the `uses` clause of `foo`, the current implementation would add all instantiations of axiom `b` as dependencies to all instantiations of `foo`. I think this would be undesirable, if it ever happens. It should be possible to determine the correct instantiation of `foo` to match with some instantiation of `b` from its argument types. I can do this change in this PR for `uses` (might need to wait until next week), or it could be added in the proposed `used_by` PR.

keyboardDrummer:
> I was having a look at [this paper](https://dspace.mit.edu/bitstream/handle/1721.1/103950/Pit-Claudel_Trigger%20Selection.pdf), and I believe this is what the paper refers to as _trigger sharing_, and is sometimes useful.
> 
> Consider Example 3 from before (ignoring pruning for now):
> 
> ```
> function bar(x: int) : int;
> function foo(x: int) : int;
> 
> axiom (forall x: int :: {foo(bar(x))} foo(bar(x)) == 42);
> axiom (forall x: int :: {bar(x)} bar(x) == 0); // label: b
> 
> procedure p()
> {
>   assert foo(0) == 42;
> }
> ```
> 
> This does not verify, because there are no terms that can trigger any of the axioms. However, if we add `foo(x)` as another trigger to the second axiom, i.e.,
> 
> ```
> axiom (forall x: int :: {bar(x)} {foo(x)} bar(x) == 0); // label: b
> ```
> 
> This verifies.
> 

After thinking about this more, your initial suggestion (quoted below for clarity) seems perfect. 

>> - if `f` **appears** as a symbol anywhere in `a`, then an edge from `f_i` (`i = 0...m`) to `a_j` (`j = 0...n`) is only added if `f_i` appears as a symbol anywhere in `a_j`.
>> - if `f` **does not appear** as a symbol in `a`, then edges from `f_i` (`i = 0...m`) to all `a_0`...`a_n` are added.

keyboardDrummer (approved):

keyboardDrummer:
@shazqadeer if you approve this, could you merge it?

shazqadeer:
@zafer-esen : There was discussion in this PR for dropping `uses` and `include_dep` in favor of `used_by`. Is that still the plan?

zafer-esen:
There was a small discussion about this recently in a Dafny PR where I wrote the following in a [comment](https://github.com/dafny-lang/dafny/pull/4428#discussion_r1304422306):

> I am starting to think it might be best if edge inference from triggers is preserved. The only correct way to use used_by might be to add exactly one tuple for each trigger of an axiom, which the current edge inference is already doing. Decoupling the semantics would make it possible to strengthen pruning, but I have difficulty seeing any cases where this would be useful.

It would be great to know what you all think @shazqadeer, @atomb and @keyboardDrummer about this.

If the edge inference from triggers infers exactly the set of axioms that can be instantiated, maybe it makes sense to keep edge inference around. With this PR merged, `uses` clauses are only needed in axioms without triggers.

Another small concern is, unlike `uses` and `include_deps`, `used_by` also requires specifying symbols whose names would need to be updated during monomorphization.

keyboardDrummer:
> If the edge inference from triggers infers exactly the set of axioms that can be instantiated, maybe it makes sense to keep edge inference around. 

I think it does

> With this PR merged, uses clauses are only needed in axioms without triggers.

Sounds good

PR #770, opened 2023-08-10
Print verification coverage only when -trace used

atomb:
The -trackVerificationCoverage option enables tracking of verification coverage but the results are printed only when -trace is also provided. This makes it possible to see the results in text, but makes the output less noisy for clients of Boogie as a library (such as Dafny).


shazqadeer (approved):

keyboardDrummer (approved):

shazqadeer:
GC.bpl is timing out in one of the configurations. I have observed some flakiness in verification times for this example before.  I found that `/vcsSplitOnEveryAssert` helps.

shazqadeer:
I will look into the flakiness of GC.bpl later. Meanwhile, if you need to merge this PR now, is there a way to override the merge requirements?

atomb:
It's not super urgent that I merge this PR. I'd like to do it by the end of next week, perhaps, but before then isn't critical. So if you fix it as part of a separate PR, this one should then go through after merging with `master`.

shazqadeer:
@atomb : I pushed a fix to GC.bpl. Looks good now.

shazqadeer (approved):

PR #773, opened 2023-08-13
[Civl] bug fix for noninterference checking for pending asyncs

shazqadeer:
For a particular layer number, those pending asyncs must be collected for which noninterference checking against yield invariants must be performed. The collection of pending asyncs was not sound earlier. This PR makes a fix to make it sound.


Source/Concurrency/YieldingProcInstrumentation.cs
```diff
@@ -756,10 +756,9 @@ private IEnumerable<Declaration> PendingAsyncNoninterferenceCheckers()
         yield break;
       }
 
-      var pendingAsyncsToCheck = new HashSet<Action>(
-        civlTypeChecker.MoverActions
-          .Where(a => a.LayerRange.Contains(layerNum) && a.HasPendingAsyncs)
-          .SelectMany(a => a.PendingAsyncs).Select(a => civlTypeChecker.Action(a)));
+      var pendingAsyncsToCheck =
+        new HashSet<Action>(civlTypeChecker.MoverActions.Where(a =>
+          a.LayerRange.Contains(layerNum) && a.ActionDecl.MaybePendingAsync));
```
bkragl, line 761: Just to confirm my understanding: We previously made the mistake of only checking noninterference against PAs that were created by actions. However, in your test case below it is *procedure* `Foo` that creates a PA at layer 2. We don't have a dedicated declaration of PAs created by procedures. Instead, we take the actions declared with `async` (which sets `MaybePendingAsync`?) and check noninterference at all their layers, irrespective of whether there are actually PAs created or not.

shazqadeer: Correct.

PR #776, opened 2023-09-01
Structured representation for program elements

atomb:
Instead of concatenating and parsing strings everywhere, this introduces a structured data type for representing the transformations on statement labels that can happen during Boogie elaboration and VC generation. It relieves clients from needing to parse the full unsat core element labels.

### DIFF
--- a/Source/Core/ProgramElement.cs
+++ b/Source/Core/ProgramElement.cs
@@ -1,0 +1,103 @@
+using System;
+
+namespace Microsoft.Boogie
+{
+  public abstract record ProgramElement()
+  {
+    public abstract string SolverLabel { get; }
+
+    // This suffix indicates that an ID string represents the assumption of
+    // a specific ensures clause after a specific call.
+    protected const string ensuresSuffix = "ensures";
+
+    // This suffix indicates that an ID string represents the goal of
+    // proving a specific requires clause before a specific call.
+    protected const string requiresSuffix = "requires";
+
+    // This suffix indicates that an ID string represents the assumption
+    // of a specific requires clause after a specific call.
+    protected const string requiresAssumedSuffix = "requires_assumed";
+
+    // This suffix indicates that an ID string represents the goal of
+    // proving that a specific loop invariant is established.
+    protected const string establishedSuffix = "established";
+
+    // This suffix indicates that an ID string represents the goal of
+    // proving that a specific loop invariant is maintained.
+    protected const string maintainedSuffix = "maintained";
+
+    // This suffix indicates that an ID string represents the asssumption
+    // of a specific loop invariant in the body of the loop.
+    protected const string assumeInBodySuffix = "assume_in_body";
+
+    public static ProgramElement ParseSolverString(string idString)
+    {
+      var parts = idString.Split('$');
+      if (parts.Length == 3 && parts[2].Equals(requiresSuffix)) {
+        var reqId = parts[0];
+        var callId = parts[1];
+        return new CallRequiresGoalElement(callId, reqId);
+      }
+      else if (parts.Length == 3 && parts[2].Equals(requiresAssumedSuffix)) {
+        var reqId = parts[0];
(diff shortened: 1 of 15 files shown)


Source/Provers/SMTLib/SMTLibInteractiveTheoremProver.cs
```diff
@@ -223,9 +223,8 @@ public override void FullReset(VCExpressionGenerator generator)
               if (resp.Name != "")
               {
                 usedNamedAssumes.Add(resp.Name);
-                if (libOptions.TrackVerificationCoverage)
-                {
-                  reporter.AddCoveredElement(resp.Name.Substring("aux$$assume$$".Length));
+                if (libOptions.TrackVerificationCoverage) {
+                  reporter.AddCoveredElement(new LabeledElement(resp.Name.Substring("aux$$assume$$".Length)));
```
keyboardDrummer, line 227: Could you not have called `ParseSolverString` here already?

Source/Provers/SMTLib/SMTLibInteractiveTheoremProver.cs
```diff
@@ -234,7 +233,7 @@ public override void FullReset(VCExpressionGenerator generator)
                 usedNamedAssumes.Add(arg.Name);
                 if (libOptions.TrackVerificationCoverage)
                 {
-                  reporter.AddCoveredElement(arg.Name.Substring("aux$$assume$$".Length));
+                  reporter.AddCoveredElement(new LabeledElement(arg.Name.Substring("aux$$assume$$".Length)));
```
keyboardDrummer, line 236: Could you not have called `ParseSolverString` here already?

Source/Core/ProgramElement.cs
```diff
@@ -0,0 +1,103 @@
+using System;
+
+namespace Microsoft.Boogie
+{
+  public abstract record ProgramElement()
```
keyboardDrummer, line 5: Subjective: my preference for the name would be `TrackedNodeComponent` and for the subclasses to be `TrackedCallRequires`, `TrackedInvariantMaintained` etc. The `Node` part of the name than overlaps with `Call` and `Invariant` while `Component` overlaps with `RequiresGoal`, `RequiresAssumed` and `Ensures` for `Call`, and `Assumed`, `Established`, `Maintained` for `Invariant`.

keyboardDrummer:
Looks great, thanks !

Source/Provers/SMTLib/SMTLibInteractiveTheoremProver.cs
atomb: There's an additional bit of string processing going on here beyond what I introduced (and that's been in the Boogie codebase for years). I avoided touching it as part of this PR because I want to more deeply consider how to do so that is both elegant and doesn't break things. Fortunately, I think that it can be done, once I have a deeper understanding of that, without changing anything about the exposed API. This PR is critical because of how it changes the exposed API.

keyboardDrummer: Okay

keyboardDrummer (approved):

Source/Core/TrackedNodeComponent.cs
```diff
@@ -2,7 +2,10 @@
 
 namespace Microsoft.Boogie
 {
-  public abstract record ProgramElement()
+  // Represents an AST node, or component of a node, that is being
+  // tracked during the proof process to determine whether it was
+  // used as part of a completed proof.
+  public abstract record TrackedNodeComponent()
```
keyboardDrummer, line 8: I think `///` would be better here: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/xmldoc/

If you then use Rider to move `TrackedNodeComponent`, the comment will be moved with it, while the `//` is AST node wise not tied to `TrackedNodeComponent`.

PR #780, opened 2023-09-11
Consistent augmented execution trace after multiple verification runs

Dargones:
This PR makes Boogie produce a consistent enhanced error trace when verifying the same implementation several times, and it also adds a corresponding test case.

In more detail: To produce the enhanced error trace triggered by the `/enhancedErrorMessages` command line option, Boogie relies on the `debugInfos` map that is [filled in during passification](https://github.com/boogie-org/boogie/blob/24047c7b2d339e91cb39ebf8f8bf8d6ac5e6184e/Source/VCGeneration/ConditionGeneration.cs#L1055-L1095). This information is lost once the corresponding `ConditionGeneration` object is disposed, but an implementation is only ever passified once (see [here](https://github.com/boogie-org/boogie/blob/24047c7b2d339e91cb39ebf8f8bf8d6ac5e6184e/Source/VCGeneration/VCGen.cs#L383-L389)), so the augmented error trace is not produced on successive calls to `VerifyImplementation`. This PR attaches the `debugInfos` field to the `Implementation` object itself, meaning that this information is preserved across verification attempts.

By submitting this pull request, I confirm that my contribution is made under the terms of the [MIT license](https://github.com/boogie-org/boogie/blob/master/LICENSE.txt).

### DIFF
--- a/Source/Core/AST/Absy.cs
+++ b/Source/Core/AST/Absy.cs
@@ -3510,6 +3510,7 @@ namespace Microsoft.Boogie
     // Both are used only when /inline is set.
     public List<Block> OriginalBlocks;
     public List<Variable> OriginalLocVars;
+    public Dictionary<Cmd, List<object>> debugInfos = new();
 
     public readonly ISet<byte[]> AssertionChecksums = new HashSet<byte[]>(ChecksumComparer.Default);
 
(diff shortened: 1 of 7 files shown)


atomb (approved):
This looks nice. Remarkably straightforward.

atomb:
The tests should be fixed once #781 is merged.

keyboardDrummer (approved):

keyboardDrummer:
Nice

PR #796, opened 2023-10-13
Add `:reproducible_resource_limit` attribute

### LINKED ISSUE #795: :rlimit value multiplied by 1000 when passed to prover
This looks like a simple case of `Checker.SetRLimit` imitating the `Checker.SetTimeout` method. The difference is that the latter is converting between seconds and milliseconds, whereas there seems to be no reason to apply the same transformation on resource units, which are totally abstract. This gets really confusing if you then read the current rlimit value, as Dafny does, because that value is not similarly adjusted.

I don't think we should just "fix" this in place since it will break existing projects. I think we should instead document this behavior more explicitly, and add a parallel option with a different name that does not alter the value. I'd suggest something like "reproducible-resource-limit" to align with the concept in SMT-LIB: https://smtlib.cs.uiowa.edu/papers/smt-lib-reference-v2.6-r2017-07-18.pdf (p. 4.1.7)

atomb:
This doesn't multiply by 1000, unlike `:rlimit`. The corresponding CLI flag is `/reproducibleResourceLimit`.

I'm a little unsure about the name. It's a bit of a mouthful. Maybe leaving out "reproducible" would be better.

Fixes #795.

### DIFF
--- a/Source/ExecutionEngine/CommandLineOptions.cs
+++ b/Source/ExecutionEngine/CommandLineOptions.cs
@@ -1206,6 +1206,10 @@ namespace Microsoft.Boogie
           return true;
 
         case "rlimit":
+          ps.GetUnsignedNumericArgument(x => ResourceLimit = Util.BoundedMultiply(x, 1000), null);
+          return true;
+
+        case "reproducibleResourceLimit":
           ps.GetUnsignedNumericArgument(x => ResourceLimit = x, null);
           return true;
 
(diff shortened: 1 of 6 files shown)


Source/ExecutionEngine/CommandLineOptions.cs
```diff
@@ -1206,6 +1206,10 @@ protected override bool ParseOption(string name, CommandLineParseState ps)
           return true;
 
         case "rlimit":
+          ps.GetUnsignedNumericArgument(x => ResourceLimit = Util.BoundedMultiply(x, 1000), null);
+          return true;
+
+        case "reproducibleResourceLimit":
```
keyboardDrummer, line 1212: Can we use a word that captures "reproducible resource that correlates with time", like `tick` or `cycle` ?

Source/ExecutionEngine/CommandLineOptions.cs
```diff
@@ -1978,7 +1982,10 @@ similar effects.
                 Limit the number of seconds spent trying to verify
                 each procedure
   /rlimit:<num>
```
keyboardDrummer, line 1984: Please update the documentation to say that this option is deprecated and will be removed.

atomb:
I simplified things by removing the backward-compatibility layer. We'll want one in Dafny, but it's not necessarily needed in Boogie, given the very few users of the feature.

keyboardDrummer (approved):

keyboardDrummer (approved):

PR #800, opened 2023-10-16
Remove all references to `bv2int`

atomb:
The `bv2nat` function is more standard now, working with CVC5, too.

Two of these references are just the Boogie-level name of functions, but it’s nice to change those, too, so that there are no results when doing `git grep bv2int`.


keyboardDrummer (approved):

Source/Provers/SMTLib/SmtLibNameUtils.cs
```diff
@@ -41,7 +41,7 @@ public static string QuoteId(string s)
       "bvsmod0", "bvsdiv_i", "bvudiv_i", "bvsrem_i", "bvurem_i", "bvumod_i", "bvule", "bvsle", "bvuge",
       "bvsge", "bvslt", "bvugt", "bvsgt", "bvxor", "bvnand", "bvnor", "bvxnor", "sign_extend", "zero_extend",
       "repeat", "bvredor", "bvredand", "bvcomp", "bvumul_noovfl", "bvsmul_noovfl", "bvsmul_noudfl", "bvashr",
-      "rotate_left", "rotate_right", "ext_rotate_left", "ext_rotate_right", "int2bv", "bv2int", "mkbv",
+      "rotate_left", "rotate_right", "ext_rotate_left", "ext_rotate_right", "int2bv", "bv2nat", "mkbv",
```
keyboardDrummer, line 44: Why not have both in the list?

atomb: Yeah, I guess that wouldn't hurt.

PR #808, opened 2023-11-15
Add `smt_option` attribute

atomb:
This adds support for an attribute of the form `{:smt_option name, value}` to set SMT-Lib options using the `(set-option :name value)` command. It's available on a per-procedure level.


shazqadeer:
Thanks for your many contributions to Boogie, @atomb .

It would be nice to add documentation for this new attribute in CommandLineOptions.cs.

atomb:
> Thanks for your many contributions to Boogie, @atomb .
> 
> It would be nice to add documentation for this new attribute in CommandLineOptions.cs.

I just greatly simplified it and added attribute help text, if you want to take another look.

shazqadeer (approved):

PR #811, opened 2023-11-20
Update version, and remove race condition

keyboardDrummer:
`ClearLocalSMTOptions()` was called _after_ the checker was released, which leads to a race condition because a new thread may already be using the checker's options while they're still being cleared. Here's an example: https://github.com/dafny-lang/dafny/actions/runs/6930078076/job/18849083894?pr=4773

This PR reduces statefulness of the code, since it no longer clears and reparses `SMTLibProcessTheoremProver.options`. Instead it has a field `IEnumerable<OptionValue> additionalSmtOptions` that can be used to tweak SmtOptions between runs.

### DIFF
--- a/Source/Directory.Build.props
+++ b/Source/Directory.Build.props
@@ -2,7 +2,7 @@
 
   <!-- Target framework and package configuration -->
   <PropertyGroup>
-    <Version>3.0.6</Version>
+    <Version>3.0.7</Version>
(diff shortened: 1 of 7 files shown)


atomb (approved):
Thanks for doing this. I was considering separating the options into two sets even before you noticed this issue, and the concurrency problem is an even stronger reason to go that way.

PR #817, opened 2023-11-28
Fix crashes with verification coverage + counterexamples

atomb:
Previously, running Boogie with `/trackVerificationCoverage` and `/enhancedErrorMessages:1` (or anything that leads to counterexample parsing) would lead to crashes when verification failed. This fixes that issue.


shazqadeer (approved):

Source/Model/ParserZ3.cs
```diff
@@ -281,6 +281,9 @@ internal override void Run()
           }
           else
           {
+            if (funName.StartsWith("aux$$")) {
```
keyboardDrummer, line 284: Does the name `aux$$` occur anywhere else in the code? Can we use a string constant to refer to it?

keyboardDrummer (approved):

Source/Model/ParserZ3.cs
atomb: Yes, it occurs in a small number of other places. We probably should create a string constant for it. I'll do that in a future refactoring.

PR #821, opened 2023-12-05
Fix hanging when solver exits with a non-zero exit code

atomb:
Previously, if the solver quit abruptly with a non-zero exit code (such as for a segmentation fault), Boogie would hang. These changes turn that into a well-behaved solver exception.


Source/Provers/SMTLib/SMTLibProcess.cs
```diff
@@ -74,6 +75,10 @@ public SMTLibProcess(SMTLibOptions libOptions, SMTLibSolverOptions options)
 
     private void SolverExited(object sender, EventArgs e)
     {
+      if (options.Verbosity >= 2) {
+        Console.WriteLine($"[SMT-ERR-{{0}}] Solver exited with code {solver.ExitCode}.");
```
keyboardDrummer, line 79: I'd ask you to avoid calls to `Console.`, but I see this file already contains them ^,^

keyboardDrummer (approved):

keyboardDrummer:
Looks amazing. How did you find this?

Source/Provers/SMTLib/SMTLibProcess.cs
atomb: Yeah, we should refactor it to use something more parameterized.

atomb:
> Looks amazing. How did you find this?

It was a rather frustrating bout of debugging, I must say! I've known that this was an issue for a while, and I expected that the `SolverExited` code would be run when the solver had a segfault, so I kept putting either breakpoints or print statements there and never having them executed. So, IIRC, I think I read a bunch of the documentation for `Process` and came across the `EnableRaisingEvents` property. I actually created this branch a month or so ago and held off creating a PR while I figured out how to create a test that intentionally crashed without needing to compile some C code in the test suite. And then I realized it didn't need to segfault, just return non-zero, so `/usr/bin/false` does the trick.

> Also, I added a commit to increment the version.

Thanks!

PR #828, opened 2023-12-11
[Civl] Ticket example

shazqadeer:
- add test driver back
- remove nil in favor of Option
- use Lval and Lset

### DIFF
--- a/Source/Core/base.bpl
+++ b/Source/Core/base.bpl
@@ -217,8 +217,8 @@ function Set_Intersection<T>(a: Set T, b: Set T): Set T
   Set(MapAnd(a->val, b->val))
 }
 
-function Set_Choose<T>(a: Set T): T;
-axiom (forall<T> a: Set T :: {Set_Choose(a)} a == Set_Empty() || Set_Contains(a, Set_Choose(a)));
+function Choice<T>(a: [T]bool): T;
+axiom (forall<T> a: [T]bool :: {Choice(a)} a == MapConst(false) || a[Choice(a)]);
 
 /// finite maps
 datatype Map<T,U> {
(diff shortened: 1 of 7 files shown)


Source/Core/base.bpl
```diff
@@ -217,8 +217,8 @@ function Set_Intersection<T>(a: Set T, b: Set T): Set T
   Set(MapAnd(a->val, b->val))
 }
 
-function Set_Choose<T>(a: Set T): T;
-axiom (forall<T> a: Set T :: {Set_Choose(a)} a == Set_Empty() || Set_Contains(a, Set_Choose(a)));
+function Choice<T>(a: [T]bool): T;
+axiom (forall<T> a: [T]bool :: {Choice(a)} a == MapConst(false) || a[Choice(a)]);
```
bkragl, line 221: Why ditch the `Set_` prefix and working with the `Set` wrapper type?

Would it be of any use to make the return type `Option<T>` and return `None` for empty sets instead of an unconstrained `T`?

bkragl (commented):
Just one comment/question about the changed choice function.

Source/Core/base.bpl
shazqadeer: The earlier Set_Choose was defined on the type ```Set T``` of finite sets. In the ticket sample, I want to make a choice on the domain of an Lset which is ```[T]bool```. So I thought I would generalize Set_Choose to Choice allowing the idea to work on any set, finite or infinite.

> Would it be of any use to make the return type Option<T> and return None for empty sets instead of an unconstrained T?

Currently, I don't see any advantage.

bkragl: Oh, I did not realize that `Set` is for finite sets. Otherwise the domain of an `Lset` could have been a `Set`.

bkragl (approved):

shazqadeer:
```Set T``` is defined [here](https://github.com/boogie-org/boogie/blob/master/Source/Core/base.bpl#L143C1-L218C1). It was added just so we could consolidate reasoning about cardinality of finite sets in one place.

PR #841, opened 2024-01-26
Verification task per split

keyboardDrummer:
Dependent Dafny PR: https://github.com/dafny-lang/dafny/pull/5031

### Changes
- Change the API `executionEngine.GetImplementationTasks` to `executionEngine.GetVerificationTasks`, which returns a task for each statically defined assertion batch. When using the split on every assert option, this will return a verification task for each assertion.

### Testing
- Added test SplitOnEveryAssertion


Source/VCGeneration/Split.cs
```diff
@@ -883,24 +870,26 @@ public Counterexample ToCounterexample(ProverContext context)
         return result;
       }
 
-      public (Bpl.SolverOutcome outcome, VerificationRunResult result, int resourceCount) ReadOutcome(int iteration, Checker checker, VerifierCallback callback)
+      public VerificationRunResult ReadOutcome(int iteration, Checker checker, VerifierCallback callback)
```
atomb, line 873: Much nicer. :)

Source/UnitTests/ExecutionEngineTests/ExecutionEngineTest.cs
```diff
@@ -49,6 +49,36 @@ public async Task DisposeCleansUpThreads()
...
+}
+".Trim();
+    Parser.Parse(programString, "fakeFilename10", out var program);
+    var options = CommandLineOptions.FromArguments(TextWriter.Null);
+    options.VcsSplitOnEveryAssert = true;
+    options.PrintErrorModel = 1;
+    var engine = ExecutionEngine.CreateWithoutSharedCache(options);
+    var tasks = engine.GetVerificationTasks(program);
+    
+    // The first split is empty. Maybe it can be optimized away
```
atomb, line 70: Yeah, I've been wondering about the empty first split for a while now. It'd be nice to get rid of it.

Source/UnitTests/ExecutionEngineTests/ExecutionEngineTest.cs
```diff
@@ -329,26 +359,24 @@ public async Task LoopInvariantDescriptions()
...
-
-    Assert.True(tasks2[1].CacheStatus is Completed);
-    Assert.AreEqual(VcOutcome.Correct, ((Completed)tasks2[1].CacheStatus).Result.VcOutcome);
+    Assert.AreEqual(firstName, statusList[2].Item1);
+    Assert.IsTrue(statusList[2].Item2 is Completed);
+    Assert.AreEqual((secondName, new Running()), statusList[3]);
+    Assert.AreEqual(secondName, statusList[4].Item1);
+    Assert.IsTrue(statusList[4].Item2 is Completed);
+  
+    // Caching is currently not working
```
atomb, line 368: Given that at least this test of caching was working before, it'd be unfortunate to merge something that breaks it. Since I know that's something you're working on, do you have an upcoming PR already planned that'll fix it? It's perhaps worth making that separate for ease of reviewing, but I'd be reluctant to push a release of Boogie until it's at least working as well as it was before.

atomb (dismissed):
I like the changes this makes a lot. I left a few comments that could result in changes, though not necessarily as part of this PR. The one critical one is that I'd only want to merge this with caching broken if you know you'll have another PR immediately following that fixes it (before the next Boogie release).

Source/UnitTests/ExecutionEngineTests/ExecutionEngineTest.cs
keyboardDrummer: Caching is currently not used by Dafny, because we don't trust it. This pathway of Boogie, for which caching does not work, is only used by Dafny. To 'fix' the caching, I would not to write some new code that does caching at the split level instead of the implementation level. Probably nothing complex, but it does not make sense to me to develop something that we won't test end-to-test because we're not using it.

Source/VCGeneration/VerificationRunResult.cs
```diff
@@ -1,26 +1,17 @@
...
 
 namespace VC
 {
-  using Bpl = Microsoft.Boogie;
-  using System.Threading.Tasks;
-
   public record VerificationRunResult
   (
-    int VcNum,
+    int vcNum,
```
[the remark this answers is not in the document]
keyboardDrummer: Fixed

Source/UnitTests/ExecutionEngineTests/ExecutionEngineTest.cs
atomb: If you're confident that it's not used by any other tools, then this sounds fine.

atomb (approved):

PR #852, opened 2024-03-01
Add a FromSeed method to IVerificationTask

keyboardDrummer:
Required for this Dafny PR: https://github.com/dafny-lang/dafny/pull/5138

### DIFF
--- a/Source/Directory.Build.props
+++ b/Source/Directory.Build.props
@@ -2,7 +2,7 @@
 
   <!-- Target framework and package configuration -->
   <PropertyGroup>
-    <Version>3.1.12</Version>
+    <Version>3.1.0</Version>
     <TargetFramework>net6.0</TargetFramework>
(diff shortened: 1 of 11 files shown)


Source/Directory.Build.props
```diff
@@ -2,7 +2,7 @@
 
   <!-- Target framework and package configuration -->
   <PropertyGroup>
-    <Version>3.0.12</Version>
+    <Version>3.1.12</Version>
```
atomb, line 5: I imagine you're bumping to 3.1 because of the API change? If so, though, wouldn't 3.1.0 make more sense?

atomb (changes requested):
Thanks for doing this! My only question is about the version number.

Source/Directory.Build.props
keyboardDrummer: It would. Updated

atomb (approved):

PR #859, opened 2024-03-12
Fix crash on low rlimit

atomb:
Z3 has multiple ways of responding that the resource limit has been exceeded. In some cases it responds with an error message including "push canceled". This updates Boogie to handle more of the strange side cases.

### DIFF
--- a/Source/Provers/SMTLib/SMTLibBatchTheoremProver.cs
+++ b/Source/Provers/SMTLib/SMTLibBatchTheoremProver.cs
@@ -190,10 +190,14 @@ namespace Microsoft.Boogie.SMTLib
           errorModel = ParseErrorModel(modelSExp);
         }
 
-        if (options.LibOptions.ProduceUnsatCores && responseStack.Count > 0) {
-          var unsatCoreSExp = responseStack.Pop();
-          if (result == SolverOutcome.Valid) {
-            ReportCoveredElements(unsatCoreSExp);
+        if (options.LibOptions.ProduceUnsatCores) {
+          if (responseStack.Count > 0) {
+            var unsatCoreSExp = responseStack.Pop();
+            if (result == SolverOutcome.Valid) {
+              ReportCoveredElements(unsatCoreSExp);
+            }
+          } else {
+            currentErrorHandler.OnProverError("Solver did not return an unsat core.");
           }
         }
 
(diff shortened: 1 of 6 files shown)


Source/Provers/SMTLib/SMTLibBatchTheoremProver.cs
```diff
@@ -165,6 +165,10 @@ private async Task<SolverOutcome> CheckSat(CancellationToken cancellationToken)
 
         var outcomeSExp = responseStack.Pop();
         var result = ParseOutcome(outcomeSExp, out var wasUnknown);
+        // Sometimes Z3 responds with "(error ...)" followed by "unknown"
+        if (outcomeSExp.Name is "error" && responseStack.TryPeek(out var sExpr) && sExpr.Name is "unknown") {
+          responseStack.Pop();
+        }
```
robin-aws, line 171: Would it be better to do this before `ParseOutcome(...)` instead? Or perhaps we should also set `wasUnknown` to true ourselves so we hit `ParseReasonUnknown` below?

Source/Provers/SMTLib/SMTLibBatchTheoremProver.cs
```diff
@@ -181,10 +185,12 @@ private async Task<SolverOutcome> CheckSat(CancellationToken cancellationToken)
...
 
-        var modelSExp = responseStack.Pop();
-        errorModel = ParseErrorModel(modelSExp);
+        if (responseStack.Count > 0) {
+          var modelSExp = responseStack.Pop();
+          errorModel = ParseErrorModel(modelSExp);
+        }
 
-        if (options.LibOptions.ProduceUnsatCores) {
+        if (options.LibOptions.ProduceUnsatCores && responseStack.Count > 0) {
```
robin-aws, line 193: Shouldn't we communicate some kind of error explicitly if `ProduceUnsatCores` is on but `responseStack` is empty?

Source/Provers/SMTLib/SMTLibProcessTheoremProver.cs
```diff
@@ -1209,6 +1209,7 @@ protected SolverOutcome ParseOutcome(SExpr resp, out bool wasUnknown)
         case "error":
           if (resp.Arguments.Length == 1 && resp.Arguments[0].IsId &&
               (resp.Arguments[0].Name.Contains("max. resource limit exceeded")
+               || resp.Arguments[0].Name.Contains("push canceled")
```
robin-aws, line 1212: Is this because the only reason a push is ever cancelled is when Boogie itself determines resource limits are reached?

atomb: Yes, exactly.

robin-aws (approved):

PR #860, opened 2024-03-21
Report resolution errors to downstream system

keyboardDrummer:
### Changes
Report resolution errors to downstream system

### Testing
Added a test for each failure branch

### DIFF
--- a/Source/ExecutionEngine/ExecutionEngine.cs
+++ b/Source/ExecutionEngine/ExecutionEngine.cs
@@ -702,9 +702,30 @@ namespace Microsoft.Boogie
 
     }
 
-    public IReadOnlyList<IVerificationTask> GetVerificationTasks(Program program) {
-      program.Resolve(Options);
-      program.Typecheck(Options);
+    class CollectingErrorSink : IErrorSink
+    {
+      public readonly List<(IToken Token, string Message)> Errors = new();
+      
+      public void Error(IToken tok, string msg)
+      {
+        Errors.Add((tok, msg));
+      }
+    }
+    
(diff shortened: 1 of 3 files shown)


Source/ExecutionEngine/ExecutionEngine.cs
```diff
@@ -702,9 +702,30 @@ private Implementation[] GetPrioritizedImplementations(Program program)
...
+      var resolutionErrors = program.Resolve(Options, sink);
+      var errors = string.Join("\n", sink.Errors.Select(t => $"{t.Token}: {t.Message}"));
+      if (resolutionErrors > 0)
+      {
+        throw new Exception($"Boogie program had {resolutionErrors} resolution errors:\n{errors}");
+      }
+      var typeErrors = program.Typecheck(Options, sink);
+      if (typeErrors > 0)
+      {
+        throw new Exception($"Boogie program had {typeErrors} type errors:\n{errors}");
```
atomb, line 727: This seems like it'll always print nothing for `{errors}` because it's only reachable if there are no resolution errors and `errors` is initialized after resolution but before type checking.

atomb (changes requested):
I'd like to see an additional test for this, partly because I'd like to see an example of what the concrete output would look like in practice, and partly to ensure that the issue I mentioned below doesn't occur.

Source/ExecutionEngine/ExecutionEngine.cs
keyboardDrummer: Agreed, added tests.

atomb (approved):

PR #861, opened 2024-04-04
Parse model values of the form 5e1 (no decimal)

atomb:
Previously, models that contained real literals with no decimal point (e.g., 5e1) would lead to an exception in the model parser. This PR adapts Boogie to recognize such values.

Ensuring that Z3 will emit one of these values in a model is tricky, and a test for it would be fragile. But I've seen it appear in practice.


keyboardDrummer (commented):
Could you add a unit test for the method you've changed?

atomb:
> Could you add a unit test for the method you've changed?

Done!

keyboardDrummer (approved):

PR #865, opened 2024-04-09
Add missing string interpolation

### LINKED ISSUE #863: Possibly Missing Interpolation in SMTLibInteractiveTheoremProver
Hi,
I was reading the source code of the SMTLibInteractiveTheoremProver class, and I saw that line 812 has the following context:
```csharp
foreach (var relaxVar in relaxVars)
{
  var resp = await SendVcRequest("(get-value ({relaxVar}))").WaitAsync(cancellationToken);
...
```
I saw this as highly suspicious, prompting me to open this issue. Should there not be a $ before `"(get-value...`?

atomb:
The fact that we didn't detect this earlier makes it clear that this functionality wasn't being tested. Unfortunately, it's not immediately clear to me how to write a test that would exercise the relevant code.

This should fix #863, even though we don't have tests to confirm that.

### DIFF
--- a/Source/Houdini/README.md
+++ b/Source/Houdini/README.md
@@ -1,0 +1,8 @@
+Houdini is annotation assistance system that can infer missing contracts.
+
+The original approach was implemented for ESC/Java, and is described in
+this paper:
+
+https://link.springer.com/chapter/10.1007/3-540-45251-6_29
+
+This project adapts that approach to Boogie programs.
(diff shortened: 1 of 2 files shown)


keyboardDrummer (approved):
Any chance you could add a README to the Houdini package? Would be good to know what the potential users are.

keyboardDrummer (approved):

PR #875, opened 2024-04-25
Move classes into separate files

keyboardDrummer:
### Changes

This PR moves some classes from the VCExpr project, that were in large files, into separate files. Merging this makes it easier to review https://github.com/boogie-org/boogie/pull/709


atomb (approved):
Approved under the assumption that it indeed just moves classes. I didn't check that no other changes happened.

PR #878, opened 2024-04-26
Added sample for snaphot

NamrathaG:
Added a sample for the snapshot algorithm which returns an atomic snapshot using the double collect trick without using locks.

### DIFF
--- a/Test/civl/samples/snapshot.bpl
+++ b/Test/civl/samples/snapshot.bpl
@@ -1,3 +1,6 @@
+// RUN: %parallel-boogie "%s" > "%t"
+// RUN: %diff "%s.expect" "%t"
+
 type Value;
 
 type Tid;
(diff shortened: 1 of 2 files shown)


Test/civl/samples/snapshot.bpl
```diff
@@ -0,0 +1,116 @@
+type Value;
```
shazqadeer, line 1: If you look at other Boogie files in the samples folder, you will see two lines at the top
```
// RUN: %parallel-boogie "%s" > "%t"
// RUN: %diff "%s.expect" "%t"
```

These lines indicate that this file is a test. Please add these lines to this file.

A test x.bpl also has a "golden" file associated with it called x.bpl.expect. Please create this file also. 

Boogie tests are run using the lit tool which you have to install on your machine. See instructions on the Boogie github site.

An automatic way to create the golden file for a test is to run lit (after adding the two lines at the top which are lit directives). The output generated is dumped into the Output folder. You can grab it from there.

Test/civl/samples/snapshot.bpl
```diff
@@ -0,0 +1,116 @@
...
+    if (*) {
+        assume {:add_to_pool "K", mem[i]->ts, k} {:add_to_pool "V", mem[i]->value, v} true;
+        assume k < mem[i]->ts;
+        r1[tid->val][i] := StampedValue(k, v);
+    } else {
+        r1[tid->val][i] := mem[i];
+    }
+}
+
+yield procedure {:layer 0} read_s({:linear} tid: One Tid, i: int);
```
shazqadeer, line 100: There is pattern in this file that is common in Civl programs. There is a single yield procedure that refines a particular atomic action (although it is allowed for many yield procedures to refine the same action). In such situations, I tend to follow a consistent way of declaring each such pair of a yield procedure and an action---write them together in a consistent order, either always procedure followed by action or the other way round.

Since this pattern is so common, it may be helpful for reducing syntactic clutter to provide each such pair with a single syntactic declaration such as the following:
```
yield procedure Foo(...) returns (...)
// other yield procedure specs here
refines left/right/both action
{
   // action body here
}
{
  // yield procedure body here
}
```

The syntax above would allow both the yield procedure and the action to share a single interface declaration

shazqadeer (changes requested):
I have left comments regarding how to convert this sample into a test.

Test/civl/samples/snapshot.bpl
NamrathaG: yes, this syntax makes sense. Has it been implemented?

shazqadeer (commented):
The only manadatory feedback is to convert the sample into a test. Everything else is optional.

Test/civl/samples/snapshot.bpl
```diff
@@ -0,0 +1,116 @@
...
+
+    while (true)
+    invariant {:yields} {:layer 1} true;
+    {
+        i := 1;
+        while (i <= n)
+        invariant {:layer 1} (forall j:int :: 1 <= j && j < i ==> r1[tid->val][j]->ts <= mem[j]->ts);
+        invariant {:layer 1} (forall j:int :: 1 <= j && j < i ==> (r1[tid->val][j]->ts == mem[j]->ts ==> r1[tid->val][j]->value == mem[j]->value));
+        {
+            call read_f(tid, i);
```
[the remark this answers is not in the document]
NamrathaG: Will procedure `read` have no body, and will it refine an action that reads the value in memory and returns the value?

Test/civl/samples/snapshot.bpl
shazqadeer: No, this syntax does not exist right now.

Test/civl/samples/snapshot.bpl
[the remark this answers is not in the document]
shazqadeer: Yes

Test/civl/samples/snapshot.bpl
```diff
@@ -0,0 +1,119 @@
...
+    
+        if (r1 == r2) {
+            assert {:layer 2} (forall j:int :: 1 <= j && j <= n ==> r1[j] == mem[j]);
+            snapshot := r1;
+            break;
+        }
+    }
+}
+
+atomic action {:layer 2} write (v:Value, i: int)
```
shazqadeer, line 66: Change the layer annotation to ```{:layer 1,2}``` to indicate that ```write``` co-exists with ```read```, ```read_s```, and ```read_s```.

shazqadeer: Also, it is nicer to move both ```read``` and ```write``` towards the bottom of the file so that layers are ordered bottom to top.

shazqadeer (dismissed):

shazqadeer (approved):

PR #888, opened 2024-05-22
Remove code that causes memory leaks

keyboardDrummer:
### Changes
- Replace two usages of `ConcurrentBag` with `ConcurrentStack`, which has a much simpler implementation that has no risk of memory leaks. ConcurrentBag was causing leaks. More information is here: https://stackoverflow.com/questions/5353164/possible-memoryleak-in-concurrentbag
- Remove cancellation support from AsyncQueue, since the cancellation token would gain a reference to the item being dequeued, which a caller might not expect and which causes a memory leak if the cancellation token is kept alive.
- Add a 'Clear' method to 'AsyncQueue' to allow cancelling all customers
- Fix a surprising memory leak in `CustomStackSizePoolTaskScheduler`
- Enable using an ExecutionEngine with a custom TaskScheduler

### Testing
- Tested with a profiler that the above changes resolve the issue that each Boogie program created by Dafny was never garbage collected.

### DIFF
--- a/Source/Core/AST/Program.cs
+++ b/Source/Core/AST/Program.cs
@@ -401,7 +401,7 @@ public class Program : Absy
     }
   }
 
-  public readonly ConcurrentBag<TrackedNodeComponent> AllCoveredElements = new();
+  public readonly ConcurrentStack<TrackedNodeComponent> AllCoveredElements = new();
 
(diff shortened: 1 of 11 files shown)


atomb (approved):
This looks great. Thanks for investigating it!

PR #891, opened 2024-06-12
Add `pushScope`, `popScope`, `hide` and `reveal` commands

keyboardDrummer:
### Changes
- Add hide and reveal commands, that work together with pruning to enable preventing specific edges in the dependency graph from being traversed, thereby causing more to be pruned. An axiom can be marked as `hideable`. The hide and reveal commands can hide or reveal a set of functions, a state which is maintained throughout the execution of a Boogie implementation. During pruning, if a live but hidden function uses a hideable axiom, then the axiom will not be made live.
- Add the `pushScope` and `popScope` commands, which represents entering and leaving lexical scopes. Right now, these commands only influence the behavior of `hide` and `reveal`, but in the future they may also allow local variable shadowing, which would reduce the translation burden of downstream tools.

### Testing
- Add test `Reveal.bpl` which tests the new feature
- Add test `AssumeFalseSplit.bpl`, which tests an existing piece of the behavior of splits


Source/Core/AST/ChangeScope.cs
```diff
@@ -0,0 +1,34 @@
...
+namespace Microsoft.Boogie;
+
+/// <summary>
+/// Allows pushing and popping scopes inside a Boogie implementation.
+/// 
+/// Right now these scopes only affect the state of what functions are hidden and revealed using the hide and reveal keywords.
+/// However, in the future these scopes should also allow lexical scoping and variable shadowing.
+/// </summary>
+public class ChangeScope : Cmd {
+  public bool Push { get; }
```
atomb, line 13: An `enum` of `Push` and `Pop` would be a bit nicer than a `bool` here.

Source/Core/AST/HideRevealCmd.cs
```diff
@@ -0,0 +1,54 @@
...
+/// Can be used to hide or reveal a specific function, or all functions
+/// If pruning is turned on, a hidden function will be pruned despite being referenced in a Boogie implementation.
+/// The function is only partially pruned though: the function definition itself is kept, and only axioms
+/// that the function depends on, that are marked as hideable, are pruned.
+///
+/// Hide and revealing takes into account lexical scoping:
+/// A popScope command will undo any hide and reveal operations that came after the last pushScope command.  
+/// </summary>
+public class HideRevealCmd : Cmd {
+  public bool Hide { get; }
```
atomb, line 16: Similar to push/pop, an `enum` would be nicer than a `bool` here.

Source/Core/StandardVisitor.cs
```diff
@@ -81,6 +81,13 @@ public virtual Cmd VisitAssertCmd(AssertCmd node)
       VisitAttributes(node);
       return node;
     }
+    
+    public virtual Cmd VisitRevealCmd(HideRevealCmd node)
```
atomb, line 85: Do you not need a method for `ChangeScope` here, too?

Source/VCGeneration/ConditionGeneration.cs
```diff
@@ -1421,6 +1421,12 @@ private void AddDebugInfo(Cmd c, Dictionary<Variable, Expr> incarnationMap, List
       else if (c is CommentCmd)
       {
         // comments are just for debugging and don't affect verification
+      } else if (c is HideRevealCmd)
+      {
+        passiveCmds.Add(c);
```
atomb, line 1426: The fact that these stick around until the passive stage is unfortunate, in the sense that it's nice for that set of commands to be as small as possible, but I think it's also necessary for the goal of this PR.

Source/Core/BoogiePL.atg
```diff
@@ -564,11 +564,16 @@ VarOrType<out TypedIdent/*!*/ tyd, out QKeyValue kv>
 
 /*------------------------------------------------------------------------*/
 Axiom<out Axiom/*!*/ m>
-= (.Contract.Ensures(Contract.ValueAtReturn(out m) != null); Expr/*!*/ e; QKeyValue kv = null; .)
+= (.
+    Contract.Ensures(Contract.ValueAtReturn(out m) != null); 
+    Expr/*!*/ e; 
+    QKeyValue kv = null; 
+    bool canHide = false; .)
+  [ "hideable" (. canHide = true; .) ]
```
atomb, line 572: I might be more inclined to make this an attribute than a keyword, though I don't have a super strong opinion about that.

Source/Core/BoogiePL.atg
```diff
@@ -1110,8 +1115,24 @@ LabelOrCmd<out Cmd c, out IToken label>
...
+      | "hide"   (. isHide = true;  .)
+      )
+      ( ident    (. hideRevealId = new IdentifierExpr(t, t.val); .)
+      | "*"
+      )
+      (. c = hideRevealId == null ? new HideRevealCmd(t, isHide) : new HideRevealCmd(hideRevealId, isHide); .)
+      ";"
+    )
+  | LabelOrAssign<out c, out label>
+  | "popScope"            (. c = new ChangeScope(t, false); .)
```
atomb, line 1132: I don't think anything else in Boogie, the language, uses camel case like this. I'd maybe prefer just `push` and `pop`. The word `Scope` doesn't really tell you what kind of scope it is, so I'm not sure it adds much.

Source/VCGeneration/Prune/DataflowAnalysis.cs
```diff
@@ -0,0 +1,61 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace VCGeneration.Prune;
+
+/// <summary>
+/// Allows defining dataflow analysis
+/// </summary>
+abstract class DataflowAnalysis<TNode, TState> {
```
atomb, line 10: This is simple enough that it might not provide a benefit, but have you looked at whether this can be a special case of what's defined in the `AbstractInterpretation` project? Dataflow analysis is generally a special case of abstract interpretation.

atomb (approved):
This looks very nice. There are a few stylistic and naming things I'd change, along with possibly using `AbstractInterpretation`, but I'll leave it up to you whether to make any of those changes.

Source/Core/BoogiePL.atg
keyboardDrummer: I thought about that, especially since adding a parsing keyword is more difficult implementation wise, but I can't really defend introducing keywords for half a feature and not for the other half.

Source/VCGeneration/ConditionGeneration.cs
keyboardDrummer: Seems fine to me :-) They are used only in the passive stage.

Source/VCGeneration/Prune/DataflowAnalysis.cs
keyboardDrummer: I had a look at that package but it's a but hard to see how to extract what I need from there. Also, there isn't much code to deduplicate, since this `DataflowAnalysis` is tiny and all it really needs that could already be defined somewhere else, is a control flow graph with `Cmd` as a `Node`.

PR #892, opened 2024-06-19
Add disjointness assumption to IS proof obligations

shazqadeer:
This PR adds disjointness assumptions to IS proof obligations. Surprisingly, they had been missing all this time. These assumptions appear to be necessary for the scatter-gather version of distributed snapshot.

This PR also rewrites ```GeneratePartitionChecker``` to make better use of APIs in ```Action``` and ```ActionDecl```.


shazqadeer:
@NamrathaG : Please take note of the revised implementation of ```GeneratePartitionChecker```. You might find it helpful to read the code to learn about convenient APIs.

PR #893, opened 2024-06-20
ISR SideCondition and InconsistencyChecker 

NamrathaG:
SideCondition ensures that actions only put permissions in globals and not take any

InconsistencyChecker checks the D premise of the ISR rule


Test/civl/inductive-sequentialization/snapshot-scatter-gather-full.bpl
```diff
@@ -0,0 +1,234 @@
...
+type Value;
+
+type Tid;
+
+datatype StampedValue {
+  StampedValue(ts: int, value: Value)
+} 
+
+datatype Permission {
+    Permission(t_id: Tid, mem_index: int)
```
shazqadeer, line 13: Let us use consistent indentation throughout the file.

Test/civl/inductive-sequentialization/snapshot-scatter-gather-full.bpl
```diff
@@ -0,0 +1,234 @@
...
+const n: int;
+axiom n >= 1;
+
+var {:layer 0,2} mem: [int]StampedValue;
+var {:layer 0,2} r1: [Tid][int]StampedValue;  
+var {:layer 0,2} r2: [Tid][int]StampedValue; 
+var {:linear} {:layer 0,2} pSet: Set Permission;
+
+function {:inline} WholeTidPermission(tid: Tid): Set Permission {
+  Set((lambda {:pool "D"} x: Permission :: x->t_id == tid && (1 <= x->mem_index) && (x->mem_index <= n)))
```
shazqadeer, line 25: It might be helpful to have more meaningful names to the pools.

Test/civl/inductive-sequentialization/snapshot-scatter-gather-full.bpl
```diff
@@ -0,0 +1,234 @@
...
+action {:layer 1} start (tid: Tid)
+modifies pSet;
+creates main_f;
+{
+    var {:linear} sps: Set Permission;
+    call sps := Set_Get(pSet, (lambda x: Permission :: (x->t_id == tid) && (1 <= x->mem_index) && (x->mem_index <= n)));
+    call create_async(main_f(tid, sps));
+}
+
+// M =  main_f , \elim = read_f, I = Inv_f, M' = main_f'
```
shazqadeer, line 37: mysterious comment.

Test/civl/inductive-sequentialization/snapshot-scatter-gather-full.bpl
```diff
@@ -0,0 +1,234 @@
...
+async action {:layer 1} main_f(tid: Tid, {:linear_in} sps: Set Permission)
+refines {:IS_right} main_f' using Inv_f;
+modifies pSet;
+creates read_f;
+{
+    // var {:linear} sps: Set Permission;
```
shazqadeer, line 43: Remove the commented line?

Source/Concurrency/InductiveSequentialization.cs
```diff
@@ -604,6 +819,17 @@ protected override List<Declaration> GenerateCheckers()
...
+      if (rule == InductiveSequentializationRule.ISR){
+       decls.AddRange(GenerateSideConditionChecker(targetAction));
+       foreach (var elim in eliminatedActions)
+      {
+        if(elim == targetAction){
+          continue;
+        }
+        decls.AddRange(GenerateSideConditionChecker(elim));
+      }
+       decls.AddRange(DescendantCheckerFull());
```
shazqadeer, line 831: Use consistent naming for names of methods: GenerateSideConditionChecker and DescendantCheckerFull.

Source/Concurrency/InductiveSequentialization.cs
```diff
@@ -435,6 +435,221 @@ private List<Declaration> GenerateStepChecker(Action pendingAsync)
         invariantAction.ImplWithChoice.InParams, invariantAction.ImplWithChoice.OutParams, locals, cmds);
     }
 
+     protected List<Declaration> GenerateSideConditionChecker(Action act)
```
shazqadeer, line 438: Let us fix the indentation problems. It will help in reviewing the code.

Test/civl/inductive-sequentialization/snapshot-scatter-gather-full.bpl
```diff
@@ -0,0 +1,234 @@
+// RUN: %parallel-boogie "%s" > "%t"
```
shazqadeer, line 1: Should we drop ```-full``` from the name?

Source/Concurrency/InductiveSequentialization.cs
```diff
@@ -435,6 +440,293 @@ private List<Declaration> GenerateStepChecker(Action pendingAsync)
...
+          notElimExprs.Add(Expr.Eq(pendingAsyncExpr, emptyExpr));
+        }
+      }
+      var cmds = new List<Cmd>() {
+        CmdHelper.AssumeCmd(GetExitCondition(action)),
+        CmdHelper.CallCmd(action.Impl.Proc, action.Impl.InParams, action.Impl.OutParams),
+        GetCheck(action.tok, Expr.And(elimExprs), "Exit condition property 1 failed"),
+      };
+
+      //CivlUtil.ResolveAndTypecheck(civlTypeChecker.Options, cmds);
```
shazqadeer, line 686: I tried to uncomment this code but I got a crash. Needs to be investigated.

Source/Concurrency/InductiveSequentialization.cs
```diff
@@ -435,6 +440,293 @@ private List<Declaration> GenerateStepChecker(Action pendingAsync)
...
+      return GetCheckerTuple($"{rule}_ExitProperty1Checker_{action.Name}", new List<Requires>(),
+        action.Impl.InParams, action.Impl.OutParams, new List<Variable>(),  cmds);
+    }
+
+    protected List<Declaration> GenerateExitConditionProperty2Checker(Action action)
```
shazqadeer, line 691: We need better names instead of Property1 and Property2 both in the names of the methods and in the error messages.

Source/Concurrency/LinearityChecker.cs
```diff
@@ -2,6 +2,24 @@
...
+      public LinearityCheck(LinearDomain domain, Expr assume, Expr assert, string message, string checkName)
+      {
+        this.domain = domain;
+        this.assume = assume;
+        this.assert = assert;
+        this.message = message;
+        this.checkName = checkName;
+      }
+}
 class LinearityChecker
```
shazqadeer, line 23: Add a newline between 22 and 23.

Source/Concurrency/InductiveSequentialization.cs
```diff
@@ -435,6 +440,293 @@ private List<Declaration> GenerateStepChecker(Action pendingAsync)
...
+        inputs, outputs, requires, act.Impl.Proc.Modifies, new List<Ensures>());
+      Implementation linCheckImpl = DeclHelper.Implementation(linCheckerProc,
+        inputs, outputs,  new List<Variable>(), blocks);
+      return new List<Declaration>(new Declaration[] { linCheckerProc, linCheckImpl });
+    }
+
+    private List<Declaration> GenerateInconsistencyChecker(Action E1, Action E2)
+    {
+      var gateE1 = (Expr)Expr.True;
+      foreach(var g in E1.Gate)
```
shazqadeer, line 519: Replace loop with the version of Expr.And that takes in an Enumerable.

Same with the loops below.

Source/Concurrency/InductiveSequentialization.cs
```diff
@@ -435,6 +440,293 @@ private List<Declaration> GenerateStepChecker(Action pendingAsync)
...
+      {
+        localsForInM.Add(civlTypeChecker.LocalVariable($"templ_{l.Name}", l.TypedIdent.Type));
+      }
+
+      locals.Add(localE1);
+      locals.Add(localE2);
+      locals.AddRange(localsForGlobals);
+      locals.AddRange(localsForInM);
+
+      List<Expr> inputExprsM = new List<Expr>();
```
shazqadeer, line 554: use ```var``` to reduce duplication of type name

Source/Concurrency/InductiveSequentialization.cs
```diff
@@ -435,6 +440,293 @@ private List<Declaration> GenerateStepChecker(Action pendingAsync)
...
+        CivlUtil.ResolveAndTypecheck(civlTypeChecker.Options, pendingAsyncLinearParamsE1);
+        var collectInE1 = ltc.UnionExprForPermissions(domain, ltc.PermissionExprs(domain, pendingAsyncLinearParamsE1));
+
+        var pendingAsyncLinearParamsE2 = E2.ActionDecl.InParams
+          .Where(v => LinearTypeChecker.InKinds.Contains(LinearTypeChecker.FindLinearKind(v)))
+          .Select(v => ExprHelper.FieldAccess(Expr.Ident(localE2), v.Name)).ToList<Expr>();
+        CivlUtil.ResolveAndTypecheck(civlTypeChecker.Options, pendingAsyncLinearParamsE2);
+        var collectInE2 = ltc.UnionExprForPermissions(domain, ltc.PermissionExprs(domain, pendingAsyncLinearParamsE2));
+
+        var noDuplicationExpr1 = Expr.Eq(
```
shazqadeer, line 612: Some comments would be helpful about what is going on here.

shazqadeer:
Currently, the right mover checks are happening only for eliminated actions? Should they be done for the main action also?

Test/civl/inductive-sequentialization/snapshot-scatter-gather-full.bpl
NamrathaG: yes, it is not full yet. I plan to make it full later, but even then we don't need full in the name.

Source/Concurrency/InductiveSequentialization.cs
NamrathaG: The CallCmd is the problem. It looks like none of the other procedures that have this cmd resolve and type check it. I am not sure why though.

Source/Concurrency/InductiveSequentialization.cs
NamrathaG: Yeah, I was just postponing this until it really matters. I guess it does now. 
Property1 of exit_condition is that if the exit_condition holds, then the action does not create any pending asyncs to \elim. 
Property2 is that if the exit_condition does not hold, then the action does not create any pending asyncs to not(\elim). 

Let's name Property1 as NoPAtoElim and Property2 as NoPAtoNotElim?

NamrathaG:
> Currently, the right mover checks are happening only for eliminated actions? Should they be done for the main action also?

Yes, they should also be done for the main action. The current soundness proof depends on this fact.

Source/Concurrency/InductiveSequentialization.cs
shazqadeer: How about ```AllPendingAsyncsNotInElim``` and ```AllPendingAsyncsInElim``` to avoid the double negation in the second name?

Source/Concurrency/InductiveSequentialization.cs
```diff
@@ -435,6 +440,285 @@ private List<Declaration> GenerateStepChecker(Action pendingAsync)
...
+          notElimExprs.Add(Expr.Eq(pendingAsyncExpr, emptyExpr));
+        }
+      }
+      var cmds = new List<Cmd>() {
+        CmdHelper.AssumeCmd(Expr.Not(GetExitCondition(action))),
+        CmdHelper.CallCmd(action.Impl.Proc, action.Impl.InParams, action.Impl.OutParams),
+        GetCheck(action.tok, Expr.And(notElimExprs), "Exit condition property 2 failed"),
+      };
+      
+      //CivlUtil.ResolveAndTypecheck(civlTypeChecker.Options, cmds);
```
shazqadeer, line 707: We need to investigate why uncommenting causes a crash.

Source/Concurrency/MoverCheck.cs
```diff
@@ -121,7 +121,18 @@ where first.LayerRange.OverlapsWith(second.LayerRange)
         {
           foreach (var action in civlTypeChecker.MoverActions.Where(x => x.LayerRange.Contains(sequentialization.Layer)))
```
shazqadeer, line 122: Add the checks for the target action as well.

Source/Concurrency/MoverCheck.cs
NamrathaG, line 1: But looks like we are already doing the mover check for main action as well on line 120, no?

shazqadeer: oh yes, we are. i failed to notice earlier that the target action is being appended to the list of eliminated actions. nice!

Source/Concurrency/InductiveSequentialization.cs
shazqadeer: I put a fix in the ```ResolveAndTypecheck``` method. The type checking context ```tc``` has a field ```Proc``` which is the reference to the enclosing procedure. This is null and being accessed to do the modifies checking. I disabled the modifies checking trusting that ```GetCheckerTuple``` will set up the modified clause of the enclosing  checker procedure appropriately.

shazqadeer (approved):

PR #904, opened 2024-07-09
Relaxing side condition check

NamrathaG:
We don’t need the entire body of the action to obey the side condition (permissions are only put into globals). So, we relax it by only checking it when the exit condition does not hold.

### DIFF
--- a/Source/Concurrency/InductiveSequentialization.cs
+++ b/Source/Concurrency/InductiveSequentialization.cs
@@ -501,7 +501,7 @@ namespace Microsoft.Boogie
         var block = BlockHelper.Block($"{lc.domain.permissionType}_{lc.checkName}", cmds);
         checkerBlocks.Add(block);
       }
-      
+
       // Create init blocks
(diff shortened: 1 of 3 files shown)


shazqadeer (dismissed):
The change looks fine.

A few comments about GenerateSideConditionChecker unrelated to the change in this PR:

1. The conditional over checkerBlocks.Count == 0 can be eliminated by early exit from GenerateSideConditionChecker if ltc.LinearDomains.Count == 0 in which case an empty list of declarations can be returned. Then you would be guaranteed that linearityChecks is nonempty.

2. A better name than GenerateSideConditionChecker is needed. Please try to come up with a name that suggests what is being checked. This name should be reflected in the error message also.

3. " Only put permissions of type..." Hmm, isn't the check about not removing permissions from globals? So perhaps "Permissions of type ... may not be removed from globals".

shazqadeer:
The naming of InconsistencyChecker can also be improved. Let us try to find a better name for it.

PR #905, opened 2024-07-11
Only async call to M

NamrathaG:
Calls to procedures that are later refined using the ISR rule must be async calls. This is needed to apply the ISR rule.


Source/Core/AST/CallCmd.cs
```diff
@@ -395,6 +395,15 @@ void CheckModifies(IEnumerable<Variable> modifiedVars)
           }
         }
       }
+      if(!IsAsync){
+        for(int i = 1; i <= callerDecl.Layer; i++){
```
shazqadeer, line 399: Isn't it better to start ```i`` from calleeDecl.Layer?

Source/Core/AST/CallCmd.cs
```diff
@@ -395,6 +395,15 @@ void CheckModifies(IEnumerable<Variable> modifiedVars)
           }
         }
       }
+      if(!IsAsync){
```
shazqadeer, line 398: Use space between before ```(``` and after ```)``` consistently.

Source/Core/AST/CallCmd.cs
shazqadeer: Calling ```RefinedActionAtLayer``` might not work. I see that its implementation has strong preconditions. It might be better to implement a custom loop similar to that used in the implementation of ```RefinedActionAtLayer```.

Source/Core/AST/CallCmd.cs
```diff
@@ -395,6 +395,20 @@ void CheckModifies(IEnumerable<Variable> modifiedVars)
           }
         }
       }
+      var a = calleeDecl.RefinedAction;
+      if (!IsAsync) {
+          while(a != null){
```
shazqadeer, line 400: space around parentheses

Source/Core/AST/CallCmd.cs
```diff
@@ -395,6 +395,20 @@ void CheckModifies(IEnumerable<Variable> modifiedVars)
...
+            if (a.ActionDecl.RefinedAction != null) {
+              if (a.ActionDecl.RefinedAction.ActionDecl.LayerRange.UpperLayer > callerDecl.Layer){
```
shazqadeer, line 402: space after right parenthesis

Source/Core/AST/CallCmd.cs
```diff
@@ -395,6 +395,20 @@ void CheckModifies(IEnumerable<Variable> modifiedVars)
           }
         }
       }
+      var a = calleeDecl.RefinedAction;
```
shazqadeer, line 398: Move this line after the test of ```!IsAsync```

shazqadeer, line 398: rename ```a``` to something more meaningful

Source/Core/AST/CallCmd.cs
```diff
@@ -395,6 +395,20 @@ void CheckModifies(IEnumerable<Variable> modifiedVars)
...
+      if (!IsAsync) {
+          while(a != null){
+            if (a.ActionDecl.RefinedAction != null) {
```
shazqadeer, line 401: Factor out the expression ```a.ActionDecl.RefinedAction``` into a separate local variable and use it in subsequent code.

Source/Core/AST/CallCmd.cs
shazqadeer, line 402: Hmm, it seems to me that this test should be something like 
```
if (a.LayerRange.UpperLayer >= callerDecl.Layer)
```

NamrathaG:
Looks good to me. Let's merge!

PR #910, opened 2024-07-17
Use large stack for verification task execution

keyboardDrummer:
To prevent stack overflows like the following: https://github.com/aws/aws-database-encryption-sdk-dynamodb/actions/runs/9964887284/job/27534128432?pr=1202

I haven't added a test, partly because this code path is not accessible through the Boogie CLI. We could use an XUnit test to trigger it, but we'd also need the right Boogie program for that, possibly just one with a lot of nesting at the right place.

### DIFF
--- a/Source/ExecutionEngine/VerificationTask.cs
+++ b/Source/ExecutionEngine/VerificationTask.cs
@@ -126,8 +126,8 @@ public class VerificationTask : IVerificationTask {
       yield return new Running();
 
       var collector = new VerificationResultCollector(Split.Options);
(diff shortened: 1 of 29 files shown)


Source/ExecutionEngine/VerificationTask.cs
```diff
@@ -7,6 +7,7 @@
 using System.Reactive.Subjects;
 using System.Runtime.CompilerServices;
 using System.Threading;
+using System.Threading.Tasks;
```
fabiomadge, line 10: Do you use that?

keyboardDrummer: Not sure. I'm not too concerned about unused usings though.

PR #918, opened 2024-07-24
Print split

keyboardDrummer:
### Changes
- Extract `Block` and `Implementation` into separate files, both done in the first commit
- Add `/printSplit` and `/printSplitDeclarations` options
- Make `ComputeReachability` slightly more readable and performant
- Rename class `Prune` to `Pruner`, so it does not collide with the `Prune` namespace
- Persist slightly more source tokens when generating jump commands
- Fix a bug in `DataflowAnalysis` that would occur when using a bool as a state type
- Fix a bug in `Pruner.GetControlFlowGraph` that could cause the edge to the jump command to be missing
- Make pruning of split declarations lazy

### Testing
- Updated existing tests to use new `/printSplit` and `/printSplitDeclarations` options


Source/Core/AST/Program.cs
```diff
@@ -3,6 +3,7 @@
 using System.Collections.Generic;
 using System.Diagnostics.Contracts;
 using System.Linq;
+using System.Text;
```
atomb, line 6: Is this needed, given that there are no other changes to this file?

Source/VCGeneration/ManualSplit.cs
```diff
@@ -1,3 +1,4 @@
+using System;
```
atomb, line 1: Is this needed?

atomb (approved):
Looks like a nice improvement.

keyboardDrummer:
I'll process the review comments in the follow-up PR: https://github.com/boogie-org/boogie/pull/919

PR #919, opened 2024-07-24
Optimize blocks

keyboardDrummer:
### Changes

1. Perform block coalescing for each split
2. Add a post-split simplification that removes empty blocks with at most one outgoing or incoming edge

### Testing
- Updated the test `AssumeFalseSplit.bpl` to take into account simplification (2)


shazqadeer:
The block coalescing feature already exists in Boogie and appears to be on by default:

https://github.com/boogie-org/boogie/blob/1b8d9637600e5e0c6fd59e5a760f323c5380ed13/Source/ExecutionEngine/CommandLineOptions.cs#L1900

keyboardDrummer:
> The block coalescing feature already exists in Boogie and appears to be on by default:
> 
> https://github.com/boogie-org/boogie/blob/1b8d9637600e5e0c6fd59e5a760f323c5380ed13/Source/ExecutionEngine/CommandLineOptions.cs#L1900

Thanks. The coalescing I'm running happens for each split, after other optimizations which may enable more coalescing, so there is benefit in doing it again.

However, I've removed the duplicate coalescing implementation that this PR adds.

Source/VCGeneration/ManualSplitFinder.cs
```diff
@@ -1,18 +1,21 @@
+#nullable enable
 using System.Collections.Generic;
 using System.Diagnostics.Contracts;
 using System.Linq;
+using System.Threading.Tasks.Dataflow;
```
atomb, line 5: This seems like a very relevant library for this kind of task, but I don't actually see how it's used. Is it?

atomb (dismissed):
This looks good. None of my comments are blocking.

I have slight worries about soundness when doing transformations like this, though I don't see anything suspicious right away. For the moment, I think I trust the test suite (plus the Dafny test suite, once it's being used there) to detect potential bugs.

atomb (approved):

PR #926, opened 2024-07-31
Prevent creating return commands with empty tokens when copying blocks

keyboardDrummer:


Source/Core/AST/Block.cs
```diff
@@ -128,8 +128,8 @@ public bool IsLive(Variable v)
     return liveVarsBefore.Contains(v);
   }
 
-  public Block()
-    : this(Token.NoToken, "", new List<Cmd>(), new ReturnCmd(Token.NoToken))
+  public Block(IToken tok)
+    : this(tok, "", new List<Cmd>(), new ReturnCmd(tok))
```
atomb, line 132: Nice to have more location information!

atomb (approved):

PR #927, opened 2024-08-01
When possible, report line numbers of splits when using `/trace`

keyboardDrummer:
### Changes
- When possible, report line numbers of splits when using `/trace`

### Testing
- One existing test has been updated to check for the new line numbers


atomb (approved):
Seems good. I can't see how the timeout would be related to these changes, so perhaps it's another instance of the one #925 was intended to fix?

Issue #933, opened 2024-08-09
{:inline 0} is inconsistent 


petemud:
Under `/inline:assume`, which is default
```boogie
procedure {:inline 0} inline0() {}

procedure inconsistent() {
    call inline0();
    assert true == false;
}
```

becomes

```boogie
procedure inconsistent() {
    assume false;
    assert true == false;
}
```
```
Boogie program verifier finished with 1 verified, 0 errors
```
---
The bug got introduced in 08e368784c1ae629d870db6b09edadbef306e1d6

shazqadeer:
This is not a bug. Rather it is intended behavior. Boogie provides ways to customize what happens when the inline limit for a procedure is reached. See the documentation [here](https://github.com/boogie-org/boogie/blob/fa0c8712eb1893dd0170c11ba81e20ea8e6dfdb5/Source/ExecutionEngine/CommandLineOptions.cs#L1555).

petemud:
If proving `true==false` is an intended behavior, at least it shouldn't be the default

shazqadeer:
What default would you prefer?

petemud:
You choose. The consistent options if inline depth reaches 0 are:
1. put a call (inline `requires` and `ensures`)
2. only inline `requires` as `assert` or - *only if there are no `requires`* - drop a call completely
3. instead of inlining requires put `assert false`

Puttin `assume false` ever is plainly inconsistent. As for dropping call when there are `requires` - it is unsound

shazqadeer:
Closing this issue for now without fixing because user is not blocked.

PR #934, opened 2024-08-09
Remove assume false on inline 0

petemud:
#933

### DIFF
--- a/Source/Core/Inline.cs
+++ b/Source/Core/Inline.cs
@@ -400,7 +400,7 @@ namespace Microsoft.Boogie
               else if (options.ProcedureInlining == CoreOptions.Inlining.Assume)
               {
                 // add assume
-                newCmds.Add(new AssumeCmd(callCmd.tok, Expr.False));
+                newCmds.Add(new AssumeCmd(callCmd.tok, Expr.True));
               }
               else
               {


Source/Core/Inline.cs
```diff
@@ -400,7 +400,7 @@ void CheckRecursion(Implementation impl, Stack<Procedure /*!*/> /*!*/ callStack)
               else if (options.ProcedureInlining == CoreOptions.Inlining.Assume)
               {
                 // add assume
-                newCmds.Add(new AssumeCmd(callCmd.tok, Expr.False));
+                newCmds.Add(new AssumeCmd(callCmd.tok, Expr.True));
```
keyboardDrummer, line 403: Adding an `assume true` seems pointless. Better to remove the command altogether then.

keyboardDrummer (changes requested):
Can you add a test-case to this PR?

petemud:
> Can you add a test-case to this PR?

Maybe you can do it? I don't have development environment set up

Source/Core/Inline.cs
petemud: I think `assume false` on `inline == 0` also should be `assume true` or empty

shazqadeer:
I don't think we want to land this PR. See my comment on the associated issue.

petemud:
Maybe this pull request can actually change the options into: `Inlining.AssertFalse`, `Inlining.Requires`, `Inlining.Spec`? All three would be consistent. See [#933](https://github.com/boogie-org/boogie/issues/933#issuecomment-2278994958). What do you think, @shazqadeer?

shazqadeer:
@petemud : Let us back up a bit. Are you actually blocked on anything here? Specifically, is there a Boogie inline option that you can already use to get your job done?

petemud:
@shazqadeer I'm not blocked on `/inline` options. I can just not use `{:inline}`, or use `/inline:spec`. It'd just be nice for a verification backend to be sound

shazqadeer:
Glad to know you are not blocked.

In principle I agree with your comments about soundness. Note though that Boogie is less a fixed verification backend and more verification backend infrastructure. It is typically used by many different tools to achieve custom combinations of "soundness" and "scalability". Mostly, we want Boogie to be predictable and not do surprising things.

For the specific question of what do do with the inline attribute, I find that Boogie documentation is accurate and there are several options available to the user to achieve what they want. It is possible though:
-  the default option is not well chosen and should perhaps have been been InlineAssert or InlineSpec
- another option, in addition to the ones currently provided, is needed

I am happy to hear suggestions/PRs along these lines.

In the meantime, I will focus my attention on the other issue of irreducible graphs because it sounds like you are actually blocked on it.

PR #935, opened 2024-08-12
Extract classes out of DeadVarElim

keyboardDrummer:
Extract classes out of DeadVarElim


zetzschest (approved):

stefan-aws (approved):

shazqadeer:
I am a bit confused by the changes I see. I see only new code files being added. I was expecting to see edits to DeadVarElim.cs (or at least a total deletion) but I don't see any change like that.

keyboardDrummer:
> I am a bit confused by the changes I see. I see only new code files being added. I was expecting to see edits to DeadVarElim.cs (or at least a total deletion) but I don't see any change like that.

The diff is confusing. One file was "renamed" from DeadVarElim.cs to something else. In any case, there are a few extra lines because of the additional using statements at the top of files, but otherwise the diff does not add lines, and has extracted some classes into separate files.

shazqadeer:
Thanks for the clarification.

shazqadeer (approved):

PR #938, opened 2024-08-14
Record the SMT input size for each split

keyboardDrummer:
### Changes
Record the SMT input string size for each split, which can be used to detect regressions in encoding performance

### Testing
Updated existing test xml.bpl

### DIFF
--- a/Source/Provers/SMTLib/SMTLibInteractiveTheoremProver.cs
+++ b/Source/Provers/SMTLib/SMTLibInteractiveTheoremProver.cs
@@ -48,7 +48,6 @@ namespace Microsoft.Boogie.SMTLib
         SetupProcess();
         var c = common.ToString();
         SentSize += c.Length;
-        SentSmt.AppendLine(c);
         Process.Send(c);
       }
     }
@@ -128,7 +127,6 @@ namespace Microsoft.Boogie.SMTLib
 
         var result = await CheckSat(cancellationToken, errorLimit);
         SentSize = 0;
-        SentSmt.Clear();
         SendThisVC("(pop 1)");
 
         return result;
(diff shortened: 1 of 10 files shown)


Source/Provers/SMTLib/SMTLibInteractiveTheoremProver.cs
```diff
@@ -647,6 +653,7 @@ protected override void Send(string s, bool isCommon)
 
       if (Process != null)
       {
+        SentSize += s.Length;
```
atomb, line 656: It would be lovely if this could only be done from a wrapper around `Send`, to ensure that it's always incremented. It would be even nicer if the reset to size 0 could be encapsulated somewhere, too, though I remember from working on this code that there isn't a clear place to put it at the moment.

atomb (commented):
This seems like a useful thing to have. I'd prefer to have the size increment and `Send` calls paired in a separate method, though I don't feel strongly about that.

Test/commandline/xml.bpl
```diff
@@ -11,11 +11,11 @@
 // CHECK: \<method name="ExampleWithSplits" startTime=".*"\>
 // CHECK:   \<assertionBatch number="1" iteration="0" startTime=".*"\>
 // CHECK:     \<assertion file="xml.bpl" line="25" column="3" /\>
-// CHECK:     \<conclusion duration=".*" outcome="valid" resourceCount=".*" /\>
+// CHECK:     \<conclusion duration=".*" outcome="valid" resourceCount=".*" smtInputSize="601" /\>
```
atomb, line 14: It looks like these values need updating again.

atomb (changes requested):
This looks good overall, but it looks like there are some minor test fixes needed.

keyboardDrummer:
Currently sometimes `(declare-fun ControlFlow (Int Int) Int)` gets put into common and sometimes it is not, which can change the numbers. Should find a way so it's always included

shazqadeer:
@keyboardDrummer : Are you interested in pursuing this PR further? If not, I will close it.

PR #941, opened 2024-08-16
Change logic related to disposed checkerPools

keyboardDrummer:
Attempt to fix this: https://github.com/boogie-org/boogie/pull/937

### DIFF
--- a/Source/VCGeneration/CheckerPool.cs
+++ b/Source/VCGeneration/CheckerPool.cs
@@ -23,9 +23,7 @@ namespace VC
 
     public async Task<Checker> FindCheckerFor(Program program, Split? split, CancellationToken cancellationToken)
     {
-      if (disposed) {
-        throw new Exception("CheckerPool was already disposed");
-      }
+      cancellationToken.ThrowIfCancellationRequested();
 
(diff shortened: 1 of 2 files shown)


Source/VCGeneration/CheckerPool.cs
```diff
@@ -23,10 +23,6 @@ public CheckerPool(VCGenOptions options)
 
     public async Task<Checker> FindCheckerFor(Program program, Split? split, CancellationToken cancellationToken)
     {
-      if (disposed) {
-        throw new Exception("CheckerPool was already disposed");
-      }
-
       await checkersSemaphore.WaitAsync(cancellationToken);
```
MikaelMayer, line 26: It seems that this task will still not be cancelled if the checkerpool is disposed of. Can you indicate me how it could be otherwise?

keyboardDrummer: The call `await checkersSemaphore.WaitAsync(cancellationToken);` throws an `ObjectDisposedException` if the `CheckerPool` disposed.

MikaelMayer (approved):

PR #945, opened 2024-09-10
Replace AddAssignedVariables with AddAssignedIdentifiers

keyboardDrummer:
### Changes
- Replace AddAssignedVariables with AddAssignedIdentifiers, because the latter can already be computed before resolution, which is useful for Dafny.
- Enable a function to be declared as "revealed", so it stays revealed even if "hide *" is used.

### Testing
- This Dafny PR uses `AddAssignedIdentifiers`: https://github.com/dafny-lang/dafny/pull/5761
- Update `Reveal.bpl` test to test Always revealed

### DIFF
--- a/Source/Core/AST/AbsyCmd.cs
+++ b/Source/Core/AST/AbsyCmd.cs
@@ -1205,9 +1205,7 @@ namespace Microsoft.Boogie
       throw new NotImplementedException();
     }
 
-    public override void AddAssignedVariables(List<Variable> vars)
-    {
-      Contract.Requires(vars != null);
+    public override void AddAssignedIdentifiers(List<IdentifierExpr> vars) {
       throw new NotImplementedException();
     }
   }
@@ -1352,14 +1350,19 @@ namespace Microsoft.Boogie
     }
 
     public abstract void Emit(TokenTextWriter /*!*/ stream, int level);
-    public abstract void AddAssignedVariables(List<Variable> /*!*/ vars);
+
+    public IEnumerable<Variable> GetAssignedVariables() {
+      List<IdentifierExpr> ids = new();
+      AddAssignedIdentifiers(ids);
+      return ids.Select(id => id.Decl);
+    }
+    
+    public abstract void AddAssignedIdentifiers(List<IdentifierExpr> /*!*/ vars);
 
     public void CheckAssignments(TypecheckingContext tc)
     {
       Contract.Requires(tc != null);
-      List<Variable> /*!*/
-        vars = new List<Variable>();
-      this.AddAssignedVariables(vars);
+      var /*!*/ vars = GetAssignedVariables().ToList();
       foreach (Variable /*!*/ v in vars)
       {
         Contract.Assert(v != null);
@@ -1527,9 +1530,8 @@ namespace Microsoft.Boogie
     public override void Resolve(ResolutionContext rc)
     {
     }
-
-    public override void AddAssignedVariables(List<Variable> vars)
-    {
+    
+    public override void AddAssignedIdentifiers(List<IdentifierExpr> vars) {
     }
 
     public override void Typecheck(TypecheckingContext tc)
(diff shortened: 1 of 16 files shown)


Source/Core/AST/AbsyCmd.cs
```diff
@@ -1352,14 +1350,19 @@ public Cmd(IToken /*!*/ tok)
     }
 
     public abstract void Emit(TokenTextWriter /*!*/ stream, int level);
-    public abstract void AddAssignedVariables(List<Variable> /*!*/ vars);
+
+    public IEnumerable<Variable> GetAssignedVariables() {
+      List<IdentifierExpr> ids = new();
+      AddAssignedIdentifiers(ids);
+      return ids.Select(id => id.Decl);
```
MikaelMayer, line 1357: What you are saying is that `.Decl` is available only after resolution, whereas the id is available after parsing, correct? If that is the case, could you just add a comment to this function that says something like "Can be called only after resolution" ? Or I'm mistaken.

Source/Core/BoogiePL.atg
```diff
@@ -459,7 +459,10 @@ Function<.out List<Declaration>/*!*/ ds.>
      Expr definition = null;
      Expr/*!*/ tmp;
      Axiom ax;
+     bool revealed = false;
   .)
+  [ "revealed"                         (. revealed = true; .)
```
MikaelMayer, line 464: What is the use of "revealed"? Every time the function is in scope, its axiom is automatically revealed? Is it like equivalent to having automatic `reveal functionName` after `hide *` ?

I would have thought the other way: That whatever visibility functions would have for themselves by default could be overridden by statements in an implementation.

At least, this is the affordance we now provide in Dafny thanks to your hide statements.

Special rules like this remind me of the CSS `!important` and their 5 prioritization rules that they ended up implementing, with a weight depending on the number of qualifiers in each categories, IDs, classes, tagNames.

So are we envisioning other uses cases? I can think of
* Functions which are opaque by default, and can be revealed
* Functions which are revealed by default, and can be hidden
* Functions which are always opaque even with `reveal *`, unless we use their name explicitly
* Functions which are always revealed even with `hide *`, unless we use their name explicitly
* Function which can't be made opaque, always revealed
* Functions which can't be revealed, truly black boxes

My mind is not clear on all these use cases, so perhaps could you detail a bit more why this "revealed" is now necessary and why there is no other way?

Test/pruning/Reveal.bpl
```diff
@@ -57,4 +57,13 @@ procedure HideSpecific() {
...
   assert inner(3) == 42;
+}
+
+revealed function alwaysRevealed(x: int): int uses {
+  hideable axiom (forall x: int :: {alwaysRevealed(x)} alwaysRevealed(x) == 42);
+}
+
+procedure AlwaysRevealed() {
+  hide *;
+  assert alwaysRevealed(3) == 42;
```
MikaelMayer, line 68:
```suggestion
  hide alwaysRevealed;
  assert alwaysRevealed(3) == 42;
```
Would that work too?

MikaelMayer (commented):
I need a bit more context or insights to accept the current design.

Source/Core/AST/AbsyCmd.cs
keyboardDrummer: That's correct.

Source/Core/BoogiePL.atg
keyboardDrummer: > Is it like equivalent to having automatic reveal functionName after hide * ?

Yes

> perhaps could you detail a bit more why this "revealed" is now necessary and why there is no other way?

Currently Dafny does not allow its users to reference definitions from the prelude, so functions from there can not be revealed, which means they should also not be hidden using `hide *`. In the future, the prelude should be a Dafny library that users can see and reference, and then we could allow hiding prelude functions using `hide *`.

> So are we envisioning other uses cases?

Indeed in the future we may want more complicated systems for hiding and revealing, but I want to wait for user feedback before designing that.

Test/pruning/Reveal.bpl
keyboardDrummer: Yes, although in Dafny you wouldn't be able to write this, since you can't mark any Dafny functions as being always revealed. Only the prelude functions will get this property.

MikaelMayer (approved):
Looks good to me ! Thanks for the explanations. I understand better why there are functions we don't want to hide, which could be for example part of the prelude.

PR #951, opened 2024-09-23
Enable Boogie variables to be monotonic

keyboardDrummer:
### Changes
Add the following field to `Variable`:
```
    /// <summary>
    /// Prevents this variable from being havoc'd. Useful in situation where havoc commands are generated,
    /// such as after WhileCmds
    ///
    /// One use-case is the definite assignment tracking variables in Dafny, that only go from false to true,
    /// And can return an error if they're still false at the wrong point.
    /// </summary>
    public bool Monotonic { get; set;  }
```

### Testing
Added a CLI test

### DIFF
--- a/Source/Concurrency/CivlUtil.cs
+++ b/Source/Concurrency/CivlUtil.cs
@@ -301,7 +301,7 @@ namespace Microsoft.Boogie
 
     public static HavocCmd HavocCmd(params IdentifierExpr[] vars)
     {
-      return new HavocCmd(Token.NoToken, vars);
+      return new HavocCmd(Token.NoToken, vars.ToList());
     }
   }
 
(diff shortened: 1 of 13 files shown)


shazqadeer:
I didn't quite understand this feature. Could you provide more background on the Dafny feature motivating this change? Perhaps provide a concrete Dafny example.

keyboardDrummer:
> I didn't quite understand this feature. Could you provide more background on the Dafny feature motivating this change? Perhaps provide a concrete Dafny example.

Dafny has assignment tracking variables. Whenever a variable is assigned, it's associated tracking variable is set to true. Whenever a variable is used, we assert that it's tracking variable is true. When a loop modifies a variable, then it's associated tracking variable is set to true inside the loop, making Boogie consider this tracking variable one of the loop targets, and havoc'ing it after the loop. If the tracking variable was already set to true before the loop, this information is lost.

To counteract that information loss, Dafny will, before the loop, assign the values of tracking variables (all of them, since it does not know what the loop targets are) to temporary storage variables, and after the loop `assume tempStorage_i ==> trackingVariable_i`. This trick works but introduces unnecessary SMT, and requires unnecessary bookkeeping on the Dafny side.

keyboardDrummer:
Closing since I felt this was too much of a hack

PR #952, opened 2024-09-24
Introduce `{:isolate}` and `{:isolate "paths"}` attributes for assert and return commands

keyboardDrummer:
### Changes
- Introduce the `{:isolate}` attribute on AssertCmd and `ReturnCmd`
- Introduce the `{:isolate "paths"}` attribute on AssertCmd and `ReturnCmd`
- Introduce new tokens that allow precisely describing where a particular Split comes from.
- Fix a bug that caused too many VCs to be created when using `/vcsSplitOnEveryAssert` together with focus
- Fix incorrect token bug for VCs originating from focus.

### Testing
- Created a folder `Test/implementationDivision` that has tests for `{:isolate}`, `{:isolate "paths"}`, `{:focus}` and `{:split_here}`


Source/Graph/Graph.cs
```diff
@@ -623,45 +623,66 @@ public List<Node> SuccessorsAsList(Node n)
...
         if (dominators[node].Count > 1)
         {
           dominators[node].Remove(node);
         }
-        immediateDominator[node] = topoSorted.ElementAt(dominators[node].Max(e => topoSorted.IndexOf(e)));
+        immediateDominator[node] = topoSorted.ElementAt(dominators[node].Max(e => indexPerNode[e]));
       }
-      immediateDominator[source] = source;
+
+      immediateDominator.Remove(source);
```
MikaelMayer, line 685: I don't have the algorithmic context to understand these changes but I'll accept it as-it, hopefully it's well-tested in the CI.

Source/VCGeneration/ConditionGeneration.cs
```diff
@@ -881,8 +879,7 @@ private void AddDebugInfo(Cmd c, Dictionary<Variable, Expr> incarnationMap, List
           {
             foreach (var param in current.Params)
             {
-              if (param is IdentifierExpr identifierExpr)
-              {
+              if (param is IdentifierExpr identifierExpr) {
```
MikaelMayer, line 882: Style consistency?

Source/VCGeneration/Counterexample/Counterexample.cs
```diff
@@ -112,21 +114,19 @@ public void Print(int indent, TextWriter tw, Action<Block> blockAction = null)
...
         if (block.tok == null)
         {
           tw.WriteLine("{0}<intermediate block>", ind);
         }
-        else
-        {
+        else {
           // for ErrorTrace == 1 restrict the output;
           // do not print tokens with -17:-4 as their location because they have been
           // introduced in the translation and do not give any useful feedback to the user
```
MikaelMayer, line 128:
```suggestion
```

Source/VCGeneration/ManualSplit.cs
```diff
@@ -1,22 +1,22 @@
...
 using Microsoft.Boogie;
+using VCGeneration;
 
 namespace VC;
 
 public class ManualSplit : Split
 {
 
-  public IToken Token { get; }
+  public ImplementationPartOrigin Token { get; }
```
MikaelMayer, line 11: A Token that is not the type IToken? That might be misleading.

Source/VCGeneration/Splits/BlockRewriter.cs
```diff
@@ -0,0 +1,130 @@
...
+      {
+        var targets = blocksToInclude == null ? gtc.LabelTargets : gtc.LabelTargets.Where(blocksToInclude.Contains).ToList();
+        newBlock.TransferCmd = new GotoCmd(gtc.tok,
+          targets.Select(blk => oldToNewBlockMap[blk].Label).ToList(),
+          targets.Select(blk => oldToNewBlockMap[blk]).ToList());
+      }
+    }
+    newBlocks.Reverse();
+    
+    BlockTransformations.DeleteBlocksNotLeadingToAssertions(newBlocks);
```
MikaelMayer, line 127: That's an excellent one for future caching

Source/VCGeneration/Splits/IsolateAttributeOnAssertsHandler.cs
```diff
@@ -0,0 +1,118 @@
...
+    if (!results.Any()) {
+      return (results,partToDivide);
+    }
+    
+    return (results,GetSplitWithoutIsolatedAssertions());
+
+    ManualSplit GetSplitForIsolatedAssertion(Block blockWithAssert, AssertCmd assertCmd) {
+      var blocksToKeep = rewriter.Dag.ComputeReachability(blockWithAssert, false);
+
+      List<Cmd> GetCommands(Block oldBlock) =>
```
MikaelMayer, line 58: I did not know this syntax for local named closures. Nice.

Source/VCGeneration/Splits/IsolateAttributeOnAssertsHandler.cs
```diff
@@ -0,0 +1,118 @@
...
+    }
+  }
+
+  private static bool ShouldIsolate(bool splitOnEveryAssert, QKeyValue? isolateAttribute) {
+    if (splitOnEveryAssert) {
+      if (isolateAttribute == null) {
+        return true;
+      }
+
+      return !isolateAttribute.Params.OfType<string>().Any(p => Equals(p, "none"));
```
MikaelMayer, line 102: Is this "non" tested somewhere?

Source/VCGeneration/Splits/IsolateAttributeOnAssertsHandler.cs
```diff
@@ -0,0 +1,118 @@
...
+
+
+public class IsolatedAssertionOrigin : TokenWrapper, ImplementationPartOrigin {
+  public AssertCmd IsolatedAssert { get; }
+
+  public IsolatedAssertionOrigin(AssertCmd isolatedAssert) : base(isolatedAssert.tok) {
+    this.IsolatedAssert = isolatedAssert;
+  }
+
+  public string ShortName => $"/assert@{IsolatedAssert.Line}";
```
MikaelMayer, line 117: Do these names need to be unique?

Source/VCGeneration/Splits/IsolateAttributeOnJumpsHandler.cs
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    if (!results.Any()) {
+      return (results,partToDivide);
+    }
+
+    return (results, GetPartWithoutIsolatedReturns());
+    
+    ManualSplit GetPartWithoutIsolatedReturns() {
+      // TODO this needs an extra test. In case the isolated jump is followed by something it dominates
```
MikaelMayer, line 65: TODO to implement in this PR?

Source/VCGeneration/Splits/PathToken.cs
```diff
@@ -0,0 +1,27 @@
...
+namespace VCGeneration;
+
+public class PathOrigin : TokenWrapper, ImplementationPartOrigin {
+
+  public PathOrigin(IToken inner, List<Block> branches) : base(inner) {
+    Branches = branches;
+  }
+  
+  public List<Block> Branches { get; }
+  public string ShortName => $"/assert@{line}[{string.Join(",", Branches.Select(b => b.tok.line))}]";
```
MikaelMayer, line 18: Interesting short way to display paths.

Source/VCGeneration/Splits/Split.cs
```diff
@@ -76,32 +75,38 @@ public readonly VerificationConditionGenerator /*!*/
...
+        
+        var printToConsole = Options.PrintSplitFile == "-";
+        if (printToConsole) {
+          Thread.Sleep(100);
+        }
 
-        using var writer = new TokenTextWriter(
-          $"{Options.PrintSplitFile}-{Util.EscapeFilename(Implementation.Name)}-{SplitIndex}.spl", false,
-          Options.PrettyPrint, Options);
+        var prefix = (this is ManualSplit manualSplit) ? manualSplit.Token.ShortName : "";
```
MikaelMayer, line 101:
```suggestion
        var prefix = this is ManualSplit manualSplit ? manualSplit.Token.ShortName : "";
```

Source/VCGeneration/Splits/SplitAttributeHandler.cs
```diff
@@ -0,0 +1,216 @@
...
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.Contracts;
+using System.Linq;
+using Microsoft.Boogie;
+using VC;
+
+namespace VCGeneration;
+
+class SplitAttributeHandler {
```
MikaelMayer, line 11: Is this file only refactored from somewhere else?

Source/VCGeneration/VerificationConditionGenerator.cs
```diff
@@ -1365,23 +1351,22 @@ public static Counterexample AssertCmdToCounterexample(VCGenOptions options, Ass
...
     /// Returns a clone of "cex", but with the location stored in "cex" replaced by those from "assrt".
     /// </summary>
-    public static Counterexample AssertCmdToCloneCounterexample(VCGenOptions options, AssertCmd assert,
+    public static Counterexample AssertCmdToCloneCounterexample(VCGenOptions options, AssertCmd assrt,
       Counterexample cex,
-      Block implEntryBlock, Dictionary<TransferCmd, ReturnCmd> gotoCmdOrigins)
+      Block implEntryBlock)
     {
-      Contract.Requires(assert != null);
+      Contract.Requires(assrt != null);
```
MikaelMayer, line 1358: Strrrrrange renaming...

Source/VCGeneration/VerificationConditionGenerator.cs
```diff
@@ -1445,7 +1430,7 @@ public static Counterexample AssertCmdToCloneCounterexample(VCGenOptions options
           {
             Contract.Assert(block != null);
             Contract.Assume(block.TransferCmd != null);
-            returnCmd = gotoCmdOrigins.GetValueOrDefault(block.TransferCmd);
+            returnCmd = block.TransferCmd.tok is GotoFromReturn gotoFromReturn ? gotoFromReturn.Origin : null;
```
MikaelMayer, line 1433: I now see how you encode meta-data in tokens.
I had launched a discussion about the design of having different tokens vs. flattened tokens here:
https://github.com/dafny-lang/dafny/issues/3354
It all boils down to deciding if you want to encode multiple information using wrapped tokens, where each token provides more information. If that is the case, then a flat token containing everything is more maintainable.
If the cases are truly disjoint, then a hierarchy of tokens is the way to go.
In the light of that, are your new tokens wrapper nestable or not?

Source/VCGeneration/VerificationRunResult.cs
```diff
@@ -27,17 +27,9 @@ public void ComputePerAssertOutcomes(out Dictionary<AssertCmd, SolverOutcome> pe
...
-          } else if (counterExample is CallCounterexample callCounterexample) {
-            underlyingAssert = callCounterexample.FailingFailingAssert;
-          } else if (counterExample is ReturnCounterexample returnCounterexample) {
-            underlyingAssert = returnCounterexample.FailingFailingAssert;
-          } else {
-            continue;
-          }
+        foreach (var counterExample in CounterExamples)
+        {
+          var underlyingAssert = counterExample.FailingAssert;
```
MikaelMayer, line 32: Nice simplification

Test/civl/regression-tests/non-interference-2.bpl.expect
```diff
@@ -2,10 +2,8 @@ non-interference-2.bpl(25,1): Error: Non-interference check failed
 Execution trace:
     non-interference-2.bpl(7,3): anon0
     non-interference-2.bpl(12,5): inline$AtomicIncr$0$anon0
-    (0,0): inline$Civl_NoninterferenceChecker_yield_PC$0$L
```
MikaelMayer, line 5: I get that (0, 0) is not useful, but what about the label on the right? It looked useful.

MikaelMayer (commented):
Here are my initial comments. It's nice to see isolate assertions, it's going to be useful.

Source/ExecutionEngine/VerificationTask.cs
```diff
@@ -44,7 +44,7 @@ public VerificationTask(ExecutionEngine engine, ProcessedProgram processedProgra
 
   public IVerificationTask FromSeed(int newSeed)
   {
-    var split = new ManualSplit(Split.Options, () => Split.Blocks, Split.GotoCmdOrigins, 
+    var split = new ManualSplit(Split.Options, () => Split.Blocks,
```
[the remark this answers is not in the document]
keyboardDrummer: At some point a final block is created that contains assertions for the ensures clauses, and all returns statement are changed into gotos that jump to this block. However, to enable detecting the return statement when failing to prove an ensures clause, the codebase must remember that this translation occurred. Previously this was done using a dictionary that was passed all over the codebase, but now it's doing using a special token that's attached to the goto command, which means a lot less code bureaucracy and also allows further transformations without breaking the error reporting.

Source/Graph/Graph.cs
keyboardDrummer: The behavior of the method is slightly changed: it no longer inserts the source node into the mapping, which corresponds better to the definition of immediate denominator which says "Every node, except the entry node, has an immediate dominator."

Source/VCGeneration/ConditionGeneration.cs
keyboardDrummer: What's your question?

Source/VCGeneration/Splits/IsolateAttributeOnAssertsHandler.cs
keyboardDrummer: It it is not and it's not even used. Was part of an earlier version.

Source/VCGeneration/Splits/IsolateAttributeOnJumpsHandler.cs
keyboardDrummer: I've removed the TODO. Right now isolated jumps are only supported for Boogie "return" commands, which covers the Dafny use-case of isolating Dafny return statements. Maybe when we support isolate on Dafny continue statements I will have to look at this Boogie code again.

Source/VCGeneration/VerificationConditionGenerator.cs
keyboardDrummer: IMO the name `Token` should be reserved for lexical tokens, so data that only describes a source location and a word. However, I think for error reporting purposes, it is useful to track what the origin of nodes in the AST is, that origin might be `ParsedFromSource(IToken token) : IOrigin`, but it can also be something related to a program transformation. In this case, the origin of the goto jump is the desugaring transformation of a return jump.

Since I don't want to refactor tokens now, I'm abusing the IToken class as if it was an `IOrigin`.

Test/civl/regression-tests/non-interference-2.bpl.expect
keyboardDrummer: Yeah.. that name is still in the codebase, but someone needs to assign a proper token to some of the generated nodes before it will show up in the execution trace. I can't really judge what the civl code is doing and what tokens should be where.

Source/VCGeneration/Splits/SplitAttributeHandler.cs
keyboardDrummer: It's part of the old ManualSplitFinder.cs, but it's changed quite a bit.

Source/VCGeneration/VerificationConditionGenerator.cs
keyboardDrummer: A change from the refactoring PR you reviewed before, that through a git error of mine did not end up in this PR.

Source/Core/VariableDependenceAnalyser.cs
```diff
@@ -523,34 +523,32 @@ private Dictionary<Block, HashSet<Block>> ComputeGlobalControlDependences()
...
+            if (DirectCallee != null)
             {
-              continue;
-            }
-
-            var indirectCallees = ComputeIndirectCallees(callGraph, directCallee);
-            foreach (var control in GetControllingBlocks(b, localCtrlDeps[impl]))
-            {
-              foreach (var c in indirectCallees.Select(Item => Item.Blocks).SelectMany(Item => Item))
+              HashSet<Implementation> IndirectCallees = ComputeIndirectCallees(callGraph, DirectCallee);
```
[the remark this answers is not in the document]
keyboardDrummer: Refactoring from a previous PR that through a git error was reverted in this PR.

Source/Core/AST/CallCmd.cs
```diff
@@ -882,7 +882,7 @@ protected override Cmd ComputeDesugaring(PrintOptions options)
 
           // Do this after copying the attributes so it doesn't get overwritten
           if (callId is not null) {
-            a.CopyIdWithModificationsFrom(tok, req,
+            (a as ICarriesAttributes).CopyIdWithModificationsFrom(tok, req,
```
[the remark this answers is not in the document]
keyboardDrummer: Refactoring from a previous PR that through a git error was reverted in this PR.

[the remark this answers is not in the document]
MikaelMayer: That issue happened to me in the past as well 😖

Source/VCGeneration/ConditionGeneration.cs
MikaelMayer: I see a curly brace put at the end of the line, but it seems that for loops in the same file have it on a second line.

Source/VCGeneration/VerificationConditionGenerator.cs
MikaelMayer: Agreed that `IToken` can be renamed to `IOrigin`, so that the token part is a sub-part of it, concerns only the lexical token, and not every metadata we don't know otherwise where to store.

In the meantime, if there is a token wrapper, I suggest there is at most one such wrapper that will contain any possible information about the origin. Is it what you did? Or are there at least two token wrapper classes?

Test/civl/regression-tests/non-interference-2.bpl.expect
MikaelMayer: Where is the message stored? If it's not in the token, we should display it without token information. If it was in the token, and is lost because of the new encoding, I'd recommend to revert that change.
I worry about breaking Civi users who relied on the last line being printed.

Source/VCGeneration/VerificationConditionGenerator.cs
keyboardDrummer: > Is it what you did? Or are there at least two token wrapper classes?

I think so, if I understand you correctly. Every AST node only has a single origin related field, whether that is a Token or a TokenWrapper.

Test/civl/regression-tests/non-interference-2.bpl.expect
keyboardDrummer: Okay, I've made it so that nothing changes in the civl tests.

MikaelMayer (approved):

keyboardDrummer:
> @keyboardDrummer : Thank you for your continued contributions to Boogie. I am adding a late comment to this PR.
> 
> I notice that this PR adds a few more attributes related to VC splitting. Perhaps, the Dafny team has added other related attributes in the past also. Could you please make sure that all added attributes are documented in CommandLineOptions.cs? Thanks.

I've added some small documentation in https://github.com/boogie-org/boogie/pull/970

PR #953, opened 2024-09-25
Add property for generating standalone binary

rahxephon89:
This PR adds three items to `<PropertyGroup>` so that using `dotnet publish Source/Boogie.sln -r <RID>` can generate standalone binary of boogie.

### DIFF
--- a/Source/BoogieDriver/BoogieDriver.csproj
+++ b/Source/BoogieDriver/BoogieDriver.csproj
@@ -10,6 +10,9 @@
     <AssemblyName>BoogieDriver</AssemblyName>
     <PackAsTool>true</PackAsTool>
     <ToolCommandName>boogie</ToolCommandName>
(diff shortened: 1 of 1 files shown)


shazqadeer:
https://github.com/dotnet/sdk/issues/10053

Looks like there is a conflict between PackAsTool and self-contained app.

rahxephon89:
> [dotnet/sdk#10053](https://github.com/dotnet/sdk/issues/10053)
> 
> Looks like there is a conflict between PackAsTool and self-contained app.

Interesting... I tried `dotnet publish  Source/Boogie.sln  -r osx-x64 -c release` and it works. Perhaps because I am using `publish` instead of `build`?

keyboardDrummer:
For publishing Boogie online I'm guessing we don't want the options that this PR configured. How about you add a `.sh` file that invoces `dotnet publish` with the arguments you require? That way rest of Boogie is unaffected.

PR #961, opened 2024-10-07
Extracted refactorings from https://github.com/boogie-org/boogie/pull/960

keyboardDrummer:
Extracted classes into separate files, corrected variable names, and moved some methods to extension classes


atomb (approved):
This looks like a nice step toward making the code base easier to understand and work with.

PR #964, opened 2024-10-11
No cache

keyboardDrummer:
### Changes
- Enable not using a verification result cache, using the new class `EmptyVerificationResultCache`

### Testing
- Manually tried out using EmptyVerificationResultCache in Dafny, which reduces memory pressure


Source/ExecutionEngine/Caching/VerificationResultCache.cs
```diff
@@ -0,0 +1,86 @@
...
+
+namespace Microsoft.Boogie
+{
+  public sealed class VerificationResultCache : IVerificationResultCache
+  {
+    private readonly MemoryCache programCache = new("ProgramCache");
+    private readonly MemoryCache Cache = new("VerificationResultCache");
+
+    public Program? CachedProgram(string programId) {
+      return programCache.Get(programId) as Program;
```
MikaelMayer, line 16: +1 for removing the "result" intermediate variable

Source/ExecutionEngine/Caching/VerificationResultCache.cs
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    private readonly CacheItemPolicy Policy = new CacheItemPolicy
+      {SlidingExpiration = new TimeSpan(0, 10, 0), Priority = CacheItemPriority.Default};
+
+    public void Insert(Implementation impl, ImplementationRunResult result)
+    {
+      Contract.Requires(result != null);
```
MikaelMayer, line 24:
```suggestion
      Contract.Requires(impl != null);
      Contract.Requires(result != null);
```
This contract was left out during refactoring. Or did you intend to remove both contracts?
The function Lookup, RemoveMatchingKeys and VerificationPriority below also had a contract removed.

MikaelMayer (commented):
I would like a justification on the removal of contracts, since I don't understand this part.

Source/Directory.Build.props
```diff
@@ -2,7 +2,7 @@
 
   <!-- Target framework and package configuration -->
   <PropertyGroup>
-    <Version>3.3.1</Version>
+    <Version>3.3.2</Version>
```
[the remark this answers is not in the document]
keyboardDrummer: It's manual

Source/ExecutionEngine/Caching/VerificationResultCache.cs
keyboardDrummer: The presence of the contract creates warning, because it implies `impl` can be `null` and this file uses `#nullable enable`, so I had to remove the check, otherwise the build fails.

MikaelMayer (approved):

PR #969, opened 2024-10-22
Isolate path refactorings

keyboardDrummer:
### Changes
- Fix a concurrency issue in commit: https://github.com/boogie-org/boogie/pull/969/commits/c19096e853ae1af1a49303df5f7c0c6936cd1225
- Refactorings that were part of https://github.com/boogie-org/boogie/pull/970

### Testing
No additional tests needed


Source/Concurrency/YieldingProcInstrumentation.cs
```diff
@@ -545,7 +541,9 @@ private void DesugarConcurrency(Implementation impl, List<Cmd> preconditions)
       impl.Blocks.Add(unchangedCheckerBlock);
       impl.Blocks.Add(returnCheckerBlock);
       impl.Blocks.Add(returnBlock);
-      impl.Blocks.AddRange(implRefinementCheckingBlocks);
+      foreach (var block in implRefinementCheckingBlocks) {
+        impl.Blocks.Add(block);
+      }
```
robin-aws, line 546: Why is this preferred?

Source/Core/AST/GotoBoogie/Block.cs
```diff
@@ -7,37 +7,11 @@ namespace Microsoft.Boogie;
...
-    }
-    set
-    {
-      Contract.Requires(value != null);
-      this.cmds = value;
-    }
-  }
+  [Rep] 
+  [ElementsPeer] 
+  public List<Cmd> /*!*/ Cmds;
```
robin-aws, line 14: Better because simpler? Technically we're losing the contract.

Source/Core/AST/StructuredBoogie/BigBlocksResolutionContext.cs
robin-aws, line 1: Assuming this is only renaming and formatting?

Source/Concurrency/YieldingProcInstrumentation.cs
keyboardDrummer: `AddRange` does not exist on `IList`

Source/Core/AST/StructuredBoogie/BigBlocksResolutionContext.cs
keyboardDrummer: There's also conversion to a switch statement, and possibly other fully automated Rider refactorings such as invert if

Source/Core/AST/GotoBoogie/Block.cs
keyboardDrummer: TBH I don't value the nullability contracts. It would be simple to add them by adding `#nullable enable`

Source/Concurrency/YieldingProcInstrumentation.cs
robin-aws: Ah damn, feels like a ripe opportunity for an extension method :P

robin-aws (approved):

PR #970, opened 2024-10-22
Enable marking if commands as {:allow_split} or not

keyboardDrummer:
### Changes
- Introduce the attribute `{:allow_split}` that can be placed on `IfCmd`. When using `{:isolate "paths"}` only, branches with `{:allow_split}` will trigger splitting of the VC.

### Testing
- Updated existing tests so they exercise the presence and absence of `{:allow_split}`
- Added a test that combines `{:vcs_split_on_every_assert}` and explicit returns.


Source/VCGeneration/Splits/BlockRewriter.cs
```diff
@@ -35,67 +38,112 @@ public Cmd TransformAssertCmd(Cmd cmd) {
...
-
-        newToOldBlocks[newPrevious] = oldPrevious;
-        if (newPrevious.TransferCmd is GotoCmd gotoCmd) {
-          newPrevious.TransferCmd =
-            new GotoCmd(gotoCmd.tok, new List<string> { firstBlock.Label }, new List<Block> {
-              firstBlock
-            });
-        }
+  /// <summary>
+  /// Each focus block creates two options.
```
robin-aws, line 43: nit: focus is probably the wrong word here. Or else there should be a more explicit statement that isolating paths is implemented in terms of `{:focus}`

Is it correct to think of `{:isolate "paths"}` as equivalent to adding `{:focus}` to each branch?

Also how does path-based splitting work on while loops?

Test/implementationDivision/focus/focus.bpl
```diff
@@ -1,4 +1,4 @@
-// RUN: %parallel-boogie /printSplit:%t /errorTrace:0 "%s" > "%t"
+// RUN: %boogie /printSplit:- /errorTrace:0 "%s" > "%t"
```
robin-aws, line 1: Whoops! :) Out of curiousity was the old behavior non-deterministic and the test flaky?

Test/implementationDivision/focus/focus.bpl.expect
```diff
@@ -1,3 +1,118 @@
...
+    assume x#AT#0 + y#AT#0 == 5;
+    assume x#AT#0 * x#AT#0 <= 25;
+    assume {:partition} 0 >= x#AT#0;
+    assume y#AT#2 == y#AT#0;
+    assert y#AT#2 >= 0;
+    return;
+}
+
+
+implementation Ex/split@15() returns (y: int)
```
robin-aws, line 16: It feels confusing that you end up with three different patterns for naming splits here: 

1. The first unqualified `Ex` split above (and what does that cover then?)
2. This "number 15" split
3. The path-based splits below

Being clear on this matters since the user has to understand their source in order to fix verification failures.

I would at least suggest being slightly clearer on the path-based labels, perhaps something like `Ex[path: 14 -> 19 -> 25]`. You could produce that in Dafny error messages or at the Boogie level.

robin-aws (changes requested):
Implementation LGTM as far as I can tell (testing coverage is good and didn't see anything obviously wrong/risky). Mainly looking for clearer labeling and documentation of the feature (and will overlap with my review of https://github.com/dafny-lang/dafny/pull/5832)

Source/VCGeneration/Splits/BlockRewriter.cs
keyboardDrummer: > focus is probably the wrong word here

Comment was accidentally copied

> Is it correct to think of {:isolate "paths"} as equivalent to adding {:focus} to each branch?

I have not managed to think of one entirely in terms of the other. I wanted to see if I could reuse the focus code for either `{:isolate "paths}` on asserts or on returns, but didn't manage.

> Also how does path-based splitting work on while loops?

A loop creates a branch that does not converge, so path-based splitting is not useful to process it.

```
LoopHead:
 assert <invariants>
 goto LoopDone, LoopBody

LoopBody:
  havoc loopTargets;
  assume <invariants>;
  assume <guard>;
  <body>
  assert <invariants>; // adding {:isolate "paths"} here does not help, since only a single path leads here
  return;

LoopDone:
  assume !<guard>
  <rest>
  assert <foo> // adding {:isolate "paths"} here does not help, since only a single path leads here
```

Test/implementationDivision/focus/focus.bpl
keyboardDrummer: Yes, at least in this PR it became flaky otherwise.

Test/implementationDivision/focus/focus.bpl.expect
keyboardDrummer: > The first unqualified Ex split above (and what does that cover then?)

Impossible to say what it covers. It is the "remainder" after processing all the focus/isolate/split_here

> This "number 15" split

It's actually the line number from which the assertion originates. I could add the column to possibly disambiguate

> Being clear on this matters since the user has to understand their source in order to fix verification failures.

This notation is only used at the Boogie level, where I optimize for power users. I use more verbose notation in Dafny.

keyboardDrummer: I've updated the Boogie naming so it's simply more correct and also slightly more verbose. It will say `implementation Ex/focus[16]/split@15()` to mean, the implementation Ex, using the `focus` at line 16, and then the `split_here` at line 15.

One thing I struggle with is the overloading of the term split. Split is used to describe both any type of VC partitioning: `focus`/`split_here`/`isolate`/`isolate "paths"`, but also occurs in the name "split_here". Should I rename `{:split_here}` to `{:cut}` ?

robin-aws: Naming is much better, thanks!

Heavily agreed on the overloading of "split" - it fits better as an operation name rather than a name for the things it produces. I think `{:split_here}` is reasonable since it's referring to the operation, it's labelling the VCs as "splits" that's a bigger problem.

keyboardDrummer: I initially wanted to rename the produced things to `Partition`, but somehow I kept referring to things as splits. I feel it would be easiest to rename `{:split_here}` to `{:cut}`, but I'm willing to rename on the other side.

robin-aws (approved):
Great stuff!

PR #975, opened 2024-10-24
Fix bug in datatype translation

shazqadeer:
SMTLib has a convention regarding the syntax for specifying the test for a variant arm of a datatype. If the variant constructor name is C, a call to function is-C must be generated. This was not being done correctly and this PR fixes the problem.

This PR also renames SmtLibNameUtils --> SMTLibNameUtils to match the convention used for other classes in the SMTLib project.


keyboardDrummer (approved):

keyboardDrummer:
Would be good to add a test as well

PR #979, opened 2024-11-06
Provide more flexibility in configuring which return statements get separate VCs with isolate_assertions

keyboardDrummer:
### Changes
Enable isolate_assertions to isolate explicit return statements in Dafny programs

### DIFF
--- a/Source/Core/Generic/TokenWrapper.cs
+++ b/Source/Core/Generic/TokenWrapper.cs
@@ -14,7 +14,7 @@ public class TokenWrapper : IToken {
     return Inner.CompareTo(other);
   }
 
-  public bool SourceToken => false;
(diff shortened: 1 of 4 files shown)


Source/Core/Generic/TokenWrapper.cs
```diff
@@ -14,6 +14,8 @@ public int CompareTo(IToken? other) {
     return Inner.CompareTo(other);
   }
 
+  public bool SourceToken => false;
```
robin-aws, line 17:
```suggestion
  public bool IsSourceToken => false;
```

I initially expected this to be a token rather than a boolean.

robin-aws (changes requested):

PR #983, opened 2024-11-13
Support multi-binding `var` expressions in Lean backend

atomb:
This allows expressions of the form `var x, y := e1, e2; e3` to be translated to Lean.

Also bumps the versions of lean and lean-auto.


Source/Provers/LeanAuto/LeanAutoGenerator.cs
```diff
@@ -364,17 +364,21 @@ public override Expr VisitLambdaExpr(LambdaExpr node)
...
+      WriteText("; ");
     }
-    WriteText("(let");
-    node.Dummies.ForEach(x => Visit(x.TypedIdent));
-    WriteText(" := ");
-    node.Rhss.ForEach(e => Visit(e));
-    WriteText("; ");
     Visit(node.Body);
-    WriteText(")");
+    bindings.ForEach(b => WriteText(")"));
```
keyboardDrummer, line 381: IMO `.ForEach` is an anti-pattern, since the IDE and debugger support is worse than when using the language-construct `foreach`

keyboardDrummer (approved):

PR #998, opened 2025-02-11
Cce rename

keyboardDrummer:
Rename class `cce` to `Cce`


Source/CodeContractsExtender/cce.cs
robin-aws, line 1: Why not rename the file too?

robin-aws (commented):
> Rename class cce to Cce

...and upgrade to .Net 8? :) Is there any installation/setup documentation that should be updated too?

MikaelMayer (approved):
All good, .net version bump and class rename is ok for me. I checked every file.

Issue #1001, opened 2025-02-24
Type checker is not flagging missing modifies


NamrathaG:
I missed adding the modifies clause to my mover procedure (voting_loop_yield_procedure), and because of that, my yield invariant passes when it should not. I think the type checker needs to be fixed.

Here is the code:

```
type TransactionId;
type ReplicaId = int;

datatype Vote { YES(), NO(), NONE() }
datatype VoteRequest { VoteRequest(xid: TransactionId, rid: ReplicaId) }
datatype VoteRequestToken { VoteRequestToken(xid: TransactionId, rid: ReplicaId) }
datatype Decision { COMMIT(), ABORT() }

const Conflict: [TransactionId][TransactionId]bool;
const n: int;
axiom 0 < n;
axiom (forall xid1: TransactionId, xid2: TransactionId :: Conflict[xid1][xid2] == Conflict[xid2][xid1]);
axiom (forall xid0: TransactionId :: Conflict[xid0][xid0] == true);

var {:layer 0,1} replica_votes: [VoteRequest]Vote;
var {:layer 0,1} locked_transactions: [ReplicaId](Set TransactionId);
var {:layer 0,1} committed_transactions: Set TransactionId;

function noConflicts(xids: (Set TransactionId), xid: TransactionId) : bool {
    !(exists x: TransactionId :: Set_Contains(xids, x) && Conflict[x][xid])
}

function allVoteRequests(xid: TransactionId) : Set VoteRequest {
    Set((lambda vr:VoteRequest :: vr->xid == xid &&  1 <= vr->rid && vr->rid < n+1))
}

function remainingVoteRequests(xid: TransactionId, i:int) : Set VoteRequest {
    Set((lambda vr:VoteRequest :: vr->xid == xid &&  i <= vr->rid && vr->rid < n+1))
}


yield procedure {:layer 0} voting_procedure ({:linear} vr: One VoteRequest);
refines voting_action;


yield right procedure {:layer 1} voting_loop_yield_procedure(xid: TransactionId, {:linear_in} vrs: Set VoteRequest) returns ({:linear} vrs': Set VoteRequest)
requires {:layer 1} vrs == allVoteRequests(xid);
ensures {:layer 1} (forall j:int :: 1 <= j && j < n+1 && replica_votes[VoteRequest(xid, j)] == YES() ==> Set_Contains(locked_transactions[j], xid));
ensures {:layer 1} vrs' == allVoteRequests(xid);
// modifies locked_transactions;
{
[...]

PR #1004, opened 2025-02-28
Move modifies inference immediately after resolution

shazqadeer:
- Fix issue #1001 by moving modifies inference earlier so that type checking can use that information.
- Change command-line option /doModSetAnalysis to /inferModifies.
- Improve documentation of cache coherence example.


shazqadeer:
@wrwg, @rakamaric, @akashlal : Note that this PR changes the name of the option ```/doModSetAnalysis``` to ```/inferModifies```.

PR #1007, opened 2025-03-04
[Civl] Remove ISR code

shazqadeer:


Source/Concurrency/InductiveSequentialization.cs
bkragl, line 1: Can `enum InductiveSequentializationRule` be removed now too?

bkragl (commented):
So is the verdict that the merits of a right-mover based rule are too limited?

Issue #993, opened 2024-12-19
boogie does not terminate


rahxephon89:
Hi, when I tried to verify the following program [boogie.bpl](https://github.com/user-attachments/files/18192730/boogie.bpl.txt) using the command:
```
boogie -doModSetAnalysis -printVerifiedProceduresCount:0 -printModel:1 -enhancedErrorMessages:1 -proverOpt:O:model_validate=true -proverOpt:PROVER_PATH=~/z3 -useArrayAxioms -proverOpt:O:smt.QI.EAGER_THRESHOLD=100 -proverOpt:O:smt.QI.LAZY_THRESHOLD=100 -vcsCores:4 -proverOpt:O:trace=true -proverOpt:O:trace_file_name=z3.trace -proverLog:@PROC@.smt boogie.bpl
```

boogie got stuck without terminating. Any hints on why it happens and/or how to debug it further? Thanks!

System spec:
```
Arch: Arm64
OS: Mac OS 15.1.1 
boogie version: tried both 3.2.4 and latest on main
z3 version: tried both 4.11.2 and 4.13.4
```

wrwg:
It appears the problem here also manifests by ever growing memory (Boogie grows over 100g in some cases). Any idea what could cause this, and a possible option to turn off/on to prevent the memory growth?

rahxephon89:
cc @shazqadeer

shazqadeer:
The file is large. There is no magic. If I were to debug this problem, I would start by trying to minimize the size while not losing the issue. Can you try to do a bit of this? Then I will also try to help.

How many verification problems are present in the file?

Is there any non-linear arithmetic? If not, most likely the problem is due to quantifiers.

wrwg:
The problem is not z3 here, with quantifiers and the like, but Boogie. We are observing that boogies memory grows unbounded (100GB and growing). Wondering what could cause this? Is there an algorithm inside Boogie which does constant rewriting of the VC such that it grows without bounds? We are used to this from Z3 but not from Boogie.

We also seeing no active z3 instance running and still Boogie using compute.

shazqadeer:
Thanks for that clarification. Very interesting! I will take a look.

wrwg:
Here is another [boogie.bpl (1).txt](https://github.com/user-attachments/files/18470093/boogie.bpl.1.txt) which should clearly show the memory explosion, run as:

```
[...]

PR #1009, opened 2025-03-09
Eliminate redundant computation in VC experession translation

shazqadeer:
This PR is in response to issue #993 which uncovers unexpected long-running execution during the translation of a large Expr into a VCExpr. My intuition suggested that the expression translation should do linear work unless the expression being translated involves deeply nested inlined function calls. Linear complexity would not exhibit the behavior being seen since the example does not have deeply nested function calls.

The root cause appears to be a piece of code that examines the current term substitution whenever the translation scope crosses a binding (let-binding or quantifier). This examination is done to collect free variables in the substituted expressions to check whether they might collide with the bound variables in the binding about to be translated. In the example, the function ConcatVec is being invoked with increasingly larger expressions for its arguments which makes the substitution mapping formals to actuals increasingly larger. There is a let binding inside ConcatVec which triggers the examination of the increasingly larger substitutions. This ultimately leads to quadratic complexity in translation.

I concluded that the collision detection is unnecessary because it will never detect any collisions in the scenarios where it is being invoked. The substitutions should never refer to the bound variables of the binding being translated. To test my understanding, I ran the Boogie tests with a Debug.Assert capturing my hypothesis. All tests passed.

This PR removes the redundant collision detection code.

### DIFF
--- a/Source/VCExpr/VCExprAST.cs
+++ b/Source/VCExpr/VCExprAST.cs
@@ -373,7 +373,7 @@ namespace Microsoft.Boogie.VCExprAST
   {
     public override string ToString()
     {
-      return $"${Op}(${String.Join(", ", Arguments)})";
+      return $"{Op}({String.Join(", ", Arguments)})";
     }
 
     [Pure]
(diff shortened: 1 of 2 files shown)


RustanLeino (approved):
Your reasoning looks correct to me. That is, I agree that, for a well-formed Boogie program, no collision would be possible. So, it seems fine to remove that logic.

shazqadeer:
Fixes issue #993

PR #1014, opened 2025-04-11
2pc protocol with par cmd

NamrathaG:
This PR accomplishes a proof of the 2pc protocol using a proof rule for the parallel cmd for summarizing unbounded number of right mover voting actions. The resulting proof is more succinct that previous attempts based on pending asyncs. This proof also has the desirable property that it works for arbitrary number of instance of 2pc potentially running concurrently.

After this PR lands, this is the only proof of 2PC in the repository.


shazqadeer:
As we have learned, the proof is sound only because the the voting action does not fail. This is currently not checked by Civl so we have to manually check it with our eyes.

shazqadeer (approved):

PR #1023, opened 2025-05-25
[Civl] Change signature of Alloc

shazqadeer:
The purpose of this PR is to make Alloc return the right loc piece.

While making the change, I discovered a crash in the type checker. Added a fix for that.

### DIFF
--- a/Test/civl/large-samples/treiber-stack.bpl
+++ b/Test/civl/large-samples/treiber-stack.bpl
@@ -6,39 +6,44 @@ function {:inline} AllLocPieces(): Set LocPiece {
   Set_Add(Set_Add(Set_Empty(), Left()), Right())
 }
 
+type LocTreiberStack = KeyedLoc LocPiece;
 type LocTreiberNode = KeyedLoc LocPiece;
 type StackElem T = Node LocTreiberNode T;
 type StackMap T = Map LocTreiberNode (StackElem T);
 datatype Treiber<T> { Treiber(top: Option LocTreiberNode, {:linear} nodes: StackMap T) }
 
+function {:inline} LeftLocPiece(loc_piece: KeyedLoc LocPiece): KeyedLoc LocPiece {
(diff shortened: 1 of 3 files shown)


Test/civl/large-samples/treiber-stack.bpl
```diff
@@ -6,39 +6,44 @@ function {:inline} AllLocPieces(): Set LocPiece {
   Set_Add(Set_Add(Set_Empty(), Left()), Right())
 }
 
+type LocTreiberStack = KeyedLoc LocPiece;
 type LocTreiberNode = KeyedLoc LocPiece;
 type StackElem T = Node LocTreiberNode T;
 type StackMap T = Map LocTreiberNode (StackElem T);
 datatype Treiber<T> { Treiber(top: Option LocTreiberNode, {:linear} nodes: StackMap T) }
 
+function {:inline} LeftLocPiece(loc_piece: KeyedLoc LocPiece): KeyedLoc LocPiece {
```
bkragl, line 15: We're not using `KeyedLoc LocPiece` directly anywhere else. Should we use a type synonym and rename the function appropriately?

bkragl (commented):
Nice! So my understanding is that, with this change, Alloc returns a linear value, which can be useful to a client, but there is no restriction for other operations. The allocated liner value would be "demoted" to ordinary for those operations.

Test/civl/large-samples/treiber-stack.bpl
shazqadeer: Yes, good idea. I will make that change.

shazqadeer:
A general comment regarding the (dropped) type parameter for Loc:

It was dropped because it did not seem to offer any benefit for verification. However, it looks likely that an erasure to heaps could benefit from knowing the type of the value to which a Loc is pointing.  The obvious strategy for generating code for heaps would allocate storage for this value when Loc is allocated and the size of the storage could be inferred from the type.

PR #1037, opened 2025-07-24
Fix `BigFloat.FromInt` incorrectly using string constructor

fabiomadge:
This PR fixes a critical bug in `BigFloat.FromInt()` where it improperly used the string constructor, causing arithmetic failures. The fix introduces IEEE 754-compliant conversions, removes redundant fields, and adds 177 comprehensive tests to ensure correctness.

### The Bug: FromInt Misused the String Constructor

The `value` field in BigFloat was designed to store only special values (`"NaN"`, `"+oo"`, `"-oo"`), with empty string (`""`) indicating a normal numeric value. However, `FromInt` was incorrectly implemented:

```csharp
// OLD BROKEN CODE:
public static BigFloat FromInt(int v) {
    return new BigFloat(v.ToString(), 24, 8);  // e.g., "42"
}
```

This stored regular integers as strings in the value field (e.g., `"42"`), breaking the invariant that value should only contain special values or be empty. This caused arithmetic operations to fail because:

1. The string constructor didn't initialize the IEEE 754 fields (significand, exponent)
2. Arithmetic operations' special value handling would see `value != "`" and try to handle `"42"` as a special value
3. The actual numeric representation was never properly initialized

The Fix

- Changed FromInt to use proper IEEE 754 conversion via FromRational
- Removed the problematic string constructor that allowed non-special values

Other Improvements

1. Parameter Order Fix: Corrected FromBigInt(value, exponentSize, significandSize) to have consistent parameter order
2. Removed Dual Representation: Eliminated the value field entirely in favor of proper IEEE 754 representation for all values
3. IEEE 754 compliant arithmetic in all cases
4. Comprehensive Tests: Added 177 tests covering integer conversion, arithmetic, and edge cases

This fix ensures that integer-to-float conversions work correctly and participate properly in arithmetic operations.


shazqadeer (dismissed):
I am approving the PR with low confidence since I am not an expert in floating point. Perhaps @zvonimir, who is familiar with both floating point and Boogie, can also take a look.

shazqadeer:
@fabiomadge : Thanks for the PR. Please look into why some of the regression tests are failing.

fabiomadge:
On it, thanks!

fabiomadge:
@shazqadeer I spoke with @zvonimir and he won't immediately get to it. If you're ok with it, I'd like to merge now and revisit this, in case we end up finding any issues.

shazqadeer (dismissed):

fabiomadge:
I don't have a compelling explanation for what went wrong, but installing .NET 6 on the runner solved the build issue we encountered after merging the README.md changes.

shazqadeer:
I have seen the CI to be flaky in the past occasionally. A retry can often do the job. Please retry after reverting your change to the runner. I would like to see if that fixes the problem.

fabiomadge:
It took three retries, but as finally worked.

shazqadeer (approved):

PR #1057, opened 2025-08-31
fix: Printing `Add` and `Mul` with non-associative types

fabiomadge:
`+` and `*` print left-associatively, so the printer dropped the parentheses of `x + (y + z)` and emitted `x + y + z`. That re-parses as `(x + y) + z` — the same expression only when the regrouping is valid. Two cases where it is not:

**Floating point.** `+` and `*` are not associative:

```boogie
procedure main() {
  var x, y, z, a, b: float24e8;
  a := x + (y + z);
  b := (x + y) + z;
  assert a == b;      // correctly fails: 0 verified, 1 error
}
```

Through `/print`, both assignments printed as `x + y + z` and the result verified — **1 verified, 0 errors**. The counterexample disappeared.

**Integers.** Not only a float problem. `i * (j div k)` printed as `i * j div k`, which means `(i * j) div k`: `2 * (3 div 2) = 2`, but `(2 * 3) div 2 = 3`. Same for `mod`, and for real `r * (s / t)` vs `(r * s) / t` at `t = 0`.

## Fix

Parentheses around a right operand of equal binding power are dropped only when the operand has an associative type (`int`/`real`) **and** applies a regrouping operator: `+` regroups with `+` and `-`, `*` only with `*`.

Both conditions are needed — the type check alone misses `div`/`mod`/`/`, the operator check alone misses floats.

`/print` runs before typechecking, so no type is known there and the parentheses stay. `Expr.ShallowType` is not an alternative: `IdentifierExpr.ShallowType` asserts `Type != null` and aborts the process.

## Testing

`Test/test0/PrintAssoc.bpl` is a golden file beside the existing `PrettyPrint.bpl`, covering valid regrouping, same precedence but truncating, and non-associative types. `-printDesugared` prints both before and after typechecking, pinning down both phases.

`Source/UnitTests/CoreTests/ExprPrintRoundTrip.cs` checks the property rather than the output text: for every arithmetic operator at every operand type it typechecks at, print the nested expression, re-parse, compare ASTs by `ContentHash`. 228 cases over two passes — the first permits the sound regroupings, the second parses without typechecking, where no type is known and nothing may be reassociated.

That second layer is worth keeping even if the fix changes shape. The parenthesization table is hand-maintained, an error in it is invisible (the program still prints, just as a different expression), and the same table governs `-`, `div`, `mod`, `/` and `**`. Against `master` it fails 20 of 228 cases across six shapes — `a + (b + c)`, `a + (b - c)`, `a * (b * c)`, `a * (b div c)`, `a * (b mod c)`, `a * (b / c)` — each reporting the offending printout.

The rule is tight, not merely safe: the 5 regroupings it permits all verify as equalities, and the 7 it rejects all genuinely differ.

No lit regressions. `z3-hard-timeout`, `CaptureInlineUnroll` and `InterestingExamples4` fail on this branch and identically on `master`; `prover/exitcode.bpl` is load-sensitive under `-j 8` but passes in isolation.

### DIFF
--- a/Source/Core/AST/Expression/AbsyExpr.cs
+++ b/Source/Core/AST/Expression/AbsyExpr.cs
@@ -2162,6 +2162,45 @@ namespace Microsoft.Boogie
       }
     }
 
+    /// <summary>
+    /// Decides whether the right operand of "op" ("+" or "*") needs parentheses, given that it binds
+    /// equally strongly. Since "+" and "*" are printed left-associatively, dropping the parentheses of
+    /// "x op (y op' z)" makes it re-parse as "(x op y) op' z", so they may only be dropped when those
+    /// two groupings agree.
+    ///
+    /// On int and real they agree whenever "op'" is the inverse-free counterpart of "op": "+" regroups
+    /// with "+" and "-", and "*" regroups only with "*". "i * (j div k)" keeps its parentheses, because
+    /// "i * j div k" means "(i * j) div k", which differs as soon as the division truncates; likewise
+    /// "r * (s / t)" differs from "(r * s) / t" when t is 0.
+    ///
+    /// Floating-point "+" and "*" are not associative at all -- rounding makes "x + (y + z)" and
+    /// "(x + y) + z" genuinely different -- so on any other type the parentheses always stay.
+    ///
+    /// A null type means Typecheck has not run yet (e.g. /print, which prints before typechecking), in
+    /// which case we conservatively keep the parentheses.
+    /// </summary>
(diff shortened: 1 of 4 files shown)


shazqadeer:
@fabiomadge : Do you want to commit this PR? If yes, please add a brief description.

shazqadeer (approved):

shazqadeer:
@fabiomadge : Please rebase and let me know that you are ready to land.

fabiomadge:
@shazqadeer Thanks for the review! As far as I'm concerned, the PR is ready to land now.

PR #1090, opened 2025-12-14
Extend trust options

NamrathaG:
Added \trustInvariants option. 
Updated \trustRefinement option to also include \trustSequentialization.

### DIFF
--- a/Source/Concurrency/CivlRewriter.cs
+++ b/Source/Concurrency/CivlRewriter.cs
@@ -26,7 +26,6 @@ namespace Microsoft.Boogie
       // Gate sufficiency checks
       Action.AddGateSufficiencyCheckers(civlTypeChecker, decls);
 
-      // Commutativity checks
       civlTypeChecker.AtomicActions.ForEach(x =>
       {
         decls.AddRange(new Declaration[] { x.Impl, x.Impl.Proc, x.InputOutputRelation });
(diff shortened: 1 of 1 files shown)


Source/Concurrency/CivlRewriter.cs
```diff
@@ -43,9 +43,17 @@ public static void Transform(ConcurrencyOptions options, CivlTypeChecker civlTyp
...
+      {
+         YieldingProcChecker.AddInvariantCheckers(civlTypeChecker, decls);
+      }
+      
+      if (!options.TrustRefinement)
+      {
+         YieldingProcChecker.AddRefinementCheckers(civlTypeChecker, decls);
+      }
+      
+      if (!options.TrustSequentialization && !options.TrustRefinement)
```
shazqadeer: Please move this code inside the then case of line 51 so that it is more obvious that one is a subcase of the other.

Source/Concurrency/CivlRewriter.cs
shazqadeer: Not an issue with your PR. But please move the comment about Commutativity checks on line 29 to the correct place.

Source/Concurrency/YieldingProcChecker.cs
```diff
@@ -5,11 +5,11 @@ namespace Microsoft.Boogie
 {
   public class YieldingProcChecker
   {
-    public static void AddCheckers(CivlTypeChecker civlTypeChecker, List<Declaration> decls)
+    public static void AddInvariantCheckers(CivlTypeChecker civlTypeChecker, List<Declaration> decls)
     {
       Program program = civlTypeChecker.program;
```
shazqadeer: You don't seem to be examining the TrustInvariants flag anywhere.

shazqadeer (dismissed):

PR #1127, opened 2026-05-15
Recursive list reversal; verification of well-formedness only

ankushdas:
This is a recursive list reversal function, inspired from Reynolds' list reversal example. The property verified is only well-formedness, i.e., if the input list is well-formed, then the reversed list is also well-formed. It does not verify if the reversed list is actually a reverse of the original list.


Test/datatypes/list-reversal-wf.bpl
```diff
@@ -0,0 +1,142 @@
+// RUN: %parallel-boogie -lib:base -lib:node "%s" > "%t"
+// RUN: %diff "%s.expect" "%t"
+
+/*
+  Linked list with nested permissions, translated from the NPL example in
```
shazqadeer: The comment does not match the PR description. Let us drop this comment entirely.

Test/datatypes/list-reversal-wf.bpl
```diff
@@ -0,0 +1,142 @@
...
+  pointer with linear ownership of all reachable nodes. The linear type system
+  enforces that no two List values share nodes, preserving the NPL invariant.
+*/
+
+datatype List { List(head: Option Loc, nodes: Map (One Loc) (Node int)) }
+
+// WF: lists are non-empty, the head node is in the map, and all reachable
+// nodes are in the map (so the structure is self-contained).
+function {:inline} WF(l: List): bool {
+  l->head is Some &&
```
shazqadeer: This conjunct indicates that l always has a first element. In that case, why not make the type of head Loc (since it is always Some)?

The other option is to leave the type signature unchanged and allow for empty lists also.

Test/datatypes/list-reversal-wf.bpl
```diff
@@ -0,0 +1,142 @@
...
+  Map_Contains(l->nodes, One(l->head->t)) &&
+  Reachable(l->nodes, l->head, None()) &&
+  InDomain(l->nodes, l->head)
```
shazqadeer: If you rebase with respect to the latest master, you will find that Reachable property is part of InDomain. So you should be able to drop the conjunct on line 27.

Test/datatypes/list-reversal-wf.bpl
```diff
@@ -0,0 +1,142 @@
...
+ensures WF(l_out);
+ensures l_out->nodes->dom == l_in->nodes->dom;
+{
+  var new_head: Loc;
+  var helper_out: List;
+  var helper_head: Option Loc;
+  var result_nodes: Map (One Loc) (Node int);
+
+  call new_head, helper_out := ReverseHelper(l_in, None());
+  List(helper_head, result_nodes) := helper_out;
```
shazqadeer: You could possibly use the Path_Store API to write 140-141 more compactly.

Test/datatypes/list-reversal-wf.bpl
```diff
@@ -0,0 +1,142 @@
...
+///
+/// Takes {:linear_in} l_in so that destructuring gives a LOCAL `nodes` variable,
+/// which is required for Map_Get/Map_Put to type-check (parameters cannot be
+/// used as linear paths to primitives — only local variables can).
+///
+/// Each call: rewrites the head node's next pointer to prev_head (the
+/// already-reversed prefix), then recurses on List(old_next, nodes).
+/// Returns (new_head, bag-of-reversed-nodes); new_head is the original tail.
+/// No new Loc is allocated: l_out->nodes->dom == l_in->nodes->dom.
+pure procedure {:vcs_split_on_every_assert} ReverseHelper({:linear_in} l_in: List,
```
shazqadeer: I thought we had decided that we will use two separate lists to do the job. We will keep removing items from one list and adding them to the reversed list.

shazqadeer (approved):

Issue #1066, opened 2025-10-27
Trivial copy assignment (y := x) is omitted from SMT named assumes → missing from unsat cores / coverage


ricostynha1:
# Reproduce 

From Dafny using this version:
4.11.0+fcb2042d6d043a2634f0854338c08feeaaaf4ae2
With this code:
```dafny
method Abs(x: int) returns (y: int)
  ensures x>=0 ==> x==y
  ensures x<0 ==> x==-y
{
  if x < 0 {
    y := -x;
  } else {
    y :=  x;
  }
}
```

Running the proof dependencies tool: 
dafny verify [file_name] --verification-coverage-report cov --log-format text --solver-option LOG_FILE=output.smt2 --bprint output.bpl

We obtain that line y:=x was unused by the proof. 

# Problem 

The thing is: 
In the generated Boogie implementation I see both assignment statements present:

```boogie
implementation {:smt_option "smt.arith.solver", "2"} {:verboseName "Abs (correctness)"} Impl$$_module.__default.Abs(x#0: int) returns (defass#y#0: bool, y#0: int, $_reverifyPost: bool)
{
  var $_ModifiesFrame: [ref,Field]bool;

    // AddMethodImpl: Abs, Impl$$_module.__default.Abs
    $_ModifiesFrame := (lambda $o: ref, $f: Field :: 
      $o != null && $Unbox(read($Heap, $o, alloc)): bool ==> false);
    assume {:captureState "abs.dfy(4,0): initial state"} true;
    $_reverifyPost := false;
    // ----- if statement ----- abs.dfy(5,3)
    assume true;
    if (x#0 < 0)
    {
        // ----- assignment statement -----  abs.dfy(6,7)
        assume true;
        assume true;
        y#0 := 0 - x#0;
        defass#y#0 := true;
        assume {:captureState "abs.dfy(6,11)"} true;
    }
    else
    {
        // ----- assignment statement -----  abs.dfy(8,7)
        assume true;
        assume true;
        y#0 := x#0;
        defass#y#0 := true;
        assume {:captureState "abs.dfy(8,11)"} true;
    }

    assert {:id "id12"} defass#y#0;
}
```

In the SMT translation (output.smt2) the arithmetic assignment (y := -x) leads to a small, named assume that can be extracted in unsat cores / coverage. Example pattern (from SMT):

```smt2
assert (! $generated@@93 :named aux$$assume$$id10))
[...]

PR #1131, opened 2026-05-21
Fix: preserve {:id} for trivial assignments during passification

ricostynha1:
When TrackVerificationCoverage is enabled, assignments with trivial RHS expressions (literals or simple variable copies like y := x) are never reported in proof dependencies / unsat cores.

During passification, Boogie optimizes these assignments by directly updating the incarnation map without generating an AssumeCmd. This means the {:id} attribute attached to the original AssignCmd (used by Dafny's --verification-coverage-report) is silently discarded — no assume is emitted, so no named assume reaches the VC, and Z3 never reports it in (get-unsat-core).

Non-trivial RHS expressions (e.g., y := -x) take the else-branch which creates an incarnation variable, emits assume y@1 == expr, and copies the :id. These work correctly.

# Fix:
In ConditionGeneration.cs TurnIntoPassiveCmd, skip the incarnation-map shortcut for LiteralExpr and IdentifierExpr RHS when:

- TrackVerificationCoverage is enabled, AND
- The AssignCmd carries an {:id} attribute

This forces creation of a proper incarnation variable and AssumeCmd (e.g., assume y@1 == x@0), which carries the :id through Wlp → NamedAssumeOp → SMT :named → unsat core → coverage report.

The shortcut is preserved for all other cases (no performance impact when coverage tracking is off).

# Reproducing on dafny : 
check issue report:  https://github.com/boogie-org/boogie/issues/1066

# Now: 
with the fix it works expected and produces coverage reports logs that marked line was never reported as covered covered.
(what is what is aspected as the tracking id of y := x exists)

```dafny
method Abs(x: int) returns (y: int)
  ensures x >= 0 ==> x == y
  ensures x < 0 ==> x + y == 0
{
  if x < 0 {
    y := -x;
  } else {
    y := x;  // was never reported as covered
  }
}
```

This fixes this issue:
https://github.com/boogie-org/boogie/issues/1066

And affects Dafnys:
--verification-coverage-report feature


shazqadeer (approved):

atomb:
Thank you for fixing this, @ricostynha1!

PR #1141, opened 2026-08-07
fix: Round `BigFloat` arithmetic once, against a sticky bit

fabiomadge:
`BigFloat`'s arithmetic rounds twice. The first rounding discards the remainder the second needs to tell "exactly halfway" from "just above halfway", so round-to-nearest-even breaks ties the wrong way. IEEE-754 requires one rounding against the exact residual.

Against an oracle that computes in `BigInteger` and rounds once, over 40000 random operand pairs:

| | before | after |
|---|---|---|
| `operator/` at (24,8) | 18.67 % wrong by ≥1 ULP | **0.00 %** |
| `operator/` at (53,11) | 18.88 % | **0.00 %** |
| `FromRational` at (24,8) | 4.65 % | **0.00 %** |
| `FromRational` at (53,11) | 4.65 % | **0.00 %** |

`operator+`/`-`/`*` truncate rather than carry a sticky bit, also 0.00 % after. That rate depends on the operands: the harness above uses integers, which do not trigger it, while subnormal and widely-separated ones do.

## Scope

The four operators are unreachable from `.bpl` input — only `BinaryOperator.Evaluate` reaches them, and nothing in the product calls it. Real float arithmetic happens in the solver, on `fp.*` terms.

`FromRational` **is** reachable. Under `/infer:j` a float variable live across a loop head acquires integer bounds from `TryFloorCeiling`, and `Node.ToExpr`'s float branch renders them back through `NumberToExpr` → `FromBigInt` → `FromRational`.

It cannot observe the defect. Those bounds are floors and ceilings of floats, whose integer parts always fit their own format, so nothing is rounded — exact for every bound emitted across 400 `Test/` files, one of them `2^72`. An inexact bound would hold anyway: a float variable takes only representable values, so `Lo <= f` with adjacent representables `a < Lo < b` already gives `f >= b`, and either neighbour is true.

Not a soundness bug, then, but for a narrower reason than "nothing calls it". Wiring up `Evaluate`, or handing the domain a bound that is not a float's own floor, would change that.

## Fix

`RoundToFormat` is the only place that rounds. Callers compute with guard bits, record a sticky bit via `WithStickyBit`, and hand the result over once. `NormalizeAndRound`, `ApplyRoundToNearestEven` and `HandleExponentBounds` are gone; `HandleUnderflow` delegates rather than rounding for itself.

`FromRational` no longer infers exactness from what the rounding did — that is what tangled the two jobs. Exactness is a property of the input, which `RepresentsExactly` settles by cross-multiplying the result back against the original fraction in integers.

Wide formats never materialise astronomical values: a 40-bit exponent puts the smallest subnormal at `2^-549755813909`, where neither `2^shift` nor the shifted value fits a `BigInteger`, so those cases are decided by comparison. `SubnormalArithmeticIsBiasIndependent` throws `OverflowException` if that path goes.

## Parsing

One behaviour change is visible from `.bpl` input. The old parser decided "below the subnormal grid" from the exponent *before* rounding, so it flushed the band between half and one times the smallest subnormal to zero, and rejected it in strict mode. Of 11664 literals swept through both modes on both branches, 22 differ, all in that band: `0x6.0e-38f24e8` is three quarters of `2^-149` and now rounds up to it. Nothing the old parser accepted changed value and nothing it accepted is now rejected — some former parse errors work. `BoogiePL.atg` takes the strict path, so that is the side users see.

## Testing

`BigFloatDifferentialTests.cs` is the prerequisite, not the follow-up: perturbing division by 1 ULP failed 3 of the existing tests, and a correct fix (18.67 % → 0.00 %) failed none. Two oracles — exact `BigInteger`, and hardware `float`/`double` — are cross-checked by `OraclesAgreeWithEachOther` first, since a `(float)((double)n/d)` reference double-rounds and reproduces the very bug. Coverage spans significand sizes 2–113 and exponent sizes 5–100.

Stryker.NET over `BigFloat.cs` puts the suite at **87.78 %**: 562 of 663 mutants killed. Most of the 81 survivors are exception-message text, `Contract.Requires` this build compiles out, or boundary flips where both branches provably agree.

Test counts rise 200 → 231. Full `lit`: 671 tests, failing only `z3-hard-timeout`, `CaptureInlineUnroll` and `InterestingExamples4`, which fail identically on pristine `master`.

## Not addressed

- **No rounding-mode parameter.** Every operator hardcodes RNE, so every rate above is RNE only. Modes reach the solver directly and never touch `BigFloat`, but retrofitting one changes every signature.
- **Whether `BigFloat` arithmetic should exist**, given it has one caller that cannot tell the difference. That needs either a rounding-sensitive consumer or deletion in favour of a value/parse/print type; this PR fixes the arithmetic without settling it.

### DIFF
--- a/Source/BaseTypes/BigFloat.cs
+++ b/Source/BaseTypes/BigFloat.cs
@@ -682,15 +682,10 @@ namespace Microsoft.BaseTypes
         return CreateZero(resultSign, x.SignificandSize, x.ExponentSize);
       }
 
-      // Normalize and round the product
-      var (normalizedProduct, normalShift) = NormalizeAndRound(product, BigInteger.Zero, x.SignificandSize);
-
-      // Calculate the final exponent - all values are already BigInteger
-      var bias = x.bias;
-      var resultExp = xExp + yExp - bias + normalShift - (x.SignificandSize - 1);
-
-      // Handle overflow, underflow, and create final result
-      return HandleExponentBounds(normalizedProduct, resultExp, resultSign, x.SignificandSize, x.ExponentSize);
+      // The product is exact, so RoundToFormat performs the only rounding. Each prepared significand
+      // is weighted by 2^(exp - bias - (significandSize - 1)), and multiplying adds those weights.
+      var productScale = xExp + yExp - 2 * x.bias - 2 * (x.SignificandSize - 1);
+      return RoundToFormat(product, productScale, resultSign, x.SignificandSize, x.ExponentSize);
     }
 
     [Pure]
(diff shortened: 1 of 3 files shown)


shazqadeer:
@fabiomadge : I need some context behind this PR. Boogie, in general, does not do numerical evaluation. It typically would translate expressions to SMTLIB format and leave all evaluation to the SMT solver. This PR appears to be fixing some evaluation bug related to rounding. Where is such evaluation needed?

Another unrelated question: Was this PR generated with the use of some AI tool?

fabiomadge:
I rewrote big parts of the float implementation in Boogie to add them to Dafny. I extended the old interface, but didn't deprecate anything. Unfortunately we can't exclusively rely on the SMT solver. `AbstractInterpretation` needs arithmetic on floats. Dafny also doesn't use the String API for literals.

I used Claude Code to implement the changes.

shazqadeer:
@fabiomadge : Are you the original implementor of the code in BigFloat.cs? If that is the case, I would trust your judgment enough to approve this PR. If  that is not the case, we need some input from the original implementor. The changes are significant and I don't understand any of this.

fabiomadge:
I'm to blame for all the lines in BigFloat.cs, which this PR touches.

shazqadeer:
@fabiomadge : Please make a pass through the AI-generated documentation in BigFloat.cs to make it proportionate. I will trust your judgment on both the code and documentation. If you are satisfied, I will approve the PR.

shazqadeer:
@fabiomadge : Are you ready to land this PR?

fabiomadge:
I think it's ready to land

shazqadeer (approved):

PR #1142, opened 2026-08-07
fix: Crash on wide-exponent floats in abstract interpretation, and a missing size check

fabiomadge:
Two independent `BigFloat` fixes, one commit each. Both apply to `master` directly.

## 1. Abstract interpretation crashes on a wide-exponent float literal

Abstract interpretation asks every literal for integer bounds. Exponent sizes are unbounded, so for a float those bounds are too. From five lines of `.bpl` under `-infer:j`:

```boogie
procedure main() returns (r: int) {
  var f: float24e40;
  f := 0x1.0e137438953471f24e40;
  if (f == 0x1.0e137438953471f24e40) { r := 1; } else { r := 0; }
}
```

```
Unhandled exception. System.OverflowException: Arithmetic operation resulted in an overflow.
  at System.Numerics.BigInteger.op_LeftShift(BigInteger, Int32)
  at Microsoft.BaseTypes.BigFloat.FloorCeiling(BigInteger&, BigInteger&)
  at ...NativeIntervalDomain.PEVisitor.VisitLiteralExpr(LiteralExpr) IntervalDomain.cs:970
```

| | before | after |
|---|---|---|
| `float24e32`, peak RSS | 1.4 GB | **80 MB** |
| `float24e40` | **unhandled `OverflowException`** | verifies |
| `float24e8` (ordinary) | 80 MB | 80 MB |

**The allocation is not waste.** `FloorCeiling` allocates a constant 3× the size of the answer it returns, and `floor(2^2147483644)` genuinely *is* a 256 MB integer. So the limit belongs to callers that cannot use a bound that large, not to the computation.

`TryFloorCeiling` declines above a width bound, **deciding from the exponent alone** so the oversized value is never built, and returns false rather than throwing on NaN and the infinities. The interval domain treats declining as "no bound" — what it already does for NaN and infinity two lines above. `FloorCeiling` keeps its exact contract.

**Testing.** `TryFloorCeiling` agrees with `FloorCeiling` on all **96695 values it accepts** across five formats, 0 mismatches: the limit decides whether to answer, never what the answer is. Declining costs 0 MB at exponent sizes 24, 32, 40 and 60. `Test/floats/WideExponentBounds.bpl` covers the end-to-end path, since the unit tests cannot reach `IntervalDomain`; it **fails on a pristine `master` worktree** and passes here. Removing the width guard is caught by both tests.

## 2. `CopySign` skips the size-compatibility check every sibling enforces

```
CopySign(float24e8, float53e11)   // returns -0x1.0e0f24e8, no exception
```

The diff only shows the line being added, so to be explicit about what was already there. Enumerating every public two-operand entry point by reflection and calling each with mismatched formats across plain, NaN, infinity and zero operands:

| operation | plain | either operand NaN |
|---|---|---|
| `+` `-` `*` `/`, `CompareTo` | reject | reject |
| `CopySign` (**this change**) | reject | reject |
| `==` `!=` `<` `>` `<=` `>=`, `Equals` | reject | **accept** |
| `Max`, `Min` | reject | **accept** |

Nothing about copying a sign *needs* matching formats, so this buys uniformity rather than correctness — but an operation that quietly accepts what its neighbours reject is the one a size confusion travels through.

**The NaN column is a pre-existing gap, left for a separate change.** Everything except the arithmetic operators reaches the check only through `CompareTo`, and all of them test for NaN first, short-circuiting before it. For the comparisons this is harmless: IEEE 754 makes every NaN comparison false regardless of format, and the returned answers are correct. `Max`/`Min` are the real defect —

```
Max(float24e8, NaN53e11)  ->  0NaN53e11    (a float24e8 operand yields a float53e11)
```

— so the result's format depends on the operands' *values* rather than their types. The arithmetic operators avoid this by validating before `HandleSpecialValues`; fixing `Max`/`Min` belongs with that special-value handling rather than in a sign-copy commit.

## Verification

Both commits build and pass individually. 200 `BaseTypesTests`, 59 `CoreTests`, 29 `ExecutionEngineTests`. Full `lit`: 669 tests, failing only `z3-hard-timeout`, `CaptureInlineUnroll` and `InterestingExamples4`, which fail identically on a pristine `master` worktree.

Merging all three float PRs together was checked in a scratch worktree: no conflicts, 234 `BaseTypesTests`, 670 `lit` tests, same three pre-existing failures.

### DIFF
--- a/Source/AbstractInterpretation/IntervalDomain.cs
+++ b/Source/AbstractInterpretation/IntervalDomain.cs
@@ -958,18 +958,19 @@ namespace Microsoft.Boogie.AbstractInterpretation
         }
         else if (node.Val is BigFloat)
         {
+          // NaN and the infinities have no integer bounds, and a float with a wide exponent has bounds
+          // too large to be worth computing -- the floor of the largest float24e32 is a 256 MB integer.
+          // Both are "no useful bound", which the interval domain already represents as null.
           var bf = (BigFloat) node.Val;
-          if (bf.IsNaN || bf.IsInfinity)
+          if (bf.TryFloorCeiling(out var floor, out var ceiling))
           {
-            // NaN and infinity have no meaningful integer bounds
-            Lo = null;
-            Hi = null;
+            Lo = floor;
+            Hi = ceiling;
           }
           else
           {
-            bf.FloorCeiling(out var floor, out var ceiling);
-            Lo = floor;
-            Hi = ceiling;
+            Lo = null;
+            Hi = null;
           }
         }
         else if (node.Val is bool)
--- a/Source/AbstractInterpretation/IntervalDomain.cs
+++ b/Source/AbstractInterpretation/IntervalDomain.cs
@@ -958,9 +958,6 @@ namespace Microsoft.Boogie.AbstractInterpretation
         }
         else if (node.Val is BigFloat)
         {
-          // NaN and the infinities have no integer bounds, and a float with a wide exponent has bounds
-          // too large to be worth computing -- the floor of the largest float24e32 is a 256 MB integer.
-          // Both are "no useful bound", which the interval domain already represents as null.
           var bf = (BigFloat) node.Val;
           if (bf.TryFloorCeiling(out var floor, out var ceiling))
           {
@@ -969,6 +966,7 @@ namespace Microsoft.Boogie.AbstractInterpretation
           }
           else
           {
+            // NaN, an infinity, or bounds too wide to compute
             Lo = null;
             Hi = null;
           }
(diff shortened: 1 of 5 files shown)


Source/BaseTypes/BigFloat.cs
```diff
@@ -851,9 +873,53 @@ private enum ArithmeticOperation
 
     #region Mathematical Operations
 
+    /// <summary>
+    /// Widest floor/ceiling <see cref="TryFloorCeiling"/> will produce, in bits.
+    ///
+    /// Exponent sizes are unbounded, so a float's integer part is too: at float24e32 the floor of the
+    /// largest finite value is a 256 MB integer, and at float24e40 it exceeds what BigInteger can
+    /// represent at all. Both are faithful answers to the question asked -- the cost is proportional to
```
shazqadeer: What is this documentation talking about? 

Which "question being asked" is it referring to?

The sheer size of the documentation is disproportionate to anything else in Boogie, even topics that are likely much more complex than that being addressed here.

Source/AbstractInterpretation/IntervalDomain.cs
```diff
@@ -958,18 +958,19 @@ public override Expr VisitLiteralExpr(LiteralExpr node)
         }
         else if (node.Val is BigFloat)
         {
+          // NaN and the infinities have no integer bounds, and a float with a wide exponent has bounds
```
shazqadeer: I fail to understand the point of this comment.

I understand the code change makes sense because floor and ceiling are used to initialize Lo and Hi only if they exist (I assume that is what the return value of TryFloorCeiling is indicating). There does not appear to any need for extra philosophizing.

fabiomadge:
Trimmed the comments as requested

shazqadeer (approved):

PR #1143, opened 2026-08-07
Fix crash on arithmetic over a nullary map select

shaobo-he:
A map type with no index arguments is represented by its result type:
`Visit(MapSelect)`/`Visit(MapStore)` translate a nullary select/store straight to
the underlying value, and `SMTLibLineariser.TypeToString` emits a nullary map as
the sort of its result type.

The translated expression still carries the map type, though, so anything that
inspects an operand's type sees a map where it expects a value type. Three crashes
follow, all `UnreachableException` from a type accessor:

```boogie
procedure P() { var x: []int;      assert x[] + 0 == 1; }          // Type.FloatSignificand
procedure P() { var x: []bv8;      assert (x[] ++ 1bv24) == 0bv32; } // Type.BvBits
procedure P() { var x: [][int]int; assert x[][1] + 0 == 0; }        // Type.FloatSignificand
```

1. `TranslateBinaryOperator` dispatches `IsInt` → `IsReal` → else *assume float*, so a
   nullary select reaches `Type.FloatSignificand`. Every other case in that method
   guards on `IsFloat` explicitly, which is why comparisons and integer division
   already worked.
2. `VCExpressionGenerator.BvConcat` and `VCExprBvConcatOp.InferType` read `Type.BvBits`
   off the same mistyped operand.
3. `VCExprSelectOp.InferType` takes `args[0].Type.AsMap` and returns its `Result`, so a
   select supplying indices to a nullary map wrapper returns the wrapper's result
   (`[int]int`) rather than what the indices actually produce (`int`). Unwrapping in the
   operator dispatch does not help here, since the inferred type is a one-argument map
   rather than a nullary one.

This adds a shared `UnwrapNullaryMapType` helper on `VCExpressionGenerator` and uses it
in all three places — the binary-operator dispatch (including `RealDiv`, which reads the
argument types separately to decide whether to coerce to real), bitvector concatenation,
and `VCExprSelectOp.InferType` when the select supplies indices.

I swept the rest of the nullary-map operator surface — arithmetic, division, comparison,
boolean, bitvector extract and concat, float, coercions, nesting to three levels, maps of
nullary maps, store, function arguments, quantifiers, polymorphic nullary maps — and no
other crash remains.

`Test/test2/NullaryMaps.bpl` covers comparison on a nullary select but not arithmetic,
which is how these went unnoticed. The new test exercises each fixed path with a matching
failing case, so the obligations cannot be discharged vacuously.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Source/VCExpr/VCExpressionGenerator.cs
+++ b/Source/VCExpr/VCExpressionGenerator.cs
@@ -483,12 +483,34 @@ namespace Microsoft.Boogie
       return Function(new VCExprBvExtractOp(start, end, bits), bv);
     }
 
+    /// <summary>
+    /// A map type with no index arguments is represented by its result type:
+    /// Boogie2VCExprTranslator.Visit(MapSelect)/Visit(MapStore) translate a nullary
+    /// select/store to the underlying value, and SMTLibLineariser.TypeToString emits a
+    /// nullary map as the sort of its result type. The translated expression keeps the
+    /// map type, so anything that inspects an operand's type has to look through the
+    /// nullary map wrappers first.
+    /// </summary>
+    public static Type UnwrapNullaryMapType(Type t)
+    {
+      Contract.Requires(t != null);
+      Contract.Ensures(Contract.Result<Type>() != null);
+      while (t.IsMap && t.AsMap.Arguments.Count == 0)
+      {
+        t = Cce.NonNull(t.AsMap.Result);
+      }
+
+      return t;
+    }
+
     public VCExpr BvConcat(VCExpr bv1, VCExpr bv2)
     {
       Contract.Requires(bv2 != null);
       Contract.Requires(bv1 != null);
       Contract.Ensures(Contract.Result<VCExpr>() != null);
-      return Function(new VCExprBvConcatOp(bv1.Type.BvBits, bv2.Type.BvBits), bv1, bv2);
+      return Function(
+        new VCExprBvConcatOp(UnwrapNullaryMapType(bv1.Type).BvBits, UnwrapNullaryMapType(bv2.Type).BvBits),
+        bv1, bv2);
     }
 
     ////////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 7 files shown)


fabiomadge (dismissed):
I'll try to come up with an alternative

shazqadeer:
I am confused.

```
var x: []int; 
```

I wasn't even aware that such a declaration is possible. Should we just rule it out by rejecting the program in the parser or the type checker?

fabiomadge:
These empty maps are in the tests, so I think it's worth fixing them.

It seems there's a case missing:

nullary-unary.bpl:

 ``` boogie
  procedure Bug()  { var x: []int; x[] := 1; assert -x[] == -1; }
  procedure Good() { var x: int;   x   := 1; assert -x   == -1; }
```
with
```shell
boogie /proverOpt:LOGIC=AUFLIA nullary-unary.bpl
```
I believe they should behave the same way and that `Boogie2VCExpr.cs:1272` is to blame.

shaobo-he:
@fabiomadge good catch. Fixed.

Source/VCExpr/VCExpressionGenerator.cs
```diff
@@ -483,12 +483,34 @@ public VCExpr BvExtract(VCExpr bv, int bits, int start, int end)
...
+      return t;
+    }
+
     public VCExpr BvConcat(VCExpr bv1, VCExpr bv2)
     {
       Contract.Requires(bv2 != null);
       Contract.Requires(bv1 != null);
       Contract.Ensures(Contract.Result<VCExpr>() != null);
-      return Function(new VCExprBvConcatOp(bv1.Type.BvBits, bv2.Type.BvBits), bv1, bv2);
+      return Function(
```
shazqadeer, line 511: Why is the UnwrapNullaryMapType method being applied only in the special case of BvConcat?

shazqadeer:
@RustanLeino : I want your opinion on whether we should simply disallow NullaryMaps in Boogie by ruling them out in the parser. What do you think?

PR #1144, opened 2026-08-07
Fix interval transfer functions that report unreachable bounds

shaobo-he:
## Why this matters

With `-infer:j`, Boogie's interval analysis inserts loop invariants as `assume {:inferred} ...`. These assumptions are not proved later, so every inferred interval must contain every reachable value. A bound that is too tight is therefore a soundness bug: it can remove a real execution and make a false assertion verify.

I intend to land this fix. My motivation is to keep Boogie sound for downstream verification tools that rely on inferred invariants.

## Concrete example

```boogie
procedure P(x: int, y: int)
  requires -1 <= x;
  requires -1 <= y;
{
  var r: int;
  var i: int;
  r := x * y;
  i := 0;
  while (i < 1) { i := i + 1; }
  if (x == 100 && y == 100) {
    assert false;
  }
}
```

The interval domain represents an integer upper bound as an exclusive endpoint, `r < Hi`. On `master`, the multiplication rule computes `Hi = lo(x) * lo(y) + 1`. With `lo(x) = lo(y) = -1`, this gives `Hi = (-1) * (-1) + 1 = 2`, so Boogie infers `r < 2` (equivalently, `r <= 1`).

That calculation incorrectly treats the product of the lower bounds as the maximum product. In this example, `x = y = 100` satisfies the preconditions and gives `r = 10000`. The inferred `r < 2` assumption makes that execution unreachable, so the false assertion is incorrectly reported as verified.

With this PR, no upper bound is inferred in this case, and Boogie correctly reports the assertion as an error.

## What changes

The patch fixes three patterns in five arithmetic transfer-function branches:

| Operation | Problem with the old rule | New rule |
| --- | --- | --- |
| `x * y` when both lower bounds are negative | Used `lo(x) * lo(y)` as the upper bound, ignoring large positive operands. | If both upper bounds exist, use the maximum of the two relevant corner products. Otherwise, infer no upper bound. |
| `div`, `mod`, and real division | Inferred bounds when the divisor could be zero. Division by zero is unconstrained in the SMT encoding. For real division, `x / y <= x` also requires `y >= 1`. | Require the divisor's lower bound to be at least one before inferring these bounds. |
| `x ** y` | Inferred bounds from the operands even though Boogie encodes exponentiation as the uninterpreted function `real_pow` without axioms. | Infer no lower or upper bound. |

These changes deliberately prefer a weaker interval over an incorrect one. Losing precision may produce an extra verification error; inventing a bound can incorrectly verify a program.

## Tests

`UnsoundTransferFunctions.bpl` checks that:

- the multiplication example above now reports an error;
- integer division and modulo with a possibly-zero divisor no longer prove a nonnegative result;
- real division with a possibly-zero divisor no longer proves a nonnegative result;
- positive integer and real divisors still yield useful bounds; and
- multiplication still gets an upper bound when both operands are bounded above.

`PowBounds.bpl` is an inference-only test. It prints the instrumented program with `-noVerify` and checks that no bound on the power result is introduced, both for nonnegative operands and for a base of at least one. This directly tests the transfer function without asking the solver to interpret the otherwise-unaxiomatized `real_pow` function.

The existing `Linear5.bpl.expect` is updated because it previously pinned the invalid `x < 2` multiplication bound.

Validation with Z3 5.1.0: the solution builds with no warnings or errors, all 17 abstract-interpretation tests pass, and all 516 unit tests pass.

AI assistance: Claude Code helped investigate and implement the original change; I reviewed and extended the code and tests and intend to maintain it.

### DIFF
--- a/Source/AbstractInterpretation/IntervalDomain.cs
+++ b/Source/AbstractInterpretation/IntervalDomain.cs
@@ -1298,8 +1298,9 @@ namespace Microsoft.Boogie.AbstractInterpretation
                   // A product over a box is extremal at a corner. With both intervals reaching
                   // below zero the maximum is at the two lower bounds or at the two upper bounds
                   // (a corner mixing the two has a non-positive factor pairing, which cannot beat
-                  // lo0 * lo1 > 0). An unbounded upper end therefore leaves the product unbounded
-                  // above: taking lo0 * lo1 there claims a bound that x = y = 100 already exceeds.
+                  // lo0 * lo1 > 0). Without both upper bounds, decline to infer one: some such
+                  // intervals are still bounded above, but recognizing them requires more sign
+                  // cases than this incomplete approximation handles.
                   if (hi0 != null && hi1 != null)
                   {
                     var upper0 = isReal ? (BigInteger) hi0 : (BigInteger) hi0 - 1;
@@ -1353,16 +1354,8 @@ namespace Microsoft.Boogie.AbstractInterpretation
 
               break;
             case BinaryOperator.Opcode.Pow:
-              // this uses an incomplete approximation that could be tightened up.
-              // The exponent's upper bound says nothing about the size of the power, so there is
-              // no upper bound to report: with y <= 3, 2.0 ** y still reaches 8.0. The result is
-              // at least one only when the base is, since 0 ** 1 is 0.
-              if (lo0 != null && lo1 != null && 0 <= (BigInteger) lo0 && 0 <= (BigInteger) lo1)
-              {
-                Lo = 1 <= (BigInteger) lo0 ? BigInteger.One : BigInteger.Zero;
-                Hi = null;
-              }
-
+              // Real exponentiation is encoded as an uninterpreted function without axioms, so
+              // no result bound follows from bounds on its operands.
               break;
             default:
               break;
--- a/Source/AbstractInterpretation/IntervalDomain.cs
+++ b/Source/AbstractInterpretation/IntervalDomain.cs
@@ -1296,14 +1296,31 @@ namespace Microsoft.Boogie.AbstractInterpretation
                 else if ((BigInteger) lo0 < 0 && (BigInteger) lo1 < 0)
                 {
                   Lo = null; // approximation
-                  Hi = isReal ? lo0 * lo1 : lo0 * lo1 + 1;
+                  // A product over a box is extremal at a corner. With both intervals reaching
+                  // below zero the maximum is at the two lower bounds or at the two upper bounds
+                  // (a corner mixing the two has a non-positive factor pairing, which cannot beat
+                  // lo0 * lo1 > 0). An unbounded upper end therefore leaves the product unbounded
+                  // above: taking lo0 * lo1 there claims a bound that x = y = 100 already exceeds.
+                  if (hi0 != null && hi1 != null)
+                  {
+                    var upper0 = isReal ? (BigInteger) hi0 : (BigInteger) hi0 - 1;
+                    var upper1 = isReal ? (BigInteger) hi1 : (BigInteger) hi1 - 1;
+                    var max = BigInteger.Max((BigInteger) lo0 * (BigInteger) lo1, upper0 * upper1);
+                    Hi = isReal ? max : max + 1;
+                  }
+                  else
+                  {
+                    Hi = null;
+                  }
                 }
               }
 
               break;
             case BinaryOperator.Opcode.Div:
-              // this uses an incomplete approximation that could be tightened up
-              if (lo0 != null && lo1 != null && 0 <= (BigInteger) lo0 && 0 <= (BigInteger) lo1)
+              // this uses an incomplete approximation that could be tightened up.
+              // The divisor must be kept away from zero: division by zero is uninterpreted in
+              // SMT-LIB, so "x div 0" is not bounded by anything the operands say.
+              if (lo0 != null && lo1 != null && 0 <= (BigInteger) lo0 && 1 <= (BigInteger) lo1)
               {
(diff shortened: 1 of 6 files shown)


shazqadeer:
I have no idea what this PR is doing. I have tried to read the PR description a couple of times. It makes no sense to me.

shazqadeer:
@shaobo-he , @fabiomadge :

General question for you. I see a bunch of PRs put up by you. These look like automatically-generated by AI to me. Is it your intention to land these PRs? Or are you just playing with AI-assisted coding in the context of Boogie?

What was your motivation to fix these specific bugs?

shaobo-he:
> These look like automatically-generated by AI to me. Is it your intention to land these PRs? Or are you just playing with AI-assisted coding in the context of Boogie?

Yes, it's my intention to land them.

> What was your motivation to fix these specific bugs?

I'm still building tools based on Boogie, and I want to eliminate soundness bugs as many as possible.

fabiomadge:
@shazqadeer I want to land my two open PRs. I rewrote most of the float code while adding support for floats to Dafny, but there are some issues, which I think deserve fixes.

shaobo-he:
> I have no idea what this PR is doing. I have tried to read the PR description a couple of times. It makes no sense to me.

Hi @shazqadeer, I revised the PR description. I hope it now makes the fix easier to understand.

Source/AbstractInterpretation/IntervalDomain.cs
```diff
@@ -1295,23 +1295,42 @@ public override Expr VisitNAryExpr(NAryExpr node)
                 else if ((BigInteger) lo0 < 0 && (BigInteger) lo1 < 0)
                 {
                   Lo = null; // approximation
-                  Hi = isReal ? lo0 * lo1 : lo0 * lo1 + 1;
+                  // A product over a box is extremal at a corner. With both intervals reaching
```
shazqadeer, line 1298: Please make this comment brief and less dramatic.

Source/AbstractInterpretation/IntervalDomain.cs
```diff
@@ -1295,23 +1295,42 @@ public override Expr VisitNAryExpr(NAryExpr node)
...
+                    Hi = null;
+                  }
                 }
               }
 
               break;
             case BinaryOperator.Opcode.Div:
-              // this uses an incomplete approximation that could be tightened up
-              if (lo0 != null && lo1 != null && 0 <= (BigInteger) lo0 && 0 <= (BigInteger) lo1)
+              // this uses an incomplete approximation that could be tightened up.
```
shazqadeer, line 1320: The incomplete axiomatization of miscellaneous arithmetic operations is irrelevant to the soundness of abstraction interpretation transfer rules which is the focus of this PR.

I agree with the change. But I would like the comment to reflect the correct intention.

Source/AbstractInterpretation/IntervalDomain.cs
```diff
@@ -1295,23 +1295,42 @@ public override Expr VisitNAryExpr(NAryExpr node)
...
-                  Hi = isReal ? lo0 * lo1 : lo0 * lo1 + 1;
+                  // A product over a box is extremal at a corner. With both intervals reaching
+                  // below zero the maximum is at the two lower bounds or at the two upper bounds
+                  // (a corner mixing the two has a non-positive factor pairing, which cannot beat
+                  // lo0 * lo1 > 0). Without both upper bounds, decline to infer one: some such
+                  // intervals are still bounded above, but recognizing them requires more sign
+                  // cases than this incomplete approximation handles.
+                  if (hi0 != null && hi1 != null)
+                  {
+                    var upper0 = isReal ? (BigInteger) hi0 : (BigInteger) hi0 - 1;
```
shazqadeer, line 1306: I don't have the complete context of this code: what is the purpose of isReal?

Source/AbstractInterpretation/IntervalDomain.cs
```diff
@@ -1295,23 +1295,42 @@ public override Expr VisitNAryExpr(NAryExpr node)
...
                 Lo = BigInteger.Zero;
                 Hi = hi0;
               }
 
               break;
             case BinaryOperator.Opcode.Mod:
-              // this uses an incomplete approximation that could be tightened up
-              if (lo0 != null && lo1 != null && 0 <= (BigInteger) lo0 && 0 <= (BigInteger) lo1)
+              // this uses an incomplete approximation that could be tightened up.
+              // As for Div, a divisor of zero leaves the result uninterpreted.
```
shazqadeer, line 1332: What is the purpose of this comment about Div in a change to Mod?

Source/AbstractInterpretation/IntervalDomain.cs
```diff
@@ -1324,22 +1343,19 @@ public override Expr VisitNAryExpr(NAryExpr node)
...
               break;
             case BinaryOperator.Opcode.Pow:
-              // this uses an incomplete approximation that could be tightened up
-              if (lo0 != null && lo1 != null && 0 <= (BigInteger) lo0 && 0 <= (BigInteger) lo1)
-              {
-                Lo = 1 <= (BigInteger) lo1 ? BigInteger.One : BigInteger.Zero;
-                Hi = hi1;
-              }
-
+              // Real exponentiation is encoded as an uninterpreted function without axioms, so
```
shazqadeer, line 1357: I don't understand this change. Is there an unsoundness in this transfer function? If not, no need to change it.

shazqadeer:
@fabiomadge , @shaobo-he : I have a general recommendation about putting up PRs based on AI assistance. Please vet the code and the comments yourself before you put up the PR. This will help the reviewers focus on the essence of the PR.

PR #1149, opened 2026-08-13
Fix stack overflow on a recursive polymorphic inlined procedure

### LINKED ISSUE #1147: Stack overflow: monomorphizer recurses forever on a recursive polymorphic `{:inline}` procedure
A recursive polymorphic `{:inline}` procedure sends the monomorphizer into unbounded recursion, aborting the process.

## Repro

Five lines, no command-line flags:

```boogie
procedure {:inline 1} P<T>(x: T, n: int) returns (y: T)
{
  if (n > 0) { call y := P(x, n-1); } else { y := x; }
}
procedure main() { var a: int; var b: int; a := 3; call b := P(a, 2); assert b == 3; }
```

```
$ boogie rec_inline.bpl
Stack overflow.
   at System.Object.MemberwiseClone()
   at Microsoft.Boogie.Absy.Clone()
   ...
   at Microsoft.Boogie.MonomorphizationVisitor.InstantiateAbsy(...)
   at Microsoft.Boogie.MonomorphizationVisitor+<>c__DisplayClass41_0.<InstantiateImplementation>b__0(Block)
   at Microsoft.Boogie.MonomorphizationVisitor.InstantiateImplementation(Implementation, List<Type>)
   at Microsoft.Boogie.MonomorphizationVisitor.InlineCallCmd(CallCmd, List<Type>)
   at Microsoft.Boogie.MonomorphizationDuplicator.VisitCallCmd(CallCmd)
   ...
[aborted, exit code 134]
```

Reproduced on master @ 977902c3 (reports version 3.5.7.0).

The program's recursion is bounded (`n = 2`), but the monomorphizer recurses on the syntactic call rather than the value, so the bound is irrelevant — any recursive polymorphic `{:inline}` procedure triggers it.

## Controls

Each row is the same file with one thing changed. All three terminate normally:

| change | result |
| --- | --- |
| `<T>` deleted (monomorphic) | terminates |
| `{:inline 1}` deleted | `1 verified, 1 error` |
| `/inline:none` added | `1 verified, 1 error` |

So the trigger is polymorphism × recursion × inlining, and `/inline:none` is a workaround.

## Cause

`Source/Core/Monomorphization.cs:1713`:

```csharp
public Implementation InstantiateImplementation(Implementation impl, List<Type> actualTypeParams)
{
  if (!implInstantiations[impl].ContainsKey(actualTypeParams))    // :1715  guard reads here
  {
    ...
    var blocks = impl.Blocks
      .Select(block => (Block)InstantiateAbsy(...)).ToList();      // :1722  recurses here
    ...
    implInstantiations[impl][actualTypeParams] = instantiatedImpl; // :1739  memo written here
  }
  return implInstantiations[impl][actualTypeParams];
}
```

Memoize-after-recurse. The block walk at :1722 re-enters through

`InstantiateImplementation → InstantiateAbsy → VisitCallCmd (:1017) → InlineCallCmd (:1584) → InstantiateImplementation`

with the same `(impl, actualTypeParams)` while the memo entry is still absent, so the guard at :1715 never fires. The dump is 95,952 frames — 5,642 turns of that cycle.

The existing `ExpandingTypeCycle` check does not catch it: that detects *growing* type arguments, whereas here every instantiation is the identical `[int]`.

This is unbounded recursion over a finite AST, so it is distinct from #711 (deep-but-finite traversals) — no stack size helps. It is also distinct from #363, which concerned a polymorphic *function* and failed during typechecking.

## Suggested fix

Register the instantiation before walking the body: construct the `Implementation` with an empty block list, store it in `implInstantiations`, then populate `.Blocks`. The recursive re-entry then hits the memo. A partially built shell is never observed, since `InlineCallCmd` discards the returned value and `VisitCallCmd` links the call to the instantiated *procedure*, not the implementation. `Implementation`'s constructor only stores the block list and `Blocks` has a public setter, so hoisting the shell above the walk is mechanical.

## Note for whoever writes the regression test

The repro above is a poor choice of `.expect` fixture, for a reason unrelated to the crash. Its monomorphic control passes for a vacuous reason: under the default `/inline:assume`, reaching the inlining depth replaces the recursive call with `assume false` (documented behaviour, `CommandLineOptions.cs:1562`), and `n = 2` needs three levels to reach the base case. Below that, every path through `main` is cut:

| `{:inline N}`, monomorphic | `assert b == 3` | `assert b == 4` |
| --- | --- | --- |
| 1 | `1 verified, 0 errors` | `1 verified, 0 errors` |
| 2 | `1 verified, 0 errors` | `1 verified, 0 errors` |
| 3 | `1 verified, 0 errors` | `0 verified, 1 error` |
| 4 | `1 verified, 0 errors` | `0 verified, 1 error` |

So an expectation built on `{:inline 1}` would pass whether or not the fix is correct. `{:inline 3}` — or `/inline:spec`, which yields `1 verified, 1 error` at any depth — discriminates.

Happy to open a PR if that direction looks right.

shaobo-he:
Fixes #1147.

A recursive polymorphic `{:inline}` procedure sent the monomorphizer into unbounded recursion, aborting the process with empty stdout:

```boogie
procedure {:inline 1} P<T>(x: T, n: int) returns (y: T)
{
  if (n > 0) { call y := P(x, n-1); } else { y := x; }
}
procedure main() { var a: int; var b: int; a := 3; call b := P(a, 2); assert b == 3; }
```

An inlined callee is instantiated at the point of the call, so the body being walked asks for its own instantiation — and that instantiation is not entered into `implInstantiations` until the walk completes, so the guard in `InstantiateImplementation` never fires. Mutual recursion and longer cycles trigger it too. `ExpandingTypeCycle` does not catch it, since the type argument never grows.

Fix: track the instantiations whose body is being walked, and have `InlineCallCmd` skip a call that leads back into one. The set is keyed by the implementation *and* its type arguments, because a different instantiation of the same implementation can legitimately be built while one is in progress.

Hoisting the `Implementation` construction above the block walk also breaks the recursion, but `Absy`'s constructor draws from a process-global id counter that feeds `MkInstanceName`, so allocating it earlier renames every generated instance — changing `/print` output, the SMT-LIB handed to the solver, and inlined labels in execution traces, for programs that never had the bug. Guarding the call site instead leaves all of that byte-identical.

The test covers self-recursion, a second type instantiation, mutual recursion, and a nested instantiation. Its error cases use an inlining depth that reaches the base case, so they are not vacuously discharged by the `assume false` that `/inline:assume` inserts at the depth bound.

### DIFF
--- a/Source/Core/Monomorphization.cs
+++ b/Source/Core/Monomorphization.cs
@@ -1288,11 +1288,6 @@ namespace Microsoft.Boogie
     private Dictionary<Function, Dictionary<List<Type>, Function>> functionInstantiations;
     private Dictionary<Procedure, Dictionary<List<Type>, Procedure>> procInstantiations;
     private Dictionary<Implementation, Dictionary<List<Type>, Implementation>> implInstantiations;
-
-    // Instantiations whose body is currently being walked. An instantiation is not entered into
-    // implInstantiations until its body is complete, so this is what tells InlineCallCmd that a
-    // call it is about to inline leads back into an instantiation that is already being built.
-    private Dictionary<Implementation, HashSet<List<Type>>> implInstantiationsInProgress;
     private Dictionary<TypeCtorDecl, Dictionary<List<Type>, TypeCtorDecl>> typeInstantiations;
     private HashSet<Declaration> newInstantiatedDeclarations;
     private List<BinderExprMonomorphizer> binderExprMonomorphizers;
@@ -1317,14 +1312,12 @@ namespace Microsoft.Boogie
       originalAxiomToSplitAxioms = program.Axioms.ToDictionary(ax => ax, _ => new HashSet<Axiom>());
       originalAxiomToOriginalSymbols = program.Axioms.ToDictionary(ax => ax, ax => new FunctionAndConstantCollector(ax));
       implInstantiations = new Dictionary<Implementation, Dictionary<List<Type>, Implementation>>();
-      implInstantiationsInProgress = new Dictionary<Implementation, HashSet<List<Type>>>();
       nameToImplementation = new Dictionary<string, Implementation>();
       program.TopLevelDeclarations.OfType<Implementation>().Where(impl => impl.TypeParameters.Count > 0).ForEach(
         impl =>
         {
           nameToImplementation.Add(impl.Name, impl);
           implInstantiations.Add(impl, new Dictionary<List<Type>, Implementation>(new ListComparer<Type>()));
-          implInstantiationsInProgress.Add(impl, new HashSet<List<Type>>(new ListComparer<Type>()));
         });
       procInstantiations = new Dictionary<Procedure, Dictionary<List<Type>, Procedure>>();
       nameToProcedure = new Dictionary<string, Procedure>();
@@ -1593,11 +1586,9 @@ namespace Microsoft.Boogie
       if (procToImpl.ContainsKey(callCmd.Proc))
       {
         var impl = procToImpl[callCmd.Proc];
-        // Inlining a call whose target instantiation is still being built would walk that body a
-        // second time and re-enter here on the same call, so a recursive inlined procedure would
-        // recurse without bound and overflow the stack. The instantiation already under
-        // construction is the one this call needs; it is completed by the frame that started it.
-        if (IsInlined(impl) && !implInstantiationsInProgress[impl].Contains(actualTypeParams))
+        // The dictionary contains a null sentinel while an instantiation's body is being walked.
+        // That prevents recursive inlined calls from trying to build the same instantiation again.
+        if (IsInlined(impl) && !implInstantiations[impl].ContainsKey(actualTypeParams))
         {
           InstantiateImplementation(impl, actualTypeParams);
         }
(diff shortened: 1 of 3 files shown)


shazqadeer:
Thanks for taking the time to fix this issue, @shaobo-he.

I would like to examine the possibility of fixing the problem without introducing another dictionary. Would it be possible to put a partially-initialized implementation in the dictionary before the recursion commences on the blocks? We can compute all pieces of the implementation except the blocks before constructing the Implementation object with a null blocks  value. After the blocks have been computed, we can update the blocks fields.

shaobo-he:
Thanks, Shaz. I updated the branch to avoid the additional dictionary.

Rather than constructing a partially initialized `Implementation`—which would move its `Absy` allocation earlier and change generated instance IDs and output—the existing `implInstantiations` dictionary now gets a temporary null entry immediately before the blocks are walked. `InlineCallCmd` checks whether the key exists, so a recursive call stops at that entry; once the traversal completes, the entry is replaced with the completed `Implementation`. This preserves the original construction order and also avoids passing null blocks to the `Implementation` constructor.

The change is in 23beff9f. I built the solution and ran all 54 monomorphization tests and all 516 unit tests; everything passes. Does this approach address your concern?

shazqadeer (approved):

shazqadeer:
Looks good. Thanks, @shaobo-he

shaobo-he:
Thanks, @shazqadeer. Can I ask when this fix can be released?

shazqadeer:
Please put up a PR to create a new release.

Source/Core/Monomorphization.cs
```diff
@@ -1586,7 +1586,9 @@ public void InlineCallCmd(CallCmd callCmd, List<Type> actualTypeParams)
       if (procToImpl.ContainsKey(callCmd.Proc))
       {
         var impl = procToImpl[callCmd.Proc];
-        if (IsInlined(impl))
+        // The dictionary contains a null sentinel while an instantiation's body is being walked.
+        // That prevents recursive inlined calls from trying to build the same instantiation again.
+        if (IsInlined(impl) && !implInstantiations[impl].ContainsKey(actualTypeParams))
```
shazqadeer, line 1591: @shaobo-he : I just realized that this change is not necessary.

If we left the code as is, then execution will descend into InstantatiateImplementation and perform the identical check. It won't attempt to instatiate the implementation again and return null.

Do you agree?

PR #1154, opened 2026-08-31
Document the contract of Boogie's float type

fabiomadge:
Boogie has no language reference, so what its float operators mean is recoverable only by reading the
SMT-LIB translator. This states it on `FloatType`, and pins the parts no test covered.

`==` and `!=` are SMT `=`: bit identity, total, reflexive even at NaN, and keeping `-0.0` and `+0.0`
apart. `<`, `<=`, `>`, `>=` are `fp.lt`, `fp.leq`, `fp.gt`, `fp.geq`: partial, false whenever an operand is
NaN, and tying the two zeros. Neither implies the other, so trichotomy fails — a negated order relation is
not the reverse relation, and a pair of numeric bounds does not describe a float. Arithmetic is the `fp.*`
operations under round-to-nearest-even; other rounding modes, the IEEE predicates and `fp.eq` are reachable
only through `{:builtin}`.

**Testing.** `Test/floats/FloatContract.bpl` pins the claims `Equal2.bpl` and `SpecialValues.bpl` do not:
four procedures verify and `assert x <= x` fails, which is what the doc says.

One of six; best read first, since #1156 and #1157 cite it.


shazqadeer (dismissed):

fabiomadge:
Thanks for the review, but feel free to wait with the others until I can have a proper look myself and undraft them.

PR #1155, opened 2026-08-31
fix: Crash in abstract interpretation on a bool measure

fabiomadge:
`MeasureChecker` injects its measure comparison into an already-typechecked program. `Expr.Not` does not
set `Type`, so the `And` it builds has an untyped first operand, and `ThresholdFinder` dereferences it.
Only a bool measure gets there: the int path compares two identifiers, whose first operand is typed.

```boogie
var y: bool;

procedure P()
{
  var n: int;
  n := 10;
  while (n >= 1)
  measure n, y;
  {
    n := n - 1;
  }
}
```

`-infer:j` aborts in `ThresholdFinder.VisitNAryExpr`.

**Fix.** The `Expr` overloads that take no token are the ones used after typechecking — the parser builds
through those taking one — so they now type what they build, from `ShallowType`. That covers `Not`, `Neg`
and every comparison and arithmetic wrapper, and needs no change in `MeasureChecker`. `TypeParameters`
stays null, because `NAryExpr.Typecheck` reads it to tell an already-checked node from a fresh one.

`ThresholdFinder` also tolerates a null type, which is not belt-and-braces: Civl's map desugaring
synthesises `FunctionCall` applications after typechecking and leaves them untyped, and 54 of the `civl/`
programs reach the reader with one. Reverting the guard leaves all 54 aborting.

**Scope.** `-infer:j` is off by default in Boogie and unconditional in Dafny. On master it aborts on 60 of
761 test programs: 6 under `measure/` and 54 under `civl/`.

**Testing.** New `Test/infer-crash-sweep.sh` runs every test program under `-infer:j` and fails on any
abort; it reports 0 here and found the 60. Not wired into CI: serial, about five times the `lit` suite.

One of six.


Source/AbstractInterpretation/IntervalDomain.cs
```diff
@@ -1629,7 +1629,10 @@ public override Expr VisitNAryExpr(NAryExpr node)
         Contract.Assert(node.Args.Count == 2);
         var arg0 = node.Args[0];
         var arg1 = node.Args[1];
-        var offset = arg0.Type.IsReal ? 0 : 1;
+        // An untyped expression should not reach here, but a producer that leaves Type unset is a
```
shazqadeer, line 1632: I am not following this code change. Shouldn't the fix be to have the Type populated always?

Source/Core/MeasureChecker.cs
```diff
@@ -353,6 +353,30 @@ private void TransformMeasureCmds(Program program)
       }
     }
 
+    // Expr.And and its siblings leave Type unset, and what this file builds goes into a program that has
+    // already been typechecked, where a visitor is entitled to read the type of what it finds --
+    // ThresholdFinder reads it of the first operand of every binary operator. So build with these.
```
shazqadeer, line 358: A comment in this file should not refer to ThresholdFinder.

Source/Core/MeasureChecker.cs
```diff
@@ -353,6 +353,30 @@ private void TransformMeasureCmds(Program program)
...
+      if (e is NAryExpr { Type: null } nary)
+      {
+        nary.Type = Type.Bool;
+        nary.TypeParameters = SimpleTypeParamInstantiation.EMPTY;
+      }
+
+      return e;
+    }
+
+    private static Expr Not(Expr e) => Typed(Expr.Not(e));
```
shazqadeer, line 374: Why shouldn't these changes be applied to the public static constructors in the Expr class? This change of typing expressions seems to be generally useful.

shazqadeer:
@fabiomadge : You are addressing a legitimate issue here. Thanks. Would it be more robust to add the Type population to public static constructors in Expr class?

PR #1156, opened 2026-08-31
fix: `!(x < y)` is rewritten to `y <= x` even for floats

fabiomadge:
`Expr.Not` turned a negated order relation into the reverse relation for every type. That is valid only
where the order is total: every IEEE comparison is false when an operand is NaN, so `!(a < b)` holds of a
NaN while `b <= a` does not.

```boogie
procedure P(f: float24e8)
{
  if (f < 0x1.0e0f24e8) { } else { assert 0x1.0e0f24e8 <= f; }
}
```

No flags → **1 verified, 0 errors**. The `else` branch is reachable with a NaN, where the assertion is
false.

**Fix.** `Expr.Not` leaves an order relation alone unless the operand's type is known and not float. It
usually is not known, and cannot be made so: the guard is negated from the `Implementation` constructor,
which flattens the structured statements while parsing, before any identifier is resolved. So no
reordering inside resolution can supply a type, and `Program.ReverseGuardNegations` asks again once types
exist. The second commit has `IntervalDomain.Constraint` push a hand-written `!` through the same helper,
which it never did.

**Scope.** Equality is untouched, `==` on floats being bit identity and so total. Inference gets sharper
where the domain can now read a negation: `aitest1/ineq.bpl` infers `i == 10` where it inferred
`0 <= i && i < 11`. Two printed forms change. `roundingmodes/InvalidOperators.bpl` loses four phantom type
errors, which named operators the program does not contain. `inline/test4.bpl` keeps an unreversed guard,
because a `-print` of the program as parsed runs before resolution while the blocks already exist, so the
reversal cannot reach it.

**Not addressed.** The pass finds guards by `{:partition}`, which a user may write; such an assume gets the
same rewriting, semantics-preserving but visible through `/print`.

One of six. `TryPushNegation` sets `Type` and `TypeParameters` independently, which matters once #1155
lands: it types comparisons at construction, and a node with a type but no type parameters crashes
`MonomorphizationDuplicator`.


shazqadeer:
@fabiomadge : Is it possible to solve the problem by avoiding any  attempt to rewrite negations until the type is determined (and known not to be float)? That seems simpler at a first glance.

PR #1157, opened 2026-08-31
fix: Abstract interpretation proves false assertions about floats

fabiomadge:
The interval domain describes a variable by a pair of integer bounds, and asked float literals for theirs.
A float has no such value to bound: arithmetic rounds and overflows, so bounds computed from the operands
need not hold of the result; the order ties `-0.0` and `+0.0` where `==` separates them, so a one-point
interval is not one point; and no bound at all holds of a NaN.

```boogie
procedure P()
{
  var m: float24e8;
  var i: int;

  m := 0x1.0e0f24e8;
  i := 0;
  while (i < 3)
  {
    if (m == 0x1.0e0f24e8) { m := 0x2.0e0f24e8; } else { m := 0x1.0e0f24e8; }
    i := i + 1;
  }
  assert m == 0x1.0e0f24e8;
}
```

`-infer:j` → **1 verified, 0 errors**. Three flips leave `m` at 2.0. A float pinned to a literal has equal
bounds, and `Meet` tests emptiness the way it does for an int, where `Lo == Hi` means empty — so the
then-branch was inferred unreachable.

**Fix.** The domain computes no bounds for a float: the `BigFloat` case in `PEVisitor.VisitLiteralExpr`
goes, and with it the float arms of `Node.ToExpr` and `NumberToExpr`. `ToExpr` keeps the real arm where it
was and replaces the float one with an unreachable guard, which also makes the diff read the way it
behaves — the two arms had identical bodies, so deleting the float one rendered as deleting the real one.
Both `Node` constructors assert the invariant, and the literal fallthrough asserts that a float leaves the
bounds cleared.

**Scope.** What is lost is any inference about a float. The domain already declines the one other type it
cannot describe — a bitvector is never tracked — so this brings floats into line with that. Of the ten
tests added, seven pin a consequence that used to be proved falsely, one an unhandled `ArgumentException`,
one that variables the domain can describe are still tracked in a program mentioning floats, and one the
property itself.

One of six; #1154 should land first, since a comment here cites it.


Source/AbstractInterpretation/IntervalDomain.cs
```diff
@@ -47,7 +48,9 @@ public override Expr ToExpr(CoreOptions options)
 
     public class Node
     {
-      public readonly Variable V; // variable has type bool or int
+      // Never of float type: see PEVisitor.VisitLiteralExpr for why, and the constructors, which say so
```
shazqadeer, line 51: This comment does not make any sense.

Source/AbstractInterpretation/IntervalDomain.cs
```diff
@@ -987,6 +946,12 @@ public override Expr VisitLiteralExpr(LiteralExpr node)
           }
         }
 
+        // No case for BigFloat: a float literal is deliberately left unbounded, which is what keeps
+        // floats out of the domain. VisitIdentifierExpr does not look one up either, so no float
+        // expression has bounds, and hence no float variable is ever constrained (Constrain) or assigned
+        // bounds (Update). A pair of integer bounds cannot describe a float at all, for the reasons set
+        // out on FloatType.
+
         return node;
```
shazqadeer, line 955: Is this 'return node' happening for BigFloat type with other cases being considered above? If so, please add an assertion here to document.

Source/AbstractInterpretation/IntervalDomain.cs
```diff
@@ -260,36 +265,9 @@ public Expr ToExpr(CoreOptions options)
 
           return e;
         }
```
shazqadeer, line 267: Why are we deleting all this code for Real type? I thought we are focused on the Float type in this PR.

PR #1158, opened 2026-08-31
fix: `{:builtin}` accepts a definition, or a symbol Boogie emits itself

fabiomadge:
Two commits, one per check.

A `{:builtin ...}` function IS the named solver symbol: its applications are translated to that symbol and
no declaration is emitted for it. Nothing checked what else the declaration says, and a definition becomes
an axiom — which would then be an axiom about the solver's symbol.

```boogie
function {:builtin "+"} plus(x: int, y: int): int { 0 }
procedure P(a: int, b: int) { assert a + b == 0; }
```

No flags → **1 verified, 0 errors**, for a program asserting that every integer sum is zero.

**Fix.** A `{:builtin}` function may not also have a `Body`, `DefinitionBody` or `DefinitionAxiom`, nor name
a symbol Boogie emits for its own encoding. `Core/EmittedSymbols.cs` holds the latter: fixed names
harvested from a prover log, plus the constructors of a boxed type. Those are named after the type itself —
`NType`, and one `NTypeInv<i>` per argument — so which names they are depends on the program, and matching
their shape keeps the check answerable from one declaration. That costs a little precision:
`{:builtin "GhostType"}` is rejected although nothing declares a `GhostType`. Nothing is lost by it, a
`{:builtin}` having to name a solver symbol exactly and no solver naming one this way.

This is distinct from `SmtLibNameUtils.reservedSmtWordsList`, which exists so a colliding Boogie
*identifier* can be renamed; renaming is not available here, the attribute naming a symbol exactly.

**Scope.** An axiom about a `{:builtin}` function stays allowed: that is the point of the attribute.
Rejection is flag-independent, so `{:builtin "int_2_U"}` errors even under the default monomorphic
encoding, which never emits that symbol — the attribute claims identity with a solver symbol and should
not mean different things per flag.

**Not addressed.** Arity and sorts are still unvalidated: a wrong arity reaches the solver and comes back
as `Prover error: invalid number of arguments`, followed by a claim that the prover died out of memory.

One of six. The two checks are independent, so this can be split if that is easier to land.


shazqadeer:
What are the "boxing names" the PR description refers to?

